PR #86, opened 2022-08-24
fix(adapter): fix error read from multi streams

lenonqing:


rueian:
Hi @lenonqing,
Thanks for your PR!

PR #87, opened 2022-08-24
Most recently released redis modules

chayim:
RedisGraph 2.8.19

RedisBloom 2.2.8

RediSearch 2.4.14

RedisTimeSeries 1.6.17

RedisAI 1.2.7

RedisGears 1.2.5

RedisJSON 2.2.0

### DIFF
--- a/hack/cmds/commands_ai.json
+++ b/hack/cmds/commands_ai.json
@@ -1,6 +1,5 @@
 {
-  "AI.TENSORSET":
-  {
+  "AI.TENSORSET": {
     "summary": "stores a tensor as the value of a key.",
     "complexity": "O(1)",
     "arguments": [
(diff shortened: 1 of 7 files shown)


rueian:
Hi @chayim,

Thank you for the PR.

However, these JSON files are meant to be used by the command generator (./hack/cmds/gen.go) and unfortunately the generator does not read the new redis command.json format (ex `type: pure-token`). So this PR does not work currently.

Instead of supporting the new format now, I suggest we maintain these json files manually until all Redis Modules switch their command.json to the new format as well. And based on my experience, their command.json sometimes differ from the real behavior and require manual fix anyway.

I also checked your diff line by line. It seems that the only module need to be updated is RedisGears? Others look like equivalent with the current version.

chayim:
@rueian If you can help me understand the generator, I'm happy to poke it a bit.

RedisBloom has changes in return types, as well as arguments - at least for the TDIGEST series of functions. Similarly, there are upcoming command changes in RedisTimeSeries 1.8, and RediSearch 2.6.

rueian:
Hi @chayim,

After verifying TDIGEST is the only series of functions that need to be regenerated, I have updated and merged your changes. Commands of RedisTimeSeries 1.8 and RediSearch 2.6 are also covered in the latest commit.

In terms of the generator, it is just a script that generates the command builder. The implementation of the generator is not important though. It may be even better if we could rewrite a new one for the new redis command.json format.

PR #90, opened 2022-08-29
Add CacheInvalidateCallback

altanozlu:
We use redis client caching with BCAST so we have to implement custom mechanism.
I also added p.background() because otherwise it didn't get called.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -121,7 +121,9 @@ func newPipe(conn net.Conn, option *ClientOption) (p *pipe, err error) {
 		}
 	}
 	p.cacheInvalidateCallback = option.CacheInvalidateCallback
-
+	if p.cacheInvalidateCallback != nil {
+		p.background()
+	}
 	return p, nil
 }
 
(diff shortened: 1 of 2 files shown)


rueian:
Hi @altanozlu, thank your for your PR. The PR looks good to me. I just feel `CacheInvalidateCallback` is a little bit too long. Maybe `OnInvalidations` is enough?

Another thing is that I don’t understand how this callback mechanism related to BCAST client-side caching. Would you mind elaborate how do you use BCAST currently and how will you use this callback with it?

altanozlu:
I use BCAST to listen for prefixes, but I don't use get directly on current connection, another service uses get, set, del. I use custom caching mechanism and I need to know if the key is invalidated or not.

rueian:
Hi @altanozlu, that is interesting. So you just need those notifications without caching values in memory of client side? Redis has another feature called keyspace notification. Maybe it also suite for your need.

altanozlu:
Hi @rueian, thanks for letting me know about keyspace notifications i'll check it. Actually we cache value but not the value in redis. It's like i call grpc.helloworld() and server uses 3 different redis get, the grpc client gets key information from headers and says this response mapped to this keys. when any of these 3 keys invalidate we invalidate the response.

PR #122, opened 2022-11-07
Exported synonyms for internal Completed, Cacheable & Builder

mdigger:
It will be possible to create functions that return prepared Redis commands:
```golang
type Object struct {
    ID        string
    Datetime time.Time
    Message string
}

func (o Object) SaveCmd(client rueidis.Client) rueidis.Completed {
    return client.B().
        Hset().
        Key("out:interval:"+o.ID).
        FieldValue().
        FieldValue("datetime", fmt.Sprint(o.Datetime.Unix())).
        FieldValue("message", o.Message).
        Build()
}
```

### DIFF
--- a/internal.go
+++ b/internal.go
@@ -1,0 +1,10 @@
+package rueidis
+
+import "github.com/rueian/rueidis/internal/cmds"
+
+// Exported synonyms for command types.
+type (
+	Completed = cmds.Completed
+	Cacheable = cmds.Cacheable
+	Builder   = cmds.Builder
+)


mdigger:
I didn't replace `cmd.Completed` with `[rueidis].Completed` in the entire library, but it makes sense to do these. Especially in the interface description:
```golang
type Client interface {
    B() Builder
    Do(ctx context.Context, cmd Completed) (resp RedisResult)
    DoMulti(ctx context.Context, multi ...Completed) (resp []RedisResult)
    DoCache(ctx context.Context, cmd Cacheable, ttl time.Duration) (resp RedisResult)
    DoMultiCache(ctx context.Context, multi ...CacheableTTL) (resp []RedisResult)
    Receive(ctx context.Context, subscribe Completed, fn func(msg PubSubMessage)) error
    Dedicated(fn func(DedicatedClient) error) (err error)
    Dedicate() (client DedicatedClient, cancel func())
    Nodes() map[string]Client
    Close()
}
```

rueian:
I guess I will probably follow this PR to make `cmd.Completed` public before rueidis v1, but I haven't have a concrete idea about how to do that.

Just reference some requirements I mentioned in the issue https://github.com/rueian/rueidis/issues/121#issuecomment-1307346862 in case if we need further discussion here.

In addition to just make it public, there might be chances of adding more features, such as allowing command to be reused.

mdigger:
I "played around" a little with the command generator and I got the following:
https://github.com/mdigger/redis-cmds-generation-experiment

There is no such problem! :)

I changed too much there while I was dealing with the generator. So I'm not suggesting this as a substitute.
Based idea: `Completed` & `Cacheabe` are interface...

rueian:
> I "played around" a little with the command generator and I got the following:
> https://github.com/mdigger/redis-cmds-generation-experiment

Hi @mdigger, is this repository public? I can't see it.

mdigger:
I'm sorry! Made it public.

rueian:
> Based idea: Completed & Cacheabe are interface...

I am worry about if this approach will cause additional GC overhead, for example the built `cmd.completed` may easily escaped to heap?

mdigger:
```golang
func BenchmarkCompleted(b *testing.B) {
	z := completed{
		command: newCommand("GET", "TEST"),
		ctags:   ctagReadOnly,
		cslot:   0,
	}
	var r Completed
	r = z

	b.Run("via interface", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			s := r.String()
			_ = s
		}
	})

	b.Run("direct", func(b *testing.B) {
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			s := z.String()
			_ = s
		}
	})
}
```

|                                     |          |             |        |             |
|-------------------------------------|----------|-------------|--------|-------------|
| BenchmarkCompleted/via_interface-10 | 46866532 | 24.32 ns/op | 8 B/op | 1 allocs/op |
| BenchmarkCompleted/direct-10        | 50613157 | 23.29 ns/op | 8 B/op | 1 allocs/op |

rueian:
Hi @mdigger,

Thank you for benchmarking it. But your benchmark is probably too minimal to test heap allocations.

I am worry about whether the built command will be easily escaped to heap if we pass it to the client.

For example with the following client implementation, which is actually very similar to the implementation rueidis used:

```go
type Client struct {
	ch chan Completed // Completed is a interface
}

//go:noinline
func (c *Client) Do(ctx context.Context, cmd Completed) {
	c.ch <- cmd
	return
}
```

We pass the built command into it:
```go
func TestPassInterface(t *testing.T) {
	cmd := Builder{}.Get().Key("").Build() // this is line 20
	new(Client).Do(nil, cmd)
}
```
Then we compile it with `-gcflags="-m"`, we can see this message saying that the `cmd` on the line 20 is allocated on heap and implying we need to pay additional GC overhead some time later:

```sh
▶ go test -gcflags="-m" client_test.go
./client_test.go:20:17: cmd escapes to heap
```

mdigger:
I understand, but I do not know any other way to solve the problem of creating `Completed` instances.

If there is a `struct`, it doesn't matter if it is public or not, but you can always create a new variable of this type. If you want to prohibit the creation of new elements of some type, then the only way I know is to define it as an interface.

```golang
// These cases should be rejected.
client.Do(ctx, cmds.Completed{})
```

rueian:
> If there is a `struct`, it doesn't matter if it is public or not, but you can always create a new variable of this type.

At least for now, the `client.Do(ctx, cmds.Completed{})` case can't be compiled, right?

But, yes, if we make it public, then we probably need to have some flag indicating its completeness inside the struct and check it on runtime.

mdigger:
```golang
cmds := make(rueidis.Commands, 1)
client.Do(context.Background(), cmds[0])
```
;)

mdigger:
Or use interfaces so that you cannot create an instance yourself. Or make it public and allow you to do anything with it, but return an error when trying to use it.

rueian:
> In general, I don't see a big problem with heap when using interfaces: what prevents you from reusing them via sync.Pool?

Good point. But I am not sure if they are worth to be stored in sync.Pool. The `Completed` struct currently occupies only just 16 bytes. Wouldn't it be nice to avoid additional GC overhead?

rueian:
Hi @mdigger,

Thank you very much for your suggestions. I have made Completed and Cacheable public in v0.0.99.

Though, I forget to pick your above commit into the latest version. Would you mind rebasing it and I pick it into the next release?

mdigger:
Thanks! There is no need.

PR #164, opened 2022-12-30
feat(message): implement `AsScanEntry`

ken8203:
Add `AsScanEntry` since there is no message function for SCAN series command, people need to parse cursor and elements by themself.

### DIFF
--- a/message.go
+++ b/message.go
@@ -325,6 +325,16 @@ func (r RedisResult) AsIntMap() (v map[string]int64, err error) {
 	return
 }
 
+// AsScanEntry delegates to RedisMessage.AsScanEntry.
+func (r RedisResult) AsScanEntry() (v ScanEntry, err error) {
+	if r.err != nil {
+		err = r.err
+	} else {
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -717,6 +727,35 @@ func (m *RedisMessage) AsZScores() ([]ZScore, error) {
...
+	Elements []string
+}
+
+// AsScanEntry check if message is a redis array/set response of length 2, and convert to ScanEntry.
+func (m *RedisMessage) AsScanEntry() (ScanEntry, error) {
+	msgs, err := m.ToArray()
+	if err != nil {
+		return ScanEntry{}, err
+	}
+	if len(msgs) != 2 {
```
rueian, line 742: Hi @ken8203,

Thanks for the contribution.

```suggestion
	if len(msgs) < 2 {
```

I think it it better to relax this length check. It is not uncommon for redis adding new element to the array in newer version.

ken8203: Sure, it makes sense.

PR #200, opened 2023-04-24
add ClientOption.DialFn to manually create TCP/TLS connections

davestibrany-dd:
Allows for the ability to add a custom dial function as a ClientOption. This is useful for getting metrics on dials, such as dial latency, dial errors/successes, etc..


rueian:
Hi @davestibrany-dd, many thanks. It looks good to me.

PR #213, opened 2023-05-02
fix: hook README

hogan-chu:
minor fix for hook README 

@rueian Thanks


rueian:
Hi @hogan-chu,

Thanks!

PR #214, opened 2023-05-02
feat: NewSimpleCacheAdapter converts a SimpleCache into CacheStore

rueian:
Provide `NewSimpleCacheAdapter` function converting a `SimpleCache`, which is just a KV store interface, into complex `CacheStore`

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -2,6 +2,7 @@ package rueidis
 
 import (
 	"context"
+	"sync"
 	"time"
 )
 
@@ -20,9 +21,9 @@ type CacheStoreOption struct {
 type CacheStore interface {
 	// Flight is called when DoCache and DoMultiCache, with the requested client side ttl and the current time.
 	// It should look up the store in single-flight manner and return one of the following three combinations:
-	// Case 1: (empty RedisMessage, nil CacheEntry)         <- when cache missed, and rueidis will send the request to redis.
-	// Case 2: (empty RedisMessage, non-nil CacheEntry)     <- when cache missed, and rueidis will use CacheEntry.Wait to wait for response.
-	// Case 3: (non-empty RedisMessage, non-nil CacheEntry) <- when cache hit
+	// Case 1: (empty RedisMessage, nil CacheEntry)     <- when cache missed, and rueidis will send the request to redis.
+	// Case 2: (empty RedisMessage, non-nil CacheEntry) <- when cache missed, and rueidis will use CacheEntry.Wait to wait for response.
+	// Case 3: (non-empty RedisMessage, nil CacheEntry) <- when cache hit
 	Flight(key, cmd string, ttl time.Duration, now time.Time) (v RedisMessage, e CacheEntry)
 	// Update is called when receiving the response of the request sent by the above Flight Case 1 from redis.
 	// It should not only update the store but also deliver the response to all CacheEntry.Wait and return a desired client side PXAT of the response.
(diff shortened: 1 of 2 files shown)


cache.go
```diff
@@ -44,3 +45,110 @@ type CacheStore interface {
 type CacheEntry interface {
 	Wait(ctx context.Context) (RedisMessage, error)
 }
+
+// SimpleCache is an alternative interface can be paired with NewSimpleCacheAdapter to construct a CacheStore
+type SimpleCache interface {
+	Get(key, cmd string, now time.Time) RedisMessage
```
Yiling-J, line 51: I think if key and cmd always come together, then joining them to a single key in Flight method is enough? This makes Get method simpler: `Get(key string, now time.Time)`. Another question is why `now` is a required param? It would be even better if the now param can be removed

rueian: Hi @Yiling-J,

When doing `Get(key, cmd, now)`, the `key` and `cmd` should both be considered to fetch an entry, however, when doing `Del(key)` all entries under the same `key` should be deleted. That's why they are separated on the interface.

The `now` parameter, provided from the upper layer to specify the current time, has several benefits:
1. Allow consistent TTL checks.
2. Avoid unnecessary `time.Now()` calls.
3. Easy for writing unit tests

Yiling-J: I think adapter will take care of all these redis related things? Which means adapter's Flight will group cmd and key first then call store's Get and adapter's Delete method will call store's Delete use key only. Store is just a simple KV store, doesn't know anything about redis. 
But I have another question, if key and cmd are used together to get from store, for example: `GET foo`, then the cache key would be "GETfoo"? In this way how Delete works, delete "foo" won't delete the entry because key is "GETfoo".

rueian: @Yiling-J 

In the scenario you mentioned, the adapter will call `store.Get("foo", "GET", now)`. It won't group cmd and key first.
Whether to group them or not is up to the store implementation.

Yiling-J: My suggestion is let the adapter handle all these things, even the ttl can be checked at adapter level and keep store simple. This is just a recommendation. I'm not sure how other redis clients implement client-side caching, maybe take a look theirs is better.

rueian: Yes, it makes sense. So here you go. The latest SimpleStore is just a simple KV store.

Yiling-J: I can't say anyhing because I haven't try yet, though your PR looks good! I will try today or tomorrow to see if it works as my expectation. But due to the fact that I don't work in any company, I can't confirm if it works well in any production environment

Yiling-J: @rueian I integrate Rueidis to my cache benchmark and I thinks it works, I only test Get and Set methods. One thing seems a little weird is that I have to change my cache from `Theine[string, string]` to `Theine[string, rueidis.RedisMessage]`, though value is still string, but I think because Rueidis cache commands, not simple key/value pairs, this is necessary.

Yiling-J: Another interesting observation is that memory cache hit ratio drops in all my benchmarks. I think this is because Rueidis cache on Get not Set, so data will be sync to client cache only after a Get command, this affect access pattern of my benchmark workloads and cause hit ratio drop. If Set can store data to client cache simultaneously the ratio may be same. But considering Rueidis cache commands, maybe it's not doable

rueian: Glad to know that it works.

Yes, I believe it is necessary too. RedisMessage represents all types of redis response, not only string but also integer, array, map, nil and other types.

Yiling-J: @rueian I have a question after reading the official Client-side caching guide. I think in that documention client should cache key/value, not commands? Example from doc:
```
This is an example of the protocol:

- Client 1 -> Server: CLIENT TRACKING ON
- Client 1 -> Server: GET foo
- (The server remembers that Client 1 may have the key "foo" cached)
- (Client 1 may remember the value of "foo" inside its local memory)
- Client 2 -> Server: SET foo SomeOtherValue
- Server -> Client 1: INVALIDATE "foo"
```
If key in server is `foo` but key in client is `fooGET`, how INVALIDATE will work?

rueian: In the above case, what client 1 remembers is the response of "GET foo". Therefore when the server notifies the client that the key "foo" is changed, the client knows that it should invalidate "fooGET".

Consider the following case:
```
- Client 1 -> Server: CLIENT TRACKING ON
- Client 1 -> Server: HGET foo f1
- Client 1 -> Server: HGET foo f2
- (The server remembers that Client 1 may have the key "foo" cached)
- (Client 1 may remember the value of the f1 and f2 fields of "foo" inside its local memory)
- Client 2 -> Server: DEL foo
- Server -> Client 1: INVALIDATE "foo"
```

In this case, what the client remembers are the responses of "HGET foo f1" and "HGET foo f2". Therefore when the server notifies the client that the key "foo" is changed, the client knows that it should invalidate "fooHGETf1" and "fooHGETf2".

Yiling-J: I think Redis should add more details and implentmention guide of different data types to their Client-side caching documentation. Then client authors won't be confused about how to implentment. For example I would try to match redis data structure instead of using complicated key pattern:
```go
type RedisData struct {
	value   string
	list    []string
	hashmap map[string]string
}
```
If it's a hash or set, put field in hashmap, if list, put into list. And local cache is a map of key/RedisData. Now that Rueidis is an offical Redis project, maybe you can suggest them to improve their Client-side caching specification

rueian: It is actually not very complicated. The client just need to remember what key the cached responses belongs to and invalidate them when being notified.

PR #217, opened 2023-05-03
Release drafter

chayim:
Adding the release drafter, as used by other redis clients


rueian:
Thanks, @chayim, LGTM!

Issue #239, opened 2023-05-16
Issue with using in-memory cache per TCP connection (wire) 


smf8:
Hi, Thank you for the great library. I want to start using this library as Redis client mainly for the client-side caching feature. As I've seen in the code, `CacheStore` is a field in `pipe` struct, and `pipe` is created per multiplex pipeline.

This will work fine for almost every scenario since each key will be consistently sent through the same pipe and thus will hit the same cache.
But what for an `MGet` with the **same keys but in different order**?

Example:
```go
// Save key1 and key2 in redis
queryA := client.DoCache(context.Background(), client.B().Mget().Key("key1", "key2").Cache(), time.Hour)

queryB := client.DoCache(context.Background(), client.B().Mget().Key("key1").Cache(), time.Hour)

queryC := client.DoCache(context.Background(), client.B().Mget().Key("key2").Cache(), time.Hour)

fmt.Printf("Cache hit result:\nA:%v\nB:%v\nC:%v\n", queryA.IsCacheHit(), queryB.IsCacheHit(), queryC.IsCacheHit())
```

I expect the following code to print: 
```text
Cache hit result:
A:false
B:true
C:true 
```
But in the actual result, I'll get **C:false**. As I've investigated. This is due to Mget using only the first key to calculate `slot`. and **if** `key2`'s slot is different from `key1`, this causes the library to use another `wire` for `queryC` and it will miss the cache although it's already present in the client application.

I wanted to ask you to confirm whether this might be the case and if you propose any workaround.

rueian:
Hi @smf8,

Thank you for looking at the code so deeply and your concern is totally correct.

I think a workaround can be made in the `MGetCache` helper, by just removing these lines: https://github.com/redis/rueidis/blob/7353e522a67befe77f233af66fdc4132b5e7bb06/helper.go#L18-L20
to split the request into multiple MGET by slots.

But for using `client.B().Mget().Key("key1", "key2").Cache()` directly, I haven't come up with a good solution except for a custom `CacheStore` implementation.

rueian:
Also note that `MGET` across multiple slots is actually not recommended since it will be rejected by redis cluster.

rueian:
Yes, using `parallelMget` for all clients is preferable way to client side caching for `MGET` and `JSON.MGET`. Actually the current usages of `clientMGetCache` should be considered bugs. So, your PR is very welcome or I may do it on this weekend. 

[...]

PR #241, opened 2023-05-17
Use ParallelMGet for single client in MGetCache helper function

smf8:
Regarding #239, I think using `parallelMget` for cached MGet helper functions will solve the issue. please let me know if anything else should be done.


rueian:
Hi @smf8, many thanks! It looks good to me.

Issue #132, opened 2022-11-14
Readonly command on Replica servers only


Xanvial:
Hello,
I'm currently starting to implement this library on our system. Using default cluster implementation with 3 masters and 3 replicas. One of the main things that I want to implement is using only replica servers for all readonly command. Is this supported in the library? I can't seem to find it in the documentation/code.

Not sure about the implementation, but on goredis this can be enabled by setting up a flag on goredis initialization https://github.com/go-redis/redis/blob/master/cluster.go#L40

rueian:
Hi @Xanvial,

Unfortunately, it doesn’t read from the replicas currently. Would you consider using Client-Side Caching instead?

Reading from replicas might make implementation be extremely complex. Currently, I don’t have idea about how to implement it.

Xanvial:
I see, well I haven't read and deep dive this library yet, but on other like goredis, the rough flow of replica readonly is like this
1. Get cluster slot/shard data
2. Assign list of node as array for each slot, usually master is first index, and the replicas will use the rest (in case there's more than one replica)
3. When there's a request that marked as readonly, it will choose one of the replica of related slot
4. On choosen node, call `READONLY` command (only needed once as long as the connection is not closed)
5. Do the command as usual

Well, like mentioned before, I have no clue how hard it's to implement in here. But this feature can help when there's a high RPS for readonly commands, we can scale the system up just by adding a new replica node so the network traffic can be distributed.

rueian:
Thank you for the detailed procedure. One more thing that also need to be considered is cluster topology migration. Procedures may become more complicated during that period.

In terms of high RPS, I believe client side caching is much better than adding more redis replicas. I would  recommend you try that.

Xanvial:
yes I agree, I also want to try this lib for the client side caching feature. But the readonly feature also helping to split the task, so replicas will only responsible for readonly command, while master node only doing Write command only

jashakimov:
Hi @rueian 
Tell me please, what will happen if we deploy app with client side cache? There are will be deleted?

smf8:
Hello @rueian

I wanted to test the read from replica feature in my **multi-node sentinel Redis** setup (I don't have much experience with Redis cluster). So I'd like to try implementing a solution for `sentinelClient` to support `ReplicaReadOnly` option. It would be great to have feedback on the design.

[...]

PR #248, opened 2023-05-19
Add Sentinel readonly replica support

smf8:
Hi, This is my suggested implementation for a read-only replica client (#132)  for **sentinel** Redis setup. Please share your thoughts on this implementation. If it's OK with you I can write the unit tests to cover what's changed.

### DIFF
--- a/internal/cmds/gen_sentinel.go
+++ b/internal/cmds/gen_sentinel.go
@@ -2,26 +2,6 @@
 
 package cmds
 
-type SentinelReplicas Completed
-
-func (b Builder) SentinelReplicas() (c SentinelReplicas) {
-	c = SentinelReplicas{cs: get(), ks: b.ks}
-	c.cs.s = append(c.cs.s, "SENTINEL", "REPLICAS")
-	return c
-}
-
-func (c SentinelReplicas) Master(master string) SentinelReplicasMaster {
-	c.cs.s = append(c.cs.s, master)
-	return (SentinelReplicasMaster)(c)
-}
-
-type SentinelReplicasMaster Completed
-
-func (c SentinelReplicasMaster) Build() Completed {
-	c.cs.Build()
-	return Completed(c)
-}
-
 type SentinelFailover Completed
 
 func (b Builder) SentinelFailover() (c SentinelFailover) {
@@ -58,6 +38,26 @@ func (c SentinelGetMasterAddrByName) Master(master string) SentinelGetMasterAddr
 type SentinelGetMasterAddrByNameMaster Completed
 
 func (c SentinelGetMasterAddrByNameMaster) Build() Completed {
+	c.cs.Build()
+	return Completed(c)
+}
+
+type SentinelReplicas Completed
+
+func (b Builder) SentinelReplicas() (c SentinelReplicas) {
+	c = SentinelReplicas{cs: get(), ks: b.ks}
+	c.cs.s = append(c.cs.s, "SENTINEL", "REPLICAS")
+	return c
+}
+
+func (c SentinelReplicas) Master(master string) SentinelReplicasMaster {
+	c.cs.s = append(c.cs.s, master)
+	return (SentinelReplicasMaster)(c)
+}
+
+type SentinelReplicasMaster Completed
+
+func (c SentinelReplicasMaster) Build() Completed {
 	c.cs.Build()
 	return Completed(c)
 }
--- a/internal/cmds/gen_sentinel.go
+++ b/internal/cmds/gen_sentinel.go
@@ -2,6 +2,26 @@
 
 package cmds
 
+type SentinelReplicas Completed
+
+func (b Builder) SentinelReplicas() (c SentinelReplicas) {
+	c = SentinelReplicas{cs: get(), ks: b.ks}
+	c.cs.s = append(c.cs.s, "SENTINEL", "REPLICAS")
+	return c
+}
+
+func (c SentinelReplicas) Master(master string) SentinelReplicasMaster {
+	c.cs.s = append(c.cs.s, master)
+	return (SentinelReplicasMaster)(c)
+}
+
+type SentinelReplicasMaster Completed
+
+func (c SentinelReplicasMaster) Build() Completed {
+	c.cs.Build()
+	return Completed(c)
+}
+
 type SentinelFailover Completed
 
 func (b Builder) SentinelFailover() (c SentinelFailover) {
--- a/rueidis.go
+++ b/rueidis.go
@@ -136,9 +136,9 @@ type ClientOption struct {
 	// AlwaysRESP2 makes rueidis.Client always uses RESP2, otherwise it will try using RESP3 first.
 	AlwaysRESP2 bool
 
-	// ReadOnlyReplica indicates that this client will only try to connect to readonly replicas of redis setup.
+	// ReplicaOnly indicates that this client will only try to connect to readonly replicas of redis setup.
 	// currently, it is only implemented for sentinel client
-	ReadOnlyReplica bool
+	ReplicaOnly bool
 }
 
 // SentinelOption contains MasterSet,
--- a/rueidis.go
+++ b/rueidis.go
@@ -135,6 +135,10 @@ type ClientOption struct {
 	AlwaysPipelining bool
 	// AlwaysRESP2 makes rueidis.Client always uses RESP2, otherwise it will try using RESP3 first.
 	AlwaysRESP2 bool
+
+	// ReadOnlyReplica indicates that this client will only try to connect to readonly replicas of redis setup.
+	// currently, it is only implemented for sentinel client
+	ReadOnlyReplica bool
 }
 
 // SentinelOption contains MasterSet,
--- a/sentinel.go
+++ b/sentinel.go
@@ -334,9 +334,11 @@ func (c *sentinelClient) listWatch(cc conn) (serverAddress string, sentinels []s
 					c.switchMasterRetry(fmt.Sprintf("%s:%s", m[3], m[4]))
 				}
 			case "+reboot":
-				m := strings.SplitN(event.Message, " ", 4)
+				m := strings.SplitN(event.Message, " ", 7)
 				if m[0] == "master" && m[1] == c.sOpt.Sentinel.MasterSet {
 					c.switchMasterRetry(fmt.Sprintf("%s:%s", m[2], m[3]))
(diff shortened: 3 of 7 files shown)


sentinel.go
```diff
@@ -289,19 +305,75 @@ func (c *sentinelClient) _refresh() (err error) {
...
+	// otherwise send master as addrress
+	m, err := resp[1].AsStrSlice()
+	if err != nil {
+		return "", nil, err
+	}
+	return fmt.Sprintf("%s:%s", m[0], m[1]), sentinels, nil
+}
+
+// watchEvents will help subscribing to channel ONLY ONCE.
+func (c *sentinelClient) watchEvents(ctx context.Context, cc conn) {
```
smf8, line 371: This is intentional because in the previous implementation, every time a message is received, eventually it will cause another launch of `cc.Receive(...)` which then will lead to multiple operations for a single event.

rueian: Hi @smf8,

Since you mentioned another issue here, could we deal with it first and in a separate PR?

I just looked back into the `refresh` and `listWatch`, and I think they would only cause duplicated `cc.Receive(...)` when `_switchMaster` failed and we can quickly fix the duplication by always closing the previous `sConn` in the `_refresh`.

rueian: Sorry, we need to send UNSUBSCRIBE first before we close the previous `sConn` so that the previous `cc.Receive(...)` can exit without error.

rueian:
Hi @smf8,

Correct me if I am wrong. So the `ReadOnlyReplica` will restrict the client to only connect to replicas. How can this be useful?

smf8:
Yes exactly.
This can provide the option for clients to define separate instances of `rueidis.Client` for read/write operations. 
I thought about using `[]conn` in client but It will introduce many complexities with not much gain in my opinion.

sentinel.go
smf8: How about only sending an `UNSUBSCRIBE`  command before starting `cc.Receive` in the main goroutine, if there wasn't any previous subscription it will execute without doing anything. So there won't be any need to close the connection.

rueian: Yes, good idea.

smf8: #249

internal/cmds/gen_sentinel.go
```diff
@@ -2,6 +2,26 @@
 
 package cmds
 
+type SentinelReplicas Completed
```
rueian, line 5: These `Completed` types are generated by `go run gen.go` from https://github.com/redis/rueidis/blob/main/hack/cmds/commands_sentinel.json.

The JSON file should be updated as well.

rueidis.go
```diff
@@ -135,6 +135,10 @@ type ClientOption struct {
 	AlwaysPipelining bool
 	// AlwaysRESP2 makes rueidis.Client always uses RESP2, otherwise it will try using RESP3 first.
 	AlwaysRESP2 bool
+
+	// ReadOnlyReplica indicates that this client will only try to connect to readonly replicas of redis setup.
+	// currently, it is only implemented for sentinel client
+	ReadOnlyReplica bool
```
rueian, line 141: Instead of `ReadOnlyReplica`, I think `ReplicasOnly` would be better.

sentinel.go
```diff
@@ -36,19 +38,21 @@ func newSentinelClient(opt *ClientOption, connFn connFn) (client *sentinelClient
...
+	sentinels    *list.List
+	mAddr        string
+	sAddr        string
+	sc           call
+	mu           sync.Mutex
+	stop         uint32
+	cmd          cmds.Builder
+	retry        bool
+	readonly     bool
+	watchCounter atomic.Int32
```
rueian, line 55: We don't need the `watchCounter` now, right?

sentinel.go
```diff
@@ -319,13 +339,28 @@ func (c *sentinelClient) listWatch(cc conn) (master string, sentinels []string,
...
 	}(cc)
 
-	resp := cc.DoMulti(ctx, sentinelsCMD, getMasterCMD)
+	commands := []Completed{
+		sentinelsCMD,
+		getMasterCMD,
+	}
+
+	if c.readonly {
+		commands = append(commands, replicasCMD)
```
rueian, line 360: It is ok just always send the `replicasCMD`.

sentinel.go
```diff
@@ -319,13 +339,28 @@ func (c *sentinelClient) listWatch(cc conn) (master string, sentinels []string,
 				if m[0] == "master" && m[1] == c.sOpt.Sentinel.MasterSet {
 					c.switchMasterRetry(fmt.Sprintf("%s:%s", m[2], m[3]))
 				}
+			case "+slave":
```
rueian, line 342: Should we subscribe to `-slave` as well?

rueian:
> Yes exactly. This can provide the option for clients to define separate instances of `rueidis.Client` for read/write operations.

Yes, indeed there might be some use case for reading only from replicas.

However, this PR can't actually work yet, right? Since we still need to set the underlying connection in the `READONLY` state, otherwise all requests will be rejected by replicas.

internal/cmds/gen_sentinel.go
smf8: Sorry, I missed that. Done.

sentinel.go
smf8: I couldn't find this channel in official sentinel documentation ([here](https://redis.io/docs/management/sentinel/#pubsub-messages)). Also removing a slave replica locally doesn't seem to trigger it. Although `+sdown` and `-sdown` is triggered when a replica (or master) is restarted and we can capture it with condition `m[0] == "slave"`.

rueian: I am sorry about the misinformation. What about failover-end?

smf8: I think the only condition that's not covered by current channels is the condition of `slave` shutting down. master restart/shutdowns are covered with `+switch-master` and `+reboot`. Also `sdown` status of replicas isn't reported by `failover-end` as I've tested. 

What do you think about the changes proposed in ab120af1ad7695448e6bc7e718f9a045820da152 and 37741dd8065a9601b4c33c5209ff70545d4b9bb4?

Feel free to check your other comments and check if they're resolved.

smf8:
> However, this PR can't actually work yet, right

I think aside from some unit/integration tests to cover all edge cases. It's ready to use. Since this option is only for `sentinel` clients and with a sentinel setup, all slaves are started with the `replica-of` config enabled. There is no need to mark the connection as `READONLY`. 

Also since the user needs to **explicitly** mark the client as `ReplicaOnly`, receiving errors on write commands will be expected behavior.

sentinel.go
rueian: Monitoring `+sdown` and `-sdown` looks good to me.

Just one concern about checking the existence of `s-down-time`: Will it be clear once the replica is alive?

Besides that, I think the way you `delete response i` may panic: https://go.dev/play/p/zWeXCrTNqwJ

smf8: Thanks for the precise review :)

Sorry, I've updated the method to avoid panic.

Also `s-down-time` seems to disappear once the replica is live and can reply to sentinel `PING`s. I couldn't find any documentation on it to confirm but I've tested it with multiple scenarios and it looks like to work that way.

sentinel.go
```diff
@@ -335,6 +370,41 @@ func (c *sentinelClient) listWatch(cc conn) (master string, sentinels []string,
...
+		}
+
+		if len(eligibleReplicas) == 0 {
+			return "", nil, fmt.Errorf("not enough ready replicas")
+		}
+
+		// choose a replica randomly
+		randomReplicaIndex := rand.Intn(len(eligibleReplicas))
+		m, err := eligibleReplicas[randomReplicaIndex].AsStrMap()
+		if err != nil {
```
rueian, line 400: We can skip the error checking here, since we have checked them above (if we also use AsStrMap above).

sentinel.go
```diff
@@ -335,6 +370,41 @@ func (c *sentinelClient) listWatch(cc conn) (master string, sentinels []string,
...
+	if c.readonly {
+		replicaResponses, err := resp[1].ToArray()
+		if err != nil {
+			return "", nil, err
+		}
+
+		eligibleReplicas := make([]RedisMessage, 0, len(replicaResponses))
+		// eliminate replicas with s_down condition
+		for i := range replicaResponses {
+			replicaResponseMap, err := replicaResponses[i].AsMap()
```
rueian, line 384: AsMap can be changed to AsStrMap to align with the below one.

rueian:
Hey @smf8, I have tried this PR manually and it worked well.

So one thing left is about testing, could you try to make test coverage of this patch 100%?

smf8:
That's awesome 🎉🎉. 

Sure, I'll try updating the tests to cover all the changes tomorrow. let me know if you can think of anything else for this patch.

sentinel.go
```diff
@@ -319,13 +338,29 @@ func (c *sentinelClient) listWatch(cc conn) (master string, sentinels []string,
 				if m[0] == "master" && m[1] == c.sOpt.Sentinel.MasterSet {
 					c.switchMasterRetry(fmt.Sprintf("%s:%s", m[2], m[3]))
 				}
+			case "+slave", "+sdown", "-sdown":
+				m := strings.SplitN(event.Message, " ", 4)
+				if m[0] == "slave" {
```
rueian, line 343: We might also need to check `c.readonly` here as well.

smf8: I think beside `c.readonly` we need to check `c.sOpt.Sentinel.MasterSet` in these events. because other master set events can trigger this client's refresh with current implementation.

smf8:
Hey @rueian, I've updated the tests to cover the changes in this PR. Let me know if I've missed anything.

sentinel_test.go
```diff
@@ -162,6 +162,229 @@ func TestSentinelClientInit(t *testing.T) {
...
+							{typ: '+', string: "ip"}, {typ: '+', string: ""},
+							{typ: '+', string: "port"}, {typ: '+', string: "4"},
+						}},
+					}}},
+					{val: RedisMessage{typ: '*', values: []RedisMessage{
+						RedisMessage(*Nil),
+						//{typ: '%', values: []RedisMessage{
+						//	RedisMessage(RedisError{}),
+						//	RedisMessage(RedisError{}),
+						//}},
```
rueian, line 280: Hi @smf8,

Just removing these comments and then I think this is good to merge. Thanks!

rueian: Yes, they look good.

PR #258, opened 2023-05-24
fix(rueidisotel): prevent panic-able do multi with zero length commands

jinwoo1225:
Dear rueidis contributors,

First of all, I would like to express my gratitude for the wonderful rueidis client. It has been an invaluable tool in my projects, and I truly appreciate the effort and dedication that have gone into its development.

I Just find out in `rueidisotel` package has a possible panic when client.DoMulti with empty commands. 

In [multiFirst](https://github.com/redis/rueidis/blob/ffd8e8e9da0772279d2368a041f7090747ecce9d/rueidisotel/trace.go#L216-L235) and [multiCacheableFirst](https://github.com/redis/rueidis/blob/ffd8e8e9da0772279d2368a041f7090747ecce9d/rueidisotel/trace.go#L237-L256) function has `strings.Builder{}.Grow()` method which will panic with negative value is given it's parameter.

So I added if statement in start of the DoMulti, DoMultiCache function with checking length of commands is over 0.

Thank you for considering my request, and I eagerly await your response.

Best regards,
@jinwoo1225


rueian:
Hi @jinwoo1225,
Many thanks for notice the edge case. I just wondering whether returning directly is a desired solution for this case, since it will cause the operation not to be recorded.

I am much expecting users of otel would like to see every operations even if they are empty.

jinwoo1225:
Oh, I get it. 

How about returning empty string on multi start function? 

I think it will create span for empty command.

jinwoo1225:
Oh my,, this test seems flaky.. I'll fix it

jinwoo1225:
just found out `otel.SetTracerProvider` on test is setting tracer in global variable. which delegates tracer to new tracerProvider with `sync.Once` :sob:. 

[link to code snippet](https://github.com/open-telemetry/opentelemetry-go/blob/46f2ce5ca6adaa264c37cdbba251c9184a06ed7f/internal/global/state.go#L72-L76)

So setting tracer provider twice doesn't work. 

We can have two choices. 

1. I can set the exporter to global variable and setting it in [TestMain](https://pkg.go.dev/testing#hdr-Main) with `tracetest.NewInMemoryExporter()` function
2. or keeping empty command test on `TestWithClient` like 301520e 🤔 

I respect all of your judgement!

rueian:
In terms of the `otel.SetTracerProvider` problem in the current tests, we probably should wait for this https://github.com/redis/rueidis/pull/259 to be merged first.

rueian:
Hey @jinwoo1225, jus to notify you the #259 is merged, and there is conflict you should resolve.

jinwoo1225:
just passed CI!

rueian:
Thanks @jinwoo1225!

Issue #257, opened 2023-05-24
Global state


ash2k:
https://github.com/redis/rueidis/blob/ffd8e8e9da0772279d2368a041f7090747ecce9d/rueidisotel/trace.go#L21-L26

It's unfortunate that instrumentation relies on global OTEL objects. I use non-global ones in my project and cannot really configure this library to use my own trace provider, etc. Please consider removing global objects and making it possible to configure what's used. It's fine to use the global ones by default, but if user provides tracer provider/etc explicitly, then global OTEL factories should not be used.

Another use case is if I want to have more than one client and instrument them with different trace providers.

rueian:
Hi @ash2k,

I think it would be nice to have the ability for changing the provider. What kind of API for that you would like to see?

ash2k:
Just two `Option`s - allow to specify a `MeterProvider` and `TraceProvider` in `WithClient()`:

https://github.com/redis/rueidis/blob/ffd8e8e9da0772279d2368a041f7090747ecce9d/rueidisotel/trace.go#L32

I think that's all that is needed. If those options are not specified then the code can use the global ones.

ash2k:
This is great, thank you!

PR #259, opened 2023-05-24
feat: add traceProvider && meterProvider option

HoganEdwardChu:
@rueian  maybe this can be option? Thanks

### DIFF
--- a/rueidisotel/trace.go
+++ b/rueidisotel/trace.go
@@ -2,6 +2,7 @@ package rueidisotel
 
 import (
 	"context"
+	"go.opentelemetry.io/otel/codes"
 	"strings"
 	"time"
 
@@ -9,7 +10,6 @@ import (
 	"github.com/redis/rueidis/internal/cmds"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 )
@@ -320,24 +320,26 @@ func multiCacheableFirst(multi []rueidis.CacheableTTL) string {
 }
 
 func (o *otelclient) start(ctx context.Context, op string, size int, attrs []attribute.KeyValue) (context.Context, trace.Span) {
-	return o.tracer.Start(ctx, op, kind, attr(op, size), trace.WithAttributes(attrs...))
+	return startSpan(o.tracer, ctx, op, size, attrs...)
 }
 
 func (o *otelclient) end(span trace.Span, err error) {
-	if err != nil && !rueidis.IsRedisNil(err) {
-		span.RecordError(err)
-		span.SetStatus(codes.Error, err.Error())
-	} else {
-		span.SetStatus(codes.Ok, "")
-	}
-	span.End()
+	endSpan(span, err)
 }
 
 func (d *dedicated) start(ctx context.Context, op string, size int, attrs []attribute.KeyValue) (context.Context, trace.Span) {
-	return d.tracer.Start(ctx, op, kind, attr(op, size), trace.WithAttributes(attrs...))
+	return startSpan(d.tracer, ctx, op, size, attrs...)
 }
 
 func (d *dedicated) end(span trace.Span, err error) {
+	endSpan(span, err)
+}
+
+func startSpan(tracer trace.Tracer, ctx context.Context, op string, size int, attrs ...attribute.KeyValue) (context.Context, trace.Span) {
+	return tracer.Start(ctx, op, kind, attr(op, size), trace.WithAttributes(attrs...))
+}
+
+func endSpan(span trace.Span, err error) {
 	if err != nil && !rueidis.IsRedisNil(err) {
 		span.RecordError(err)
 		span.SetStatus(codes.Error, err.Error())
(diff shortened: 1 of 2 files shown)


rueian:
Hey @HoganEdwardChu,

Thank you for your quick support for https://github.com/redis/rueidis/issues/257.

rueidisotel/trace.go
```diff
@@ -27,12 +27,25 @@ var (
...
+		opt(o)
+	}
+
+	// Set the MeterProvider and TracerProvider if specified
+	if o.meterProvider != nil {
+		otel.SetMeterProvider(o.meterProvider)
+	}
+	if o.tracerProvider != nil {
+		otel.SetTracerProvider(o.tracerProvider)
+	}
```
rueian, line 47: These will set the global providers. We should not do that.
We should instead do this after applying `opts`:
```go
	if o.meterProvider == nil {
		o.meterProvider = otel.GetMeterProvider()
	}
	if o.tracerProvider == nil {
		o. tracerProvider  = otel.GetTracerProvider()
	}
```

rueidisotel/trace.go
```diff
@@ -53,10 +66,26 @@ func TraceAttrs(attrs ...attribute.KeyValue) Option {
...
+
 type otelclient struct {
-	client rueidis.Client
-	mAttrs []attribute.KeyValue
-	tAttrs []attribute.KeyValue
+	client         rueidis.Client
+	mAttrs         []attribute.KeyValue
+	tAttrs         []attribute.KeyValue
+	meterProvider  metric.MeterProvider
+	tracerProvider trace.TracerProvider
```
rueian, line 88: Besides the providers, we should also move `tracer`, `cscMiss` and `cscHits` into this struct.

HoganEdwardChu: ah... got it. I trying to add test codes and i faced the problem what you said

rueidisotel/trace.go
HoganEdwardChu: sorry for late you mean https://github.com/redis/rueidis/blob/1defd2dba8930c58d3f790ca89381ba370fd4111/rueidisotel/trace.go#L55 ?

rueidisotel/trace.go
```diff
@@ -27,12 +27,26 @@ var (
...
+	}
+	if o.tracerProvider == nil {
+		o.tracerProvider = otel.GetTracerProvider() // Default to global TracerProvider
+	}
+	// Now that we have the meterProvider and tracerProvider, get the Meter and Tracer
+	o.meter = o.meterProvider.Meter(name)
+	o.tracer = o.tracerProvider.Tracer(name)
+	// Now create the counters using the meter
+	o.cscMiss, _ = o.meter.Int64Counter("rueidis_do_cache_miss")
+	o.cscHits, _ = o.meter.Int64Counter("rueidis_do_cache_hits")
```
rueian, line 49: So, after moving the tracer and counters to here, the original global ones should be removed.

HoganEdwardChu: got it

rueidisotel/trace.go
```diff
@@ -17,93 +17,122 @@ import (
...
+func WithMetricAttrs(attrs ...attribute.KeyValue) Option {
 	return func(o *otelclient) {
 		o.mAttrs = attrs
 	}
 }
 
-// TraceAttrs set additional attributes to append to each trace.
-func TraceAttrs(attrs ...attribute.KeyValue) Option {
+// WithTraceAttrs set additional attributes to append to each trace.
+func WithTraceAttrs(attrs ...attribute.KeyValue) Option {
```
rueian, line 59: Oh, this is a breaking change. We can’t do that directly. We have at least to maintain an alias to the original function name.

HoganEdwardChu: Got it

rueidisotel/trace.go
rueian: Hi @HoganEdwardChu,

I just looked into these tests. They all failed with nil dereferencing on the `tracer` field. So you need to add them to all `&otelclient{...}` and `&dedicated{...}`

HoganEdwardChu: Hi @rueian  i added it in https://github.com/redis/rueidis/blob/d29aa821a42d9a5affbe0a2217bdaa99fe715a5c/rueidisotel/trace.go#L41 and https://github.com/redis/rueidis/blob/d29aa821a42d9a5affbe0a2217bdaa99fe715a5c/rueidisotel/trace.go#L85 you mean https://github.com/redis/rueidis/blob/d29aa821a42d9a5affbe0a2217bdaa99fe715a5c/rueidisotel/trace.go#L171 here?

rueidisotel/trace.go
```diff
@@ -2,37 +2,46 @@ package rueidisotel
 
 import (
 	"context"
+	"go.opentelemetry.io/otel/codes"
```
rueian, line 5: This should be in the below section

HoganEdwardChu:
It seem's done. Now we have options for providerThanks! @rueian 
```golang
	client = WithClient(
		client,
		TraceAttrs(attribute.String("any", "label")),
		MetricAttrs(attribute.String("any", "label")),
		WithTracerProvider(tracerProvider),
		WithMeterProvider(meterProvider),
	)
```

rueian:
Yeah! Thank you very much @HoganEdwardChu !

PR #262, opened 2023-05-26
fix(message): use errors is with redis nil

jinwoo1225:
This pull request aims to enhance error handling by utilizing errors.Is and errors.As.

Just found out, the IsRedisNil function fails to capture wrapped errors.

### DIFF
--- a/message.go
+++ b/message.go
@@ -2,6 +2,7 @@ package rueidis
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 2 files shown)


rueian:
Thanks @jinwoo1225, but that was actually intended. I don't think they should return true when the err is modified.

jinwoo1225:
Oh I see! I get it. I will close this pull request.

rueian:
Hey @jinwoo1225, I think it would be nice if we can have tests to make sure that they return false instead. Would you be willing to turn your tests to the opposite side?

jinwoo1225:
@rueian you mean create a test case with wrapped error in isRedisNil function?

rueian:
Yes. And IsRedisErr as well. To make sure they return false if the err is a wrapped Nil.

Issue #271, opened 2023-06-03
Confusing Type Error Messages When Using As* and To* methods on RedisMessage and RedisResult


kevinxmorales:
When using RedisMessage methods like AsMap or ToArray, there can be confusing error messages that result from a no-op type conversion. 
For example, rueidis does not have a way for me to convert a geosearch query into a GeoLocation struct the way go-redis does, but I figured I might be able to get a map with the coords, dist, and hash. So I tried the following
```golang
resp = client.Do(context.Background(), client.B().
	Geosearch().
	Key("cities").
	Fromlonlat(77.6088, 43.1566).
	Bybox(500.0).
	Height(500.0).
	Mi().
	Withcoord().Withdist().Withhash().
	Build())
geoArr, err := resp.ToArray()
if err != nil {
        return err
}
for _, v := range geoArr {
	x, err := v.AsMap()
	if err != nil {
		log.Fatal(err)
	}
}
``` 
Trying to run this results in a panic with an error message 
```
panic: redis message type : as map key is not supported
```

I understand the need to reject bad type conversions, however error messages that use symbols like ':' to represent types are hard to understand. I had to look into the source code to find out that ':' means int64
From message.go
```golang
func (m *RedisMessage) IsInt64() bool {
	return m.typ == ':'
}
```

Could there be a way to swap in a string that describes the type in these messages instead of seeing a symbol?
The above error message could instead be 
```
panic: redis message type int64 as map key is not supported
```

Maybe there be a way to keep all these symbols as constants and have a map that maps the symbol to the string representation of the type? That way the panic can just reference the symbol map and sub in the type
```go
var (
	i64 = byte(':')
	f64 = byte(',')

	types = map[byte]string{
		i64: "int64",
		f64: "float64",
		// more types below
	}
)
```
```golang
//Then in code use
func (m *RedisMessage) IsInt64() bool {
	return m.typ == types[i64]
}
```
```golang
//Then in error message
panic(fmt.Sprintf("redis message type %s as map key is not supported", types[typ]))
```

If this is deemed actionable I would be interested in implementing the solution to this

rueian:
Hi @kevinxmorales,

Thank you for your suggestions. It sounds like a good idea.

> If this is deemed actionable I would be interested in implementing the solution to this

Many thanks. Your PR will be very welcome. I would suggest putting the proposed `types` in `resp.go` file and somewhere closed these:
[...]

PR #272, opened 2023-06-04
Use type name in RedisMessage error instead of RESP3 character

kevinxmorales:
This change is made to address issue [127](https://github.com/redis/rueidis/issues/271). Currently when a no-op type conversion happens as a result of calling To* or As* the error message produced looks like this
```
panic: redis message type : as map key is not supported
```
The RESP3 character that represents a 64 bit integer ':' is shown to be reason for the panic
With the current change applied, the error message would now look like this 
```
panic: redis message type int64 as map key is not supported
```
Current unit tests are covering the changes to the error messages in message.go, let me know if more tests are desired to add to coverage of the different types.

Notes for maintainers:
For this PR I decided to keep the change small while still addressing the root problem. I see that there is quite a bit of usage of RESP3 characters throughout the library, I could sub those in for these constants if that's asked of me (In this PR or a few others).
Also I left unit tests message_test.go and resp_test.go as is in terms of RESP3 characters instead of swapping in the new constants due to creating a massive diff that would be hard to review, but I could I also get that done if that's wanted.

### DIFF
--- a/message.go
+++ b/message.go
@@ -15,7 +15,7 @@ import (
 const messageStructSize = int(unsafe.Sizeof(RedisMessage{}))
 
 // Nil represents a Redis Nil message
-var Nil = &RedisError{typ: '_'}
+var Nil = &RedisError{typ: typeNull}
 
 // IsRedisNil is a handy method to check if error is a redis nil response.
 // All redis nil response returns as an error.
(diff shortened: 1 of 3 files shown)


rueian:
Thanks @kevinxmorales! LGTM

PR #273, opened 2023-06-06
Feature: Add AsGeosearch method for RedisResult

kevinxmorales:
This change is to add an AsGeosearch method to RedisResult and RedisMessage to convert Geosearch query results into the GeoLocation type that is available in go-redis. Geosearch responses can be tricky to parse as you can get two types back. A standard request like below
```golang
resp = client.Do(context.Background(), client.B().
		Geosearch().
		Key("cities").
		Fromlonlat(76.1474, 43.0481).
		Bybox(200).
		Height(200).
		Mi().
		Build())
```
Will result in receiving an Array of BlobString where the contents are the geolocation names assigned when performing a GeoAdd insertion query
```
//Using github.com/davecgh/go-spew/spew to inspect the type
([]interface {}) (len=2 cap=2) {
 (string) (len=8) "Syracuse",
 (string) (len=6) "Albany"
}
```
However, if additional features are requested, then the data shape changes
```golang
resp := client.Do(context.Background(), client.B().
		Geosearch().
		Key("cities").
		Fromlonlat(76.1474, 43.0481).
		Bybox(900).
		Height(900).
		Mi().Withcoord().Withdist().Withhash().
		Build())
```

```
([]interface {}) (len=2 cap=2) {
 ([]interface {}) (len=4 cap=4) {
  (string) (len=8) "Syracuse",
  (string) (len=6) "0.0001",
  (int64) 3837869069283989,
  ([]interface {}) (len=2 cap=2) {
   (float64) 76.14740163087845,
   (float64) 43.04810095907946
  }
 },
 ([]interface {}) (len=4 cap=4) {
  (string) (len=6) "Albany",
  (string) (len=8) "147.4430",
  (int64) 3649518021478809,
  ([]interface {}) (len=2 cap=2) {
   (float64) 75.14739900827408,
   (float64) 41.04809950607234
  }
 }
}
```
The data shape is different and the size of the Geosearch response array is different depending on the options specified too. I have added unit tests for all possible permutations of the Redis response for Geosearches. 
Doing some local testing with my own geoqueries I was able to successfully parse these queries to GeoLocation structs. I can add in some integration tests where actual queries are made to Redis if that is desired.

### DIFF
--- a/message.go
+++ b/message.go
@@ -984,7 +984,7 @@ func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
 			loc.Name = info[i].string
 			i++
 			//distance
-			if i < len(info) && info[i].IsString() {
+			if i < len(info) && info[i].IsFloat64() {
 				loc.Dist, err = util.ToFloat64(info[i].string)
 				if err != nil {
 					return nil, err
--- a/message.go
+++ b/message.go
@@ -342,6 +342,15 @@ func (r RedisResult) AsFtSearch() (total int64, docs []FtSearchDoc, err error) {
 	return
 }
 
+func (r RedisResult) AsGeosearch() (locations []GeoLocation, err error) {
+	if r.err != nil {
+		err = r.err
+	} else {
+		locations, err = r.val.AsGeosearch()
+	}
+	return
+}
+
 // AsMap delegates to RedisMessage.AsMap
 func (r RedisResult) AsMap() (v map[string]RedisMessage, err error) {
 	if r.err != nil {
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -951,6 +960,78 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+
+func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
+	arr, err := m.ToArray()
+	if err != nil {
+		return nil, err
+	}
+	geoLocations := make([]GeoLocation, 0, len(arr))
+	for _, v := range arr {
+		var loc GeoLocation
+		if v.typ == '$' {
```
rueian, line 977:
```suggestion
		if v.IsString() {
```

message.go
```diff
@@ -951,6 +960,78 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+
+			//name
+			loc.Name, err = info[i].ToString()
+			if err != nil {
+				return nil, err
+			}
+			i++
+			//distance
+			if i < len(info) && info[i].string != "" {
+				loc.Dist, err = info[i].AsFloat64()
```
rueian, line 997:
```suggestion
				loc.Dist, err = util.ToFloat64(m.string)
```

message.go
```diff
@@ -951,6 +960,78 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+				}
+				i++
+			}
+			//coordinates
+			if i < len(info) && info[i].values != nil {
+				cord, err := info[i].ToArray()
+				if err != nil {
+					return nil, err
+				}
+				if len(cord) != 2 {
```
rueian, line 1017:
```suggestion
				if len(cord) < 2 {
```

message.go
```diff
@@ -951,6 +960,78 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+					return nil, fmt.Errorf("got %d, expected 2", len(info))
+				}
+				loc.Longitude, err = cord[0].AsFloat64()
+				if err != nil {
+					return nil, err
+				}
+				loc.Latitude, err = cord[1].AsFloat64()
```
rueian, line 1024:
```suggestion
				loc.Latitude, _ = cord[1].AsFloat64()
```

rueian:
Hi @kevinxmorales, many thanks!

It generally looks good to me. But I think we can omit some unnecessary error handlings to shorten the code.

message.go
```diff
@@ -951,6 +960,78 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+			}
+		} else {
+			info, err := v.ToArray()
+			if err != nil {
+				return nil, err
+			}
+			var i int
+
+			//name
+			loc.Name, err = info[i].ToString()
```
rueian, line 990:
```suggestion
			loc.Name = info[i].string
```

message.go
```diff
@@ -951,6 +960,57 @@ func (m *RedisMessage) AsFtSearch() (total int64, docs []FtSearchDoc, err error)
...
+			loc.Name = v.string
+		} else {
+			info := v.values
+			var i int
+
+			//name
+			loc.Name = info[i].string
+			i++
+			//distance
+			if i < len(info) && info[i].IsFloat64() {
```
rueian, line 987: We can’t use IsFloat64 here, since it only works with RESP3 and fails on RESP2. Checking the field not empty string string is enough.

rueian:
Thanks @kevinxmorales!

PR #274, opened 2023-06-07
perf: use runtime.GOMAXPROCS() to determine max concurrency

ash2k:
Program may be running on an N core machine but only allowed to use M cores via GOMAXPROCS=M. Spawning more than M CPU-bound goroutines just creates useless work for Go scheduler. It can also result in CPU throttling when limits are enforced by cgroups. See https://github.com/golang/go/issues/33803.


rueian:
Thanks @ash2k!

PR #276, opened 2023-06-09
fix: Use net.JoinHostPort to join ip to handle IPv6

SinimaWath:
Fix IPv6 joining with port bug, which affect Redis Sentinel.

**Bug:**

If we got IP address of master is `fdca:4b8:c25:2c15:010:1589:8670:756c` and port `123`. Joining via `fmt.Sprintf("%s:%s", ip,port)` lead to incorrect ip + port adress `fdca:4b8:c25:2c15:010:1589:8670:756c:123`.

**Solution:**

Use `net.JoinHostPort`, which try to determine IPv4 or IPv6 
https://pkg.go.dev/net#JoinHostPort

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -5,13 +5,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"github.com/redis/rueidis/internal/cmds"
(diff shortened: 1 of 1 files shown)


SinimaWath:
@rueian Could you make a review?

rueian:
Sure, @SinimaWath. Many thanks!

Issue #278, opened 2023-06-14
Performance issue for bulk MGETs 


smf8:
Hello, after testing Rueidis in our production setup we faced noticable performance issues in Rueidis clients.

## Use case: 
We're using redis (**sentinel setup**) to cache user metadata with userID as the key. we have around **600MB** worth of data in redis.

Each request consists of a batch of userIDs (**size varies between 50-300**) for which we perform an `MGET` command to get the metadata.
request rate is around **1.5K RPS**.

With go-redis, there is a considerable load on Redis but 99th percentile of latency is **under 50ms.**

But as for rueidis, the total number of queries sent to redis decreased by 70% but 99th percentile of latency has increased to **160ms**
 followed by more than 20% CPU usage.

![image](https://github.com/redis/rueidis/assets/14906511/47e3d479-3344-4bd1-afc2-d9a8d38ce767)

As I've investigated with some benchmarks. This is mainly due to the design of `parallelMGetCache` which is splitting keys and again merging the result which is unnecessary for `single` and `sentinel` clients.

The only solution to the performance issue was to use `MGet` directly (not with the helper function) which will cause cache entry duplication.

## Possible Solution
I think moving the cache to `Client` level will be a good idea for `single` and `sentinel` clients. As for the requirement to flush the data in case of connection failure, we can do so by mapping cache entries from key + slot to value and flushing a subset of the cache in case any connection for a specific slot is lost.

But this will require heavy changes in the library. Can you think of any other solution?

rueian:
Hi @smf8,

I guess the latencies come from capped `concurrency` used here:
https://github.com/redis/rueidis/blob/62f2ced49e5a134b23b57e39c799e5cbc07d3b3f/internal/util/parallel.go#L26-L29

The capped `concurrency` is likely going to make your batch of userIDs complete after many turns. Is that the same as your observation?

If that is the case, I propose to make `rueidis.MGetCache` first grouping multiple `GET`s by mapping their key slot to the underlying `pipe` and pipelining them into the `pipe.DoMultiCache`.

smf8:
Hi @rueian 

Thank you, I'll try to write a benchmark to reproduce the scenario but I've been kind of busy. I'll update the issue as soon as it's ready.

smf8:
You can run the following benchmark to confirm MGET helper functions are having very poor performance.

## Benchmark 
[Benchmark Link](https://gist.github.com/smf8/6a3613d7263af904795cb00532626a9d)

[...]

PR #279, opened 2023-06-16
New Method for grouping MCMDs (MGet, MDel, JsonMget) based on slot connection

smf8:
Hi, This is regarding #278. It's not a very mature implementation and I haven't tested it yet but wanted to share it though to get any feedback regarding the idea.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/redis/rueidis/internal/cmds"
+	"strconv"
 )
 
 type singleClient struct {
@@ -30,6 +31,14 @@ func newSingleClient(opt *ClientOption, prev conn, connFn connFn) (*singleClient
 
 func newSingleClientWithConn(conn conn, builder cmds.Builder, retry bool) *singleClient {
 	return &singleClient{cmd: builder, conn: conn, retry: retry}
+}
+
+func (c *singleClient) commandConnectionID(slot uint16) string {
+	if mux, ok := c.conn.(*mux); ok {
+		return strconv.Itoa(int(slot & uint16(len(mux.wire)-1)))
+	}
+
+	return ""
 }
 
 func (c *singleClient) B() cmds.Builder {
--- a/cluster.go
+++ b/cluster.go
@@ -206,6 +206,11 @@ func parseSlots(slots RedisMessage) map[string]group {
 	return groups
 }
 
+func (c *clusterClient) commandConnectionID(slot uint16) string {
+	slotConnection := c._pick(slot)
+	return slotConnection.Addr()
+}
+
 func (c *clusterClient) _pick(slot uint16) (p conn) {
 	c.mu.RLock()
 	if slot == cmds.InitSlot {
--- a/helper.go
+++ b/helper.go
@@ -15,7 +15,18 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 	if len(keys) == 0 {
 		return make(map[string]RedisMessage), nil
 	}
-	return parallelMGetCache(client, ctx, ttl, cmds.MGets(keys), keys)
+	var slotIDFunc func(uint16) string
+
+	switch client.(type) {
+	case *sentinelClient:
+		slotIDFunc = client.(*sentinelClient).commandConnectionID
+	case *clusterClient:
+		slotIDFunc = client.(*clusterClient).commandConnectionID
+	case *singleClient:
+		slotIDFunc = client.(*singleClient).commandConnectionID
+	}
+
+	return parallelMGetCache(client, ctx, ttl, cmds.UniqueMGets(keys, slotIDFunc), keys)
 }
 
 // MGet is a helper that consults the redis directly with multiple keys by grouping keys within same slot into MGETs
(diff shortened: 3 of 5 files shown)


rueian:
Hi @smf8, Many thanks.

However, after revisiting the code, I found that the `mux.DoMultiCache` already grouped commands to its underlying pipes.

So, we probably should let these helpers not group commands but just use the `client.DoMultiCache` and `client.DoMulti` directly. Doing it this way requires less effort.

Then, the one thing left is to make the `clusterclient.DoMultiCache` and `clusterclient.DoMulti` group commands to `mux`s.

smf8:
Hi @rueian,

I've checked `DoMultiCache` implementation and it seems to work if we convert a single `MGET` to separate `Get` commands. correct?

if that is the case, based on our previous benchmarks and [this comparison](https://www.distributedstack.dev/redis-mget-vs-pipeline/), it seems to have some performance drawbacks. 

> Then, the one thing left will be to make the `clusterclient.DoMultiCache` and `clusterclient.DoMulti` group commands to `mux`s.

Isn't it better to isolate the changes to the `helper.go` only? This can guarantee to not cause any unwanted behavior in the `clusterclient.DoMultiCache` and `clusterclient.DoMulti` functions.

rueian:
Hi @smf8,

> I've checked `DoMultiCache` implementation and it seems to work if we convert a single `MGET` to separate `Get` commands. correct?

Yes, that is correct.

> if that is the case, based on our previous benchmarks and [this comparison](https://www.distributedstack.dev/redis-mget-vs-pipeline/), it seems to have some performance drawbacks.

Yes, there are some drawbacks. But I don't think the benchmark can apply to the case of `DoCache+MGET` since it is far more complex than the benchmark. I just did some benchmarking to compare `DoCache+MGET` and `DoCacheMulti+GET`, and the results:

When all cache miss:
```
Benchmark
Benchmark/MGET_DoCache
Benchmark/MGET_DoCache-10         	    2059	    551316 ns/op	  242338 B/op	    1523 allocs/op
Benchmark/GET_DoMultiCache
Benchmark/GET_DoMultiCache-10     	    1938	    680229 ns/op	  276854 B/op	    1823 allocs/op
PASS
```

When all cache hit:
```
Benchmark
Benchmark/MGET_DoCache
Benchmark/MGET_DoCache-10         	    8089	    157797 ns/op	   65807 B/op	       2 allocs/op
Benchmark/GET_DoMultiCache
Benchmark/GET_DoMultiCache-10     	    7653	    156690 ns/op	   82271 B/op	       4 allocs/op
PASS
```

It looks like the differences are not too large.

> Isn't it better to isolate the changes to the helper.go only? This can guarantee to not cause any unwanted behavior in the clusterclient.DoMultiCache and clusterclient.DoMulti functions.

Yes, we can do it just in the `helper.go` either by special grouping if the client is a single client or by going to use `DoMultiCache` instead. I personally consider the former one as a quick fix.

---
Benchmark source:

```go
package sep

import (
	"context"
	"math/rand"
	"testing"
	"time"

	"github.com/redis/rueidis"
)

type NoCache struct{}

func (n *NoCache) Get(key string) rueidis.RedisMessage {
	return rueidis.RedisMessage{}
}

func (n *NoCache) Set(key string, val rueidis.RedisMessage) {
}

func (n *NoCache) Del(key string) {
}

func (n *NoCache) Flush() {
}

var charset = []byte("abcdefghijklmnopqrstuvwxyz")

func randstr(n int) string {
	b := make([]byte, n)
	for i := range b {
		b[i] = charset[rand.Intn(len(charset))]
	}
	return string(b)
}

func Benchmark(b *testing.B) {
	makeclient := func() rueidis.Client {
		client, err := rueidis.NewClient(rueidis.ClientOption{
			InitAddress: []string{"127.0.0.1:6379"},
			NewCacheStoreFn: func(option rueidis.CacheStoreOption) rueidis.CacheStore {
				return rueidis.NewSimpleCacheAdapter(&NoCache{})
			},
			PipelineMultiplex: -1,
		})
		if err != nil {
			panic(err)
		}
		return client
	}

	// prepare keys
	keys := make([]string, 1000)
	for i := range keys {
		keys[i] = randstr(10)
	}

	{
		// fill the data
		client := makeclient()
		defer client.Close()
		cmd := client.B().Mset().KeyValue()
		for _, k := range keys {
			cmd = cmd.KeyValue(k, randstr(50))
		}
		if err := client.Do(context.Background(), cmd.Build()).Error(); err != nil {
			panic(err)
		}
	}

	b.Run("MGET DoCache", func(b *testing.B) {
		client := makeclient()
		defer client.Close()
		cmd := client.B().Mget().Key(keys...).Cache()
		cmd.Pin()
		b.ReportAllocs()
		b.ResetTimer()
			for pb.Next() {
				if arr, err := client.DoCache(context.Background(), cmd, time.Minute).ToArray(); err != nil {
					panic(err)
				} else {
					for _, r := range arr {
						if err := r.Error(); err != nil {
							panic(err)
						}
					}
				}
			}
		})
	})
	b.Run("GET DoMultiCache", func(b *testing.B) {
		client := makeclient()
		defer client.Close()
		cmds := make([]rueidis.CacheableTTL, 1000)
		for i := range cmds {
			cmds[i].Cmd = client.B().Get().Key(keys[i]).Cache()
			cmds[i].Cmd.Pin()
			cmds[i].TTL = time.Minute
		}
		b.ReportAllocs()
		b.ResetTimer()
			for pb.Next() {
				for _, resp := range client.DoMultiCache(context.Background(), cmds...) {
					if _, err := resp.ToString(); err != nil {
						panic(err)
					}
				}
			}
		})
	})
}
```

PR #280, opened 2023-06-17
perf: boost DoMulti/Cache of the cluster by grouping cmds to conns, not slots

rueian:
related to https://github.com/redis/rueidis/issues/278

The old implementation of `clusterclient.DoMulti()` and `clusterclient.DoMultiCache()` performs poorly when the number of commands passed in is large. That is because those commands will be grouped into their key slots, resulting in many small chunks being parallelly fired. If one's CPU has limited concurrency, then the performance of these functions degrades significantly.

The new implementation grouping commands into their underlying connections, resulting in a minimum number of chunks to be parallelly fired. Even with limited concurrency, this implementation will have a better performance.

Before:
```
Benchmark
Benchmark/GET_DoMulti
Benchmark/GET_DoMulti-10         	     144	   8138918 ns/op	  484360 B/op	    5015 allocs/op
Benchmark/GET_DoMultiCache
Benchmark/GET_DoMultiCache-10    	    6067	    187792 ns/op	  440528 B/op	    4926 allocs/op
PASS
```

After:
```
Benchmark
Benchmark/GET_DoMulti
Benchmark/GET_DoMulti-10         	    2194	    491935 ns/op	  363532 B/op	    1100 allocs/op
Benchmark/GET_DoMultiCache
Benchmark/GET_DoMultiCache-10    	   13555	     88418 ns/op	  349316 B/op	     104 allocs/op
PASS
```

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/redis/rueidis/internal/cmds"
-	"github.com/redis/rueidis/internal/util"
 )
 
 // ErrNoSlot indicates that there is no redis node owns the key slot.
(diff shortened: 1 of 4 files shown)


rueian:
Hi @smf8, if you are interested in this, please help review the change.

smf8:
Hey @rueian, the changes looks great to me. There is one minor improvement, the implementation of `mux.DoMultiCache` uses the `ParallelKeys` util function. Having only 1 CPU core will limit the client to execute only 1 command at a time which I believe is limiting. Although it can be modified with `GOMAXPROCS`, but changing this env for this single purpose might not be a good idea. 

Can we increase the lower bound of `concurrency` in `closeThenParallel`? for my usage, increasing it to `runtime.GOMAXPROCS(0) * 4` resulted in around 20% improvement in benchmarks But I think it can increase even more for high latency networks and slow commands.

rueian:
> Hey @rueian, the changes looks great to me. There is one minor improvement, the implementation of `mux.DoMultiCache` uses the `ParallelKeys` util function. Having only 1 CPU core will limit the client to execute only 1 command at a time which I believe is limiting. Although it can be modified with `GOMAXPROCS`, but changing this env for this single purpose might not be a good idea.
> 
> Can we increase the lower bound of `concurrency` in `closeThenParallel`? for my usage, increasing it to `runtime.GOMAXPROCS(0) * 4` resulted in around 20% improvement in benchmarks But I think it can increase even more for high latency networks and slow commands.

Sure, I think `runtime.GOMAXPROCS(0) * 4` is good in most case.

PR #284, opened 2023-06-22
feat: force single client connection

fabienjuif:
👋 Hello

First thank you for this superb lib!

## Issue
On a redis instance with one node, but spawned as a cluster (yes a cluster of one node), if we are using the rueidis default behaviour, we can connect to this redis instance (HELLO 3 is ok), but once we send a command, the redis cluster is flooded with `CLUSTER SLOTS` commands.

## Solution
I was not able (and do not have time right now to dig further) to find the issue, I've seen we are looking at InitAddress and we should not flood `CLUSTER SLOTS` commands in for loop from `cluster.go/_refresh`, but I have not dag further.

## Quick fix (this PR)
The quick fix for us is to force rueidis to connect as a single redis instance.
I can understand you don't want to merge this PR since this is a dirty fix, no problem!
But I wanted to share this behaviour!

If someone has the same issue, and while waiting for a proper fix, you can use this in your go.mod:
`replace github.com/redis/rueidis v1.0.9 => github.com/fabienjuif/rueidis v1.0.9-fjuif`

Have a nice day,

### DIFF
--- a/rueidis.go
+++ b/rueidis.go
@@ -81,7 +81,12 @@ type ClientOption struct {
 	// Rueidis will connect to them one by one and issue CLUSTER SLOT command to initialize the cluster client until success.
 	// If len(InitAddress) == 1 and the address is not running in cluster mode, rueidis will fall back to the single client mode.
 	// If ClientOption.Sentinel.MasterSet is set, then InitAddress will be used to connect sentinels
+	// You can bypass this behaviour by using ClientOption.ForceSingleClient.
 	InitAddress []string
+
+	//  ForceSingleClient force the usage of a single client connection, without letting the lib guessing
+	//  if redis instance is a cluster or a single redis instance.
+	ForceSingleClient bool
 
 	// ClientTrackingOptions will be appended to CLIENT TRACKING ON command when the connection is established.
 	// The default is []string{"OPTIN"}
@@ -296,6 +301,12 @@ func NewClient(option ClientOption) (client Client, err error) {
 	}
 	pmbk := option.PipelineMultiplex
 	option.PipelineMultiplex = 0 // PipelineMultiplex is meaningless for cluster client
+
+	if option.ForceSingleClient {
+		option.PipelineMultiplex = singleClientMultiplex(pmbk)
+		conn := makeConn(option.InitAddress[0], &option)
+		return newSingleClientWithConn(conn, cmds.NewBuilder(cmds.NoSlot), !option.DisableRetry), nil
+	}
 	if client, err = newClusterClient(&option, makeConn); err != nil {
 		if len(option.InitAddress) == 1 && (err.Error() == redisErrMsgCommandNotAllow || strings.Contains(strings.ToUpper(err.Error()), "CLUSTER")) {
 			option.PipelineMultiplex = singleClientMultiplex(pmbk)
--- a/rueidis.go
+++ b/rueidis.go
@@ -84,10 +84,6 @@ type ClientOption struct {
 	// You can bypass this behaviour by using ClientOption.ForceSingleClient.
 	InitAddress []string
 
-	//  ForceSingleClient force the usage of a single client connection, without letting the lib guessing
-	//  if redis instance is a cluster or a single redis instance.
-	ForceSingleClient bool
-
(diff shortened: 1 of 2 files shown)


rueian:
Hi @fabienjuif,

Thank you for your PR, but it is weird that your cluster is flooded with `CLUSTER SLOTS` commands.

What is your cluster response to `CLUSTER SLOTS` commands?

fabienjuif:
Yeah this is weird I agree.
I've been trying to reproduce this behaviour with unit test provided with the code, without success.

```
redis@redis-0:/data$ redis-cli CLUSTER SLOTS
1) 1) (integer) 0
   2) (integer) 16383
   3) 1) ""
      2) (integer) 6379
      3) "<40char hash>"
      4) (empty array)
redis@redis-0:/data$
```

rueian:
> Yeah this is weird I agree. I've been trying to reproduce this behaviour with unit test provided with the code, without success.
> 
> ```
> redis@redis-0:/data$ redis-cli CLUSTER SLOTS
> 1) 1) (integer) 0
>    2) (integer) 16383
>    3) 1) ""
>       2) (integer) 6379
>       3) "<40char hash>"
>       4) (empty array)
> redis@redis-0:/data$
> ```

Hey @fabienjuif, thank you so much, From your report,  I think the problem should be this one:
image
that rueidis does not handle the empty address correctly.

rueian:
So, I guess a proper fix will be:
image
Would you like to change your PR to something like this?

fabienjuif:
@rueian As said I'll try the fix your propose, this is great for "auto" discovering.

But I am wondering if it can be a good idea to also let the option introduced in this PR so a user that knows if his cluster is a single redis instance or not can force the client "mode" by himself?

It removes 1 "CLUSTER SLOTS" calls this way, it's cheap I agree.

WDYT?

rueian:
> But I am wondering if it can be a good idea to also let the option introduced in this PR so a user that knows if his cluster is a single redis instance or not can force the client "mode" by himself?

@fabienjuif, I actually don't think it is a good idea since Redis in cluster mode has different behaviors such as it may produce `CROSSSLOT` errors and complex `ASK` redirections. But I agree that the `ForceSingleClient` can be useful in some way that I am not awarded. So we can go proceed with this PR.

rueidis.go
```diff
@@ -296,6 +301,12 @@ func NewClient(option ClientOption) (client Client, err error) {
 	}
 	pmbk := option.PipelineMultiplex
 	option.PipelineMultiplex = 0 // PipelineMultiplex is meaningless for cluster client
+
+	if option.ForceSingleClient {
+		option.PipelineMultiplex = singleClientMultiplex(pmbk)
+		conn := makeConn(option.InitAddress[0], &option)
+		return newSingleClientWithConn(conn, cmds.NewBuilder(cmds.NoSlot), !option.DisableRetry), nil
```
rueian, line 308:
```suggestion
		return newSingleClient(&option, nil, makeConn)
```
The `newSingleClientWithConn` will always return a nil error which is not good.

rueidis.go
```diff
@@ -81,8 +81,13 @@ type ClientOption struct {
 	// Rueidis will connect to them one by one and issue CLUSTER SLOT command to initialize the cluster client until success.
 	// If len(InitAddress) == 1 and the address is not running in cluster mode, rueidis will fall back to the single client mode.
 	// If ClientOption.Sentinel.MasterSet is set, then InitAddress will be used to connect sentinels
+	// You can bypass this behaviour by using ClientOption.ForceSingleClient.
 	InitAddress []string
 
+	//  ForceSingleClient force the usage of a single client connection, without letting the lib guessing
+	//  if redis instance is a cluster or a single redis instance.
+	ForceSingleClient bool
```
rueian, line 89: This bool option should be placed nearby other bool fields to reduce memory alignment space.

fabienjuif:
@rueian I have split your proposal / real fix into a dedicated PR https://github.com/redis/rueidis/pull/285

Regarding this current PR, I have not rebased the branch so we don't loose tracks of commits assuming you'll squash commits on merge, if you want me to rebase once you are happy with it please tell me

rueian:
It is fine. Thank you very much!

PR #285, opened 2023-06-23
fix: cluster of 1 node with unknown ip

fabienjuif:
## What

Related to #284 described issue, this fix proposed by @rueian should fix the parsing of an empty slot IP on clusters of 1 node, whom node does not know his own IP.

## TODOs
- [x] Fix unit test - ok weird test are passing now 🤔 
- [x] Create new unit test
- [x] Test it in real life against such a described cluster

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -205,7 +205,15 @@ func parseSlots(slots RedisMessage, firstAddr string) map[string]group {
 			g.slots = make([][2]int64, 0)
 			g.nodes = make([]string, 0, len(v.values)-2)
 			for i := 2; i < len(v.values); i++ {
-				dst := net.JoinHostPort(v.values[i].values[0].string, strconv.FormatInt(v.values[i].values[1].integer, 10))
+				var dst string
+				switch v.values[i].values[0].string {
+				case "":
+					dst = firstAddr
+				case "?":
+					continue
+				default:
+					dst = net.JoinHostPort(v.values[i].values[0].string, strconv.FormatInt(v.values[i].values[1].integer, 10))
+				}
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -112,7 +112,7 @@ func (c *clusterClient) _refresh() (err error) {
 		return err
 	}
 
-	groups := parseSlots(reply)
+	groups := parseSlots(reply, c.opt.InitAddress[0])
```
rueian, line 115: Always using the first address is probably not correct. According to the document, the address should be the one where the reply was sent from.

rueian: I think we might need to change the `results` from being
```go
results := make(chan RedisResult, len(c.conns))
```
to being
```go
results := make(chan struct{reply RedisResult; conn conn}, len(c.conns))
```
So that we can keep track of reply and conn pairs and use `conn.Addr()` later to get the reply address.

cluster_test.go
```diff
@@ -19,12 +19,12 @@ var slotsResp = newResult(RedisMessage{typ: '*', values: []RedisMessage{
 		{typ: ':', integer: 0},
 		{typ: ':', integer: 16383},
 		{typ: '*', values: []RedisMessage{ // master
-			{typ: '+', string: ""},
+			{typ: '+', string: "127.0.0.1"},
```
fabienjuif, line 22: @rueian I choose to put IP everywhere since it was the default behaviour you had in mind right?
And I created a new slot response without the IP set: `singleSlotWithoutIP`

rueian: Yes, many thanks!

PR #298, opened 2023-07-11
feat: use uber-go/mock instead of golang/mock

### LINKED ISSUE #297: golang/mock archived
- golang/mock is archived and uber-go/mock is maintained fork instead.
- rueidis/mock is still using golang/mock. Can we change to uber-go/mock?

dlsrb6342:
- Resolves #297


rueian:
Thanks @dlsrb6342!

Issue #338, opened 2023-08-16
Redis 7.2.0-v0 returns array not object


davidroman0O:
Hello 😄 

I don't know if it's me 🤷‍♂️ 
I've been testing the on v7.2.0 and `Fetch` or other methods don't works.

So I've been looking at the code, i saw the `ToString` and i decided to test it manually.

On a function in my codebase, i've tested this kind of code

```go
var cover T
record, err := r.Client.Do(context.Background(), r.Client.B().JsonGet().Key(fmt.Sprintf("%v:%v", PrefixName, id)).Build()).ToString()
if err == nil {
   fmt.Println("record", record)
   return json.Unmarshal([]byte(record), &conver)
}
```

Get what! It returned me an array! And it wasn't doing that before 👀 

So I delete and reinstall the previous version which was 6.2.6-v9 for me.

Re-tested the same kind of code... guess what it's an object now 😄 

Something is missing on the configuration of the lib? or connection?

davidroman0O:
For context, I'm using a [`bitnami`](https://github.com/bitnami/charts/tree/main/bitnami/redis) Chart (the latest one updated recently) and just changed for `redis-stack-server` for both 6 and 7.

I used this simple code

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"

	"github.com/redis/rueidis"
	"github.com/redis/rueidis/om"
)

type Example struct {
	Key string `json:"key" redis:",key"` // the redis:",key" is required to indicate which field is the ULID key
	Ver int64  `json:"ver" redis:",ver"` // the redis:",ver" is required to do optimistic locking to prevent lost update
	Str string `json:"myStr"`            // both NewHashRepository and NewJSONRepository use json tag as field name
}

func main() {
	ctx := context.Background()
	c, err := rueidis.NewClient(rueidis.ClientOption{InitAddress: []string{"127.0.0.1:6379"}})
	if err != nil {
		panic(err)
	}
	// create the repo with NewHashRepository or NewJSONRepository
	repo := om.NewJSONRepository("something", Example{}, c)

	exp := repo.NewEntity()
	exp.Key = "MYKEY"
	exp.Str = "mystr"
	fmt.Println(exp.Key) // output 01FNH4FCXV9JTB9WTVFAAKGSYB
	repo.Save(ctx, exp)  // success

	// lookup "my_prefix:01FNH4FCXV9JTB9WTVFAAKGSYB" through client side caching
[...]

PR #339, opened 2023-08-17
fix: always use the . legacy root syntax in om.JSONRepository

rueian:
Fix https://github.com/redis/rueidis/issues/338


davidroman0O:
It's about this documentation, just in case someone need to know 
https://github.com/RedisJSON/RedisJSON/releases/tag/v2.6.4

PR #342, opened 2023-08-24
Set readonly flag for FT.SEARCH command

chkp-omris:
I tried to use a redis cluster with DoMulti of FT.SEARCH commands and found that it's not marked as a readonly command.
I made sure that redis treats it as readonly with `COMMAND INFO FT.SEARCH`.

### DIFF
--- a/internal/cmds/gen_search.go
+++ b/internal/cmds/gen_search.go
@@ -3841,7 +3841,7 @@ func (c FtProfileQuerytypeSearch) Query(query string) FtProfileQuery {
 type FtSearch Completed
 
 func (b Builder) FtSearch() (c FtSearch) {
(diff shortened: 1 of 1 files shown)


rueian:
Hi @chkp-omris,

Thank you for verifying that the `FT.SEARCH` should be marked with readonly flag. The original [commands.json](https://github.com/RediSearch/RediSearch/blob/master/commands.json) somehow doesn't mention this.

The `internal/cmds/gen_search.go` file you modified is actually generated by running `go run gen.go` in the `hack/cmds` directory. Would you like to add `ftsearch` here: https://github.com/redis/rueidis/blob/92d534f0a529a18985a2a3eca320c1395e461a19/hack/cmds/gen.go#L1092 and generate to file again?

PR #346, opened 2023-08-25
feat: add ReplicaOnly support for cluster client

NeoHuang:
added ReplicaOnly support for cluster client. 

tested with a cluster with two shards, 3 nodes each.

@rueian could you let me know if this implementation makes sense? I'll add tests later

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -3,6 +3,7 @@ package rueidis
 import (
 	"context"
 	"errors"
+	"math/rand"
 	"net"
 	"runtime"
 	"strconv"
@@ -176,7 +177,6 @@ func (c *clusterClient) _refresh() (err error) {
 
 	groups := parseSlots(reply, addr)
 
-	// TODO support read from replicas
 	conns := make(map[string]conn, len(groups))
 	for _, g := range groups {
 		for _, addr := range g.nodes {
@@ -204,7 +204,11 @@ func (c *clusterClient) _refresh() (err error) {
 
 	slots := [16384]conn{}
 	for master, g := range groups {
-		cc := conns[master]
+		addr := master
+		if c.opt.ReplicaOnly && len(g.nodes) > 1 {
+			addr = g.nodes[1+rand.Intn(len(g.nodes)-1)]
+		}
+		cc := conns[addr]
 		for _, slot := range g.slots {
 			for i := slot[0]; i <= slot[1]; i++ {
 				slots[i] = cc
--- a/pipe.go
+++ b/pipe.go
@@ -173,6 +173,9 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps bool) 
 	if option.SelectDB != 0 {
 		init = append(init, []string{"SELECT", strconv.Itoa(option.SelectDB)})
 	}
+	if option.ReplicaOnly && option.Sentinel.MasterSet == "" {
+		init = append(init, []string{"READONLY"})
+	}
 	if option.ClientNoTouch {
 		init = append(init, []string{"CLIENT", "NO-TOUCH", "ON"})
 	}
@@ -204,6 +207,10 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps bool) 
 				err = r.Error()
 			}
 			if err != nil {
+				if init[i][0] == "READONLY" {
+					// igore READONLY command error
+					continue
+				}
 				if re, ok := err.(*RedisError); ok {
 					if !r2 && noHello.MatchString(re.string) {
 						r2 = true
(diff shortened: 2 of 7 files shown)


cluster.go
```diff
@@ -205,6 +205,15 @@ func (c *clusterClient) _refresh() (err error) {
 	slots := [16384]conn{}
 	for master, g := range groups {
 		cc := conns[master]
+		if c.opt.ReplicaOnly {
+			if replicaAddr := g.pickReplica(); replicaAddr != "" {
+				cc = conns[replicaAddr]
+				result := cc.Do(context.Background(), cmds.ReadOnlyCmd)
```
rueian, line 211: Hi @NeoHuang,

Thank you for implementing the `ReplicaOnly` for the cluster client.

Your implementation generally looks good to me. However, setting the `READONLY` state right here is not a good idea. It will not only make this refresh function slow but also make it hard to finish its job.

I think it will be better to pipeline the `READONLY` right after the `SELECT` command.

Here for RESP3:
https://github.com/redis/rueidis/blob/6556f2fa4d07c416b3029f04419ef77067f09b8a/pipe.go#L173-L175

And here for RESP2:
https://github.com/redis/rueidis/blob/6556f2fa4d07c416b3029f04419ef77067f09b8a/pipe.go#L247-L249

We can probably rely on the `option.ReplicaOnly` flag to pipeline the `READONLY` command after these two blocks if the pipe is managed under a cluster client.

NeoHuang: Hi @rueian , thanks for the feedback and I think it's a good idea. I'll make that change

NeoHuang: actually if we call it here in newPipe, it would panic if the redis doesn't have the cluster mode enabled. but we don't seem to have any context that whether cluster mode is supported or not here as we always try cluster mode first? 
any idea how we can get around?

rueian: Was the panic you mentioned caused by the `return nil, err` here?
https://github.com/redis/rueidis/blob/19e115b2d48d3e434542cd22dcd9b15f873a3285/pipe.go#L217-L218

We probably need to intercept that case and return `p, err` instead.

NeoHuang: it's panic here https://github.com/redis/rueidis/blob/19e115b2d48d3e434542cd22dcd9b15f873a3285/rueidis.go#L306

the client is nil and not checked in this line. and yeah, the nil client was returned by your above code, but I am not sure if it's a good idea to intercept that specific case. I think I would just return error after this line

https://github.com/redis/rueidis/blob/19e115b2d48d3e434542cd22dcd9b15f873a3285/rueidis.go#L303
like this
```go
	if client, err = newClusterClient(&option, makeConn); err != nil {
		if client == nil || client == (*clusterClient)(nil) {
			return nil, err
		}
```

reason is that setting ReplicaOnly on single client is not expected anyway, I would rather throw this error instead of hiding it. also it would add more graceful handling of nil client cases for other scenarios. WDYT? :)

rueian: Because if the target server belongs to a redis cluster, the `READONLY` command will succeed, and if the target server doesn't belong to a cluster, the `READONLY` command will have no effect.

rueian: So, we can safely pipeline the `READONLY` command if `option.ReplicaOnly` is set without checking its response.

NeoHuang:
@rueian thanks. updated the code according to the discussions and added tests

pipe.go
```diff
@@ -204,6 +209,10 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps bool)
 				err = r.Error()
 			}
 			if err != nil {
+				if readOnlyCmdIndex >= 0 && i == readOnlyCmdIndex {
```
rueian, line 212: Maybe this way is better and we don't have to introduce the `readOnlyCmdIndex`:
```suggestion
				if init[i][0] == "READONLY" {
```

cluster.go
```diff
@@ -1069,6 +1075,15 @@ func (r *connretrycache) ResetLen(n int) {
 	}
 }
 
+func (g group) pickReplica() string {
+	// no replica exists
+	if len(g.nodes) <= 1 {
+		return ""
```
rueian, line 1081: Hi @NeoHuang,

Almost there. I find that this line is not covered by tests:
image

However, I am also wondering if the double-checking logic `len(g.nodes) <= 1` and `replicaAddr != ""` could be merged?

For example just:
```go
addr := master
if c.opt.ReplicaOnly && len(g.nodes) > 1 {
    addr = g.nodes[1+rand.Intn(len(g.nodes)-1)]
}
```

rueian:
> btw, [CLUSTER SLOTS](https://redis.io/commands/cluster-slots/) seems to be deprecated. do we have plan to implement [CLUSTER SHARDS](https://redis.io/commands/cluster-shards/)

Yes, we do need to implement CLUSTER SHARDS. We already have an issue opened for this https://github.com/redis/rueidis/issues/350. Given that CLUSTER SLOTS can still work for a long time, we have plenty of time on this.

rueian:
Merged. Many thanks! @NeoHuang

PR #353, opened 2023-08-28
Feat: Human readable results

### LINKED ISSUE #292: Dump RedisResult response structures to console/log/JSON/etc.
`RedisResult` structures can be quite complex and counterintuitive to navigate initially (being nested lists of lists of different sizes).

To support rapid prototyping, there should be a convenient way to dump response structures to a human-readable format for inspection during early development and debugging phases.

Exca-DK:
closes #292

### DIFF
--- a/message.go
+++ b/message.go
@@ -449,6 +449,30 @@ func (r RedisResult) CachePXAT() int64 {
 	return r.val.CachePXAT()
 }
 
+// String returns human-readable representation of RedisResult
+func (r RedisResult) String() string {
+	v, err := r.MarshalJSON()
+	if err != nil {
+		return ""
+	}
+	return string(v)
+}
+
+// MarshalJSON implements json.Marshaler interface
+func (r *RedisResult) MarshalJSON() ([]byte, error) {
+	type PrettyRedisResult struct {
+		Error   string        `json:"Error,omitempty"`
+		Message *RedisMessage `json:"Message,omitempty"`
+	}
+	obj := PrettyRedisResult{}
+	if r.err != nil {
+		obj.Error = r.err.Error()
+	} else {
+		obj.Message = &r.val
+	}
+	return json.Marshal(obj)
+}
+
 // RedisMessage is a redis response message, it may be a nil response
 type RedisMessage struct {
 	attrs   *RedisMessage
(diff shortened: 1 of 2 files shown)


rueian:
Hi @Exca-DK,

Thank you for implementing this. However, I am hesitant about adding the `json.Marshaler` interface.

People will most likely expect the `json.Marshaler` should be performant and its output can be restored back to the source struct by using `json.Unmarshal`.

I think we had better just keep the Stringer interface which is widely accepted as a debugging helper.

message.go
```diff
@@ -1236,3 +1260,44 @@ func (m *RedisMessage) approximateSize() (s int) {
...
+		obj.Ttl = time.UnixMilli(m.CachePXAT()).UTC().String()
+	}
+	if err := m.Error(); err != nil {
+		obj.Error = err.Error()
+	}
+	switch m.typ {
+	case typeFloat, typeBlobString, typeSimpleString, typeVerbatimString, typeBigNumber:
+		obj.Value = m.string
+	case typeBool:
+		obj.Value = strconv.FormatBool(m.integer == 1)
```
rueian, line 1296: Couldn't it just be like this?
```suggestion
		obj.Value = m.integer == 1
```

Exca-DK:
You are right, I made it into an internal type, it shouldn't be known to people at all. What do you think?

message.go
```diff
@@ -1236,3 +1262,51 @@ func (m *RedisMessage) approximateSize() (s int) {
...
+}
+
+type prettyRedisMessage RedisMessage
+
+// MarshalJSON implements json.Marshaler interface
+func (m *prettyRedisMessage) MarshalJSON() ([]byte, error) {
+	type PrettyRedisMessage struct {
+		Type  string `json:"Type,omitempty"`
+		Error string `json:"Error,omitempty"`
+		Ttl   string `json:"Ttl,omitempty"`
```
rueian, line 1282:
```suggestion
		TTL   string `json:"TTL,omitempty"`
```

message.go
```diff
@@ -449,6 +449,32 @@ func (r RedisResult) CachePXAT() int64 {
 	return r.val.CachePXAT()
 }
 
+// String returns human-readable representation of RedisResult
+func (r *RedisResult) String() string {
+	v, err := (*prettyRedisResult)(r).MarshalJSON()
+	if err != nil {
```
rueian, line 455: We can just ignore the err here.

```suggestion
```

message.go
```diff
@@ -1236,3 +1262,51 @@ func (m *RedisMessage) approximateSize() (s int) {
 	}
 	return
 }
+
+// String returns human-readable representation of RedisMessage
+func (m *RedisMessage) String() string {
+	v, err := (*prettyRedisMessage)(m).MarshalJSON()
+	if err != nil {
```
rueian, line 1269: We can just ignore the err here.

```suggestion
```

rueian:
Hi @Exca-DK,

Could you also add tests for these uncovered lines?
image

rueian:
Thanks @Exca-DK!

PR #361, opened 2023-09-02
feat: use CLUSTER SHARDS 

### LINKED ISSUE #350: Use CLUSTER SHARDS instead of CLUSTER SLOTS.
With the release of Redis 7, the command "CLUSTER SLOTS" will be deprecated, and you should use "CLUSTER SHARDS" instead.

Exca-DK:
closes #350

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -316,32 +316,20 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
 // defaultAddr is needed in case the node does not know its own IP
 func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]group {
 	parseNodeEndpoint := func(msg map[string]RedisMessage) string {
-		var ip string
-		switch msg["ip"].string {
+		endpoint := msg["endpoint"].string
+		switch endpoint {
 		case "":
 			return defaultAddr
 		case "?":
 			return ""
-		default:
-			ip = msg["ip"].string
 		}
 
-		_, isPort := msg["port"]
-		_, isTlsPort := msg["tls-port"]
-		var port int64
-		switch {
-		case isPort && !isTlsPort:
-			port = msg["port"].integer
-		case !isPort && isTlsPort:
+		port := msg["port"].integer
+		if tls && msg["tls-port"].integer > 0 {
 			port = msg["tls-port"].integer
-		default:
-			if tls {
-				port = msg["tls-port"].integer
-			} else {
-				port = msg["port"].integer
-			}
 		}
-		return net.JoinHostPort(ip, strconv.FormatInt(port, 10))
+
+		return net.JoinHostPort(endpoint, strconv.FormatInt(port, 10))
 	}
 
 	groups := make(map[string]group, len(shards.values))
--- a/cluster.go
+++ b/cluster.go
@@ -197,7 +197,7 @@ func (c *clusterClient) _refresh() (err error) {
 	if version < 7 {
 		groups = parseSlots(reply, addr)
 	} else {
-		groups = parseShards(reply, addr)
+		groups = parseShards(reply, addr, c.opt.TLSConfig != nil)
 	}
 
 	conns := make(map[string]conn, len(groups))
@@ -314,7 +314,7 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
 
 // parseShards - map redis shards for each redis nodes/addresses
 // defaultAddr is needed in case the node does not know its own IP
-func parseShards(shards RedisMessage, defaultAddr string) map[string]group {
+func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]group {
 	parseNodeEndpoint := func(msg map[string]RedisMessage) string {
 		var ip string
 		switch msg["ip"].string {
@@ -326,18 +326,20 @@ func parseShards(shards RedisMessage, defaultAddr string) map[string]group {
 			ip = msg["ip"].string
 		}
 
-		_, isPort := msg["port"]
-		_, isTlsPort := msg["tls-port"]
 		var port int64
-		switch {
-		case isPort && !isTlsPort:
-			port = msg["port"].integer
-		case !isPort && isTlsPort:
+		switch tls {
+		case true:
+			if _, ok := msg["tls-port"]; !ok {
+				return ""
+			}
 			port = msg["tls-port"].integer
-		default:
-			// TODO which one to choose if both set?
+		case false:
+			if _, ok := msg["port"]; !ok {
+				return ""
+			}
 			port = msg["port"].integer
 		}
+
 		return net.JoinHostPort(ip, strconv.FormatInt(port, 10))
 	}
 
@@ -345,12 +347,11 @@ func parseShards(shards RedisMessage, defaultAddr string) map[string]group {
 	for _, v := range shards.values {
 		var master string
 		nodes := v.values[3].values
-		// Not specified that 1st entry is always master
 		for i := 0; i < len(nodes); i++ {
-			if nodes[i].values[len(nodes[i].values)-3*2+1].string != "master" {
+			dict, _ := nodes[i].ToMap()
+			if dict["role"].string != "master" {
 				continue
 			}
-			dict, _ := nodes[i].ToMap()
 			master = parseNodeEndpoint(dict)
 			break
 		}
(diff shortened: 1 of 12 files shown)


rueidiscompat/adapter_test.go
```diff
@@ -129,6 +129,30 @@ func testCluster(resp3 bool) {
 	})
 
 	Describe("Cluster", func() {
+		It("ClusterShards", func() {
+			// TODO Is there other way to ensure that it runs only on server which supports it?? resp3 can be v6.
```
rueian, line 133: We don't need to worry about that here. Because we use Redis 7 in testing.

cluster.go
```diff
@@ -289,6 +312,74 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
...
+			port = msg["port"].integer
+		}
+		return net.JoinHostPort(ip, strconv.FormatInt(port, 10))
+	}
+
+	groups := make(map[string]group, len(shards.values))
+	for _, v := range shards.values {
+		var master string
+		nodes := v.values[3].values
+		// Not specified that 1st entry is always master
```
Exca-DK, line 348: Only ClusterSlots docs mention that master is always first. Although all example messages in ClusterShards keep master as 1st entry I wasn't sure if it should be assumed.

cluster.go
```diff
@@ -289,6 +312,74 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
 	return groups
 }
 
+// parseShards - map redis shards for each redis nodes/addresses
+// defaultAddr is needed in case the node does not know its own IP
+func parseShards(shards RedisMessage, defaultAddr string) map[string]group {
+	parseNodeEndpoint := func(msg map[string]RedisMessage) string {
+		var ip string
```
Exca-DK, line 319: I kept it the same as in slots although I think that endpoint should be used instead.
> endpoint: The preferred endpoint to reach the node, see below for more information about the possible values of this field.

cluster.go
rueian: Hi @Exca-DK,

We probably need to iterate through the node list and find the first master role.

cluster.go
```diff
@@ -289,6 +312,74 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
...
+		_, isPort := msg["port"]
+		_, isTlsPort := msg["tls-port"]
+		var port int64
+		switch {
+		case isPort && !isTlsPort:
+			port = msg["port"].integer
+		case !isPort && isTlsPort:
+			port = msg["tls-port"].integer
+		default:
+			// TODO which one to choose if both set?
```
rueian, line 338: It should depend on whether the client enables TLS config or not.

Exca-DK: Okay I get it, One thing I am unsure of is whether normal ports should even be accepted if TLS has been set.

rueian: Yes, I believe it should be accepted if there is no "tls-port" provided.

rueidiscompat/command.go
```diff
@@ -2109,6 +2112,134 @@ func (cmd *ClusterSlotsCmd) Result() ([]ClusterSlot, error) {
...
+		var (
+			start, end int64
+			nodesArr   []ClusterNode
+		)
+		{
+			slots, ok := dict["slots"]
+			if !ok {
+				return &ClusterShardsCmd{err: errors.New("slots not found")}
+			}
+			arr, err := slots.ToArray()
```
rueian, line 2135: We can use `AsIntSlice` instead and ignore the `err`.
```suggestion
			arr, _ := slots.AsIntSlice()
```

rueidiscompat/command.go
```diff
@@ -2109,6 +2112,134 @@ func (cmd *ClusterSlotsCmd) Result() ([]ClusterSlot, error) {
...
+			return &ClusterShardsCmd{err: err}
+		}
+		var (
+			start, end int64
+			nodesArr   []ClusterNode
+		)
+		{
+			slots, ok := dict["slots"]
+			if !ok {
+				return &ClusterShardsCmd{err: errors.New("slots not found")}
```
rueian, line 2133: I think we can just ignore this error.

rueidiscompat/command.go
```diff
@@ -2109,6 +2112,134 @@ func (cmd *ClusterSlotsCmd) Result() ([]ClusterSlot, error) {
...
+			}
+			arr, err := slots.ToArray()
+			if err != nil {
+				return &ClusterShardsCmd{err: err}
+			}
+			if len(arr) != 2 {
```
rueian, line 2139: We can relax the condition to be
```suggestion
			if len(arr) >= 2 {
						start = arr[0]
						end = arr[1]
```

rueidiscompat/command.go
```diff
@@ -2109,6 +2112,134 @@ func (cmd *ClusterSlotsCmd) Result() ([]ClusterSlot, error) {
...
+		val = append(val, ClusterShard{
+			Start: start,
+			End:   end,
+			Nodes: nodesArr,
+		})
+	}
+	return &ClusterShardsCmd{val: val, err: err}
+}
+
+type ClusterShard struct {
```
rueian, line 2212: This struct should be aligned with go-redis

```suggestion
type SlotRange struct {
	Start int64
	End   int64
}

type Node struct {
	ID                string
	Endpoint          string
	IP                string
	Hostname          string
	Port              int64
	TLSPort           int64
	Role              string
	ReplicationOffset int64
	Health            string
}

type ClusterShard struct {
	Slots []SlotRange
	Nodes []Node
}
```

rueian:
image
These two cases should also be covered in tests.

rueidiscompat/command.go
Exca-DK: The `slots` in `RedisMessage`  are of type `blob string` therefore `AsIntSlice()` doesn't work.

rueidiscompat/command.go
Exca-DK: goredis seems to keep track of not just first two slots. I aligned it with that now.

rueidiscompat/command.go
rueian: That's weird. Examples in the document are integers.

cluster.go
rueian: After revisiting the document, I noticed that this logic should be applied to the `endpoint` field, not the `ip` field.

Therefore, we need to change the parse function to be like this:

```go
func parseNodeEndpoint(msg map[string]RedisMessage, defaultAddr string, tls bool) string {
	endpoint := msg["endpoint"].string
	switch endpoint {
	case "":
		return defaultAddr
	case "?":
		return ""
	}

	port := msg["port"].integer
	if tls && msg["tls-port"].integer > 0 {
		port = msg["tls-port"].integer
	}

	return net.JoinHostPort(endpoint, strconv.FormatInt(port, 10))
}
```

cluster.go
```diff
@@ -289,6 +312,76 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
...
+				port = msg["port"].integer
+			}
+		}
+		return net.JoinHostPort(ip, strconv.FormatInt(port, 10))
+	}
+
+	groups := make(map[string]group, len(shards.values))
+	for _, v := range shards.values {
+		var master string
+		nodes := v.values[3].values
```
rueian, line 350: `nodes` may not always be on index 3 anymore.

rueidiscompat/command.go
Exca-DK: It's based on the provided docker stack.
```
127.0.0.1:7010> CLUSTER SHARDS
1) 1) "slots"
   2) 1) "10923"
      2) "16383"
   3) "nodes"
   ....
```

cluster.go
```diff
@@ -289,6 +312,69 @@ func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
...
+				dst := parseNodeEndpoint(dict)
+				if dst == "" {
+					continue
+				}
+				g.nodes = append(g.nodes, dst)
+			}
+		}
+		slots := slotsAndNodes["slots"]
+		arr, _ := slots.ToArray()
+		for i := 0; i < len(arr); i += 2 {
```
rueian, line 368: Use `i + 1 < len(arr)` to make sure that the loop won't panic.
```suggestion
		for i := 0; i + 1 < len(arr); i += 2 {
```

rueidiscompat/command.go
```diff
@@ -2109,6 +2110,116 @@ func (cmd *ClusterSlotsCmd) Result() ([]ClusterSlot, error) {
...
+	for _, v := range arr {
+		dict, err := v.ToMap()
+		if err != nil {
+			return &ClusterShardsCmd{err: err}
+		}
+		shard := ClusterShard{}
+		{
+			slots := dict["slots"]
+			arr, _ := slots.ToArray()
+			for i := 0; i < len(arr); i += 2 {
```
rueian, line 2128:
```suggestion
			for i := 0; i + 1 < len(arr); i += 2 {
```

rueian:
Hi @Exca-DK,

It looks good now. Thank you very much for your contribution!

PR #373, opened 2023-09-21
Add servername and min tls version to tls config

kevinxmorales:
This change sets the TLS server name to the url host and it also sets the minimum TLS version to use. This keeps the compatibility with go-redis url parsing and how it handles TLS configs
I added a unit test to assert that the server name is set correctly in the TLS config. The existing unit test on line 12 should cover the fact that 'redis://' urls are not affected by this change


rueian:
LGTM. Thanks @kevinxmorales

PR #378, opened 2023-09-26
feat: avoid using QUIT command

### LINKED ISSUE #377: Avoid using QUIT
The `QUIT`, as [redis documentation](https://redis.io/commands/quit/) mentioned, is not only deprecated but also hurts server performance actually. Clients should not use it anymore.

However, rueidis currently queues a `QUIT` command here in `Close()` to wait for other pending commands to be completed:
https://github.com/redis/rueidis/blob/2f2f4da3b9a41c23f922ce5551a90088243393da/pipe.go#L320

I think for this purpose, we may use a `PING` to replace `QUIT`.

unknowntpo:
This PR resolves #377.

From the suggestion in the documentation in Redis [QUIT (deprecated)](https://redis.io/commands/quit/)

> Note: Clients should not use this command. Instead,
> clients should simply close the connection when they're not used anymore.
> Terminating a connection on the client side is preferable,
> as it eliminates TIME_WAIT lingering sockets on the server side.

We should close the connection by ourselves,
so I changed to last command from QUIT to PING to wait for other pending commands to be completed.

### DIFF
--- a/internal/cmds/cmds.go
+++ b/internal/cmds/cmds.go
@@ -33,12 +33,7 @@ var (
 	RoleCmd = Completed{
 		cs: newCommandSlice([]string{"ROLE"}),
 	}
-	// QuitCmd is predefined QUIT
-	// We shouldn't use this because its deprecated
-	// https://github.com/redis/rueidis/issues/377
-	QuitCmd = Completed{
-		cs: newCommandSlice([]string{"QUIT"}),
-	}
+
 	// UnsubscribeCmd is predefined UNSUBSCRIBE
 	UnsubscribeCmd = Completed{
 		cs: newCommandSlice([]string{"UNSUBSCRIBE"}),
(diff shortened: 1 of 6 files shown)


rueian:
Hi @unknowntpo,

Thank you. Could you also replace all QUIT in the existing tests?

rueian:
> Sure, so we need to replace all QUIT with PING in tests related to Redis Pipeline, is that correct?

Yes, correct.

internal/cmds/cmds.go
```diff
@@ -34,6 +34,8 @@ var (
 		cs: newCommandSlice([]string{"ROLE"}),
 	}
 	// QuitCmd is predefined QUIT
+	// We shouldn't use this because its deprecated
+	// https://github.com/redis/rueidis/issues/377
 	QuitCmd = Completed{
```
rueian, line 39: Please just remove the `QuitCmd`

unknowntpo: Should I also remove this QUIT in builder ?
https://github.com/redis/rueidis/blob/6a804776c5e2b6e9582d3873e8806342b1751f75/internal/cmds/gen_connection.go#L1198-L1207

rueian: No. We should keep the builder one.

rueian:
Thanks @unknowntpo!

PR #381, opened 2023-10-04
Add support for 'REPLICAOF NO ONE'

nirrattner:
The `REPLICAOF` command supports the arguments `NO ONE` ([link to documentation](https://redis.io/commands/replicaof/)) to stop replication. This doesn't seem to be supported by the current command builder, so this PR adds support for this one special case which breaks from the standard types for `host` and `port`.

Let me know if I'm missing an existing way to do this, or if there's a cleaner implementation.

### DIFF
--- a/internal/cmds/gen_server.go
+++ b/internal/cmds/gen_server.go
@@ -1342,6 +1342,11 @@ func (c ReplicaofHost) Port(port int64) ReplicaofPort {
 	return (ReplicaofPort)(c)
 }
 
+func (c Replicaof) NoOne() ReplicaofPort {
+	c.cs.s = append(c.cs.s, "NO ONE")
+	return (ReplicaofPort)(c)
+}
+
 type ReplicaofPort Incomplete
 
 func (c ReplicaofPort) Build() Completed {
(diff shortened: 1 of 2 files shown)


rueian:
Hi @nirrattner,

The command builder and its tests are generated from here:
https://github.com/redis/rueidis/blob/f10a1ee7c7b42f00469a841ebc79afb3e83342d0/hack/cmds/commands.json#L4229-L4243

After editing the file, you can generate them by running:

```sh
cd hack/cmds
go run gen.go
```

Regarding the `NO ONE` implementation, we may first need to file an issue to the `redis-doc` repo to have an updated commands.json. [The official one](https://github.com/redis/redis-doc/blob/529616854125ddc227dc4edef1a7b195f06b2935/commands.json#L11063-L11092) doesn't support the `NO ONE` either.

If you are in a hurry, you can use the `Arbitrary` builder instead:
```go
c.B().Arbitrary("REPLICAOF", "NO", "ONE").Build()
```

rueian:
Unfortunately, they decided to use [the new `oneof` syntax](https://github.com/redis/redis/issues/12632#issuecomment-1748372706) that the current builder generator of rueidis does not support well. We should wait for me to update the generator for that syntax.

nirrattner:
I'm going to close this PR for now because the change will ultimately look very different. I'm happy to open a new PR when the syntax is supported

rueian:
Hi @nirrattner,

I have made the support of the `oneof` syntax here https://github.com/redis/rueidis/commit/1e6830314fc96fb0ea9599a0b68f4a05b9f00c55

Would you like to correct the `REPLICAOF` and the `SLAVEOF` command definitions according to https://github.com/redis/redis/pull/12633 and run `go generate` again?

Issue #384, opened 2023-10-05
Error during a failover


arielzach:
Hello,

I'm using Elasticache with the cluster-require-full-coverage config set to 'no.' When I perform a failover on a 3-node cluster, I notice that the client stops reading. It should continue reading from the other two nodes until the failover completes execution.

rueian:
Hi @arielzach,

Could you provide more details about your setup? Such as your rueidis version and the roles of your 3 redis nodes. Are the two nodes replicas of the failed node?

arielzach:
We are using version 1.0.19, and we have a cluster with 3 masters and 3 replicas (Redis 7). When I execute failover on one of the nodes, the others remain available and should continue to receive traffic.

rueian:
Hi @arielzach. I do find that elasticache’s  configuration endpoint may keep responding old CLUSTER SHARDS information until reconnection.

However, I was hit by the failover rate limit, which only allows 5 failovers per day, so this may takes me more time to fix.

arielzach:
If you need me to test any version, let me know; in my case, I haven't seen the 5 failover warning again.

rueian:
Hi @arielzach, I have pushed the `v1.0.20-pre.1` tag which is based on https://github.com/redis/rueidis/pull/387. It will periodically and proactively refresh shards from the configuration endpoint and should fix your issue. Could you have a try?

arielzach:
Hello @rueian, I just tried it, it worked perfectly. Failover is now almost instantaneous, and it continues writing to the other nodes. The issue I'm noticing is that it's constantly executing the CLUSTER SLOTS/SHARDS command, at a rate of 60 RPM per node, which is quite high considering that, for example, we have applications with 200 nodes, resulting in 12k RPM.

rueian:
> The issue I'm noticing is that it's constantly executing the CLUSTER SLOTS/SHARDS command, at a rate of 60 RPM per node

Yes, you are right. The v1.0.20-pre.1 refreshes shards periodically. I will see if there is a better way to do that.

arielzach:
Hello, I've been running tests, and it's not working correctly. I tried it several times, and sometimes it reconnects without any issues, while other times it simply loses the connection, and no more traffic is visible in the cluster.

arielzach:
[...]

PR #387, opened 2023-10-07
fix: refresh AWS shards periodically

rueian:
Related to #384


rueian:
# Improvements

* When connecting to an AWS Elasicache or MemoryDB, get cluster shard information only from its configuration endpoint.
* Refresh cluster shard information when there are unexpected disconnections.
* Reduce unnecessary goroutines and `CLUSTER SHARDS` requests when refreshing shard information.
* Fix rare inconsistent between shards/slots commands and parsing methods.

PR #402, opened 2023-11-09
feat: read-write splitting

proost:
previous discussion: https://github.com/redis/rueidis/discussions/400

add read-write splitting when cluster mode

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -525,8 +525,7 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last 
 				init = true
 
 				var p conn
-				isSendToReplicas := c.opt.SendToReplicas(cmd)
-				if isSendToReplicas {
+				if c.opt.SendToReplicas(cmd) {
 					p = c.rslots[rand.Intn(int(cmds.InitSlot))]
 				} else {
 					p = c.pslots[rand.Intn(int(cmds.InitSlot))]
@@ -543,8 +542,7 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last 
 			}
 
 			var p conn
-			isSendToReplicas := c.opt.SendToReplicas(cmd)
-			if isSendToReplicas {
+			if c.opt.SendToReplicas(cmd) {
 				p = c.rslots[cmd.Slot()]
 			} else {
 				p = c.pslots[cmd.Slot()]
--- a/cluster.go
+++ b/cluster.go
@@ -249,8 +249,11 @@ func (c *clusterClient) _refresh() (err error) {
 
 	c.mu.RLock()
 	for addr, cc := range c.conns {
-		if _, ok := conns[addr]; ok {
-			conns[addr] = cc
+		if newConn, ok := conns[addr]; ok {
+			conns[addr] = connrole{
+				conn:    cc.conn,
+				replica: newConn.replica,
+			}
 		} else {
 			removes = append(removes, cc.conn)
 		}
--- a/cluster.go
+++ b/cluster.go
@@ -834,6 +834,21 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
 			}
 			count.m[p]++
 		}
+
+		retries := connretrycachep.Get(len(count.m), len(count.m))
+		for cc, n := range count.m {
+			retries.m[cc] = retrycachep.Get(0, n)
+		}
+		conncountp.Put(count)
+
+		for i, cmd := range multi {
+			cc := c.pslots[cmd.Cmd.Slot()]
+			re := retries.m[cc]
+			re.commands = append(re.commands, cmd)
+			re.cIndexes = append(re.cIndexes, i)
+		}
+
+		return retries
 	} else {
 		for _, cmd := range multi {
 			var p conn
@@ -848,22 +863,13 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
 			}
 			count.m[p]++
 		}
-	}
 
-	retries := connretrycachep.Get(len(count.m), len(count.m))
-	for cc, n := range count.m {
-		retries.m[cc] = retrycachep.Get(0, n)
-	}
-	conncountp.Put(count)
-
-	if c.opt.SendToReplicas == nil {
-		for i, cmd := range multi {
-			cc := c.pslots[cmd.Cmd.Slot()]
-			re := retries.m[cc]
-			re.commands = append(re.commands, cmd)
-			re.cIndexes = append(re.cIndexes, i)
+		retries := connretrycachep.Get(len(count.m), len(count.m))
+		for cc, n := range count.m {
+			retries.m[cc] = retrycachep.Get(0, n)
 		}
-	} else {
+		conncountp.Put(count)
+
 		for i, cmd := range multi {
 			var cc conn
 			if c.opt.SendToReplicas(Completed(cmd.Cmd)) {
@@ -876,9 +882,9 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
 			re.commands = append(re.commands, cmd)
 			re.cIndexes = append(re.cIndexes, i)
 		}
-	}
 
-	return retries
+		return retries
+	}
 }
 
 func (c *clusterClient) pickMultiCache(ctx context.Context, multi []CacheableTTL) (*connretrycache, error) {
--- a/cluster.go
+++ b/cluster.go
@@ -18,6 +18,7 @@ import (
 
 // ErrNoSlot indicates that there is no redis node owns the key slot.
 var ErrNoSlot = errors.New("the slot has no redis node")
+var ErrReplicaOnlyConflictsWithSendToReplicas = errors.New("ReplicaOnly conflicts with SendToReplicas option")
 
 type retry struct {
 	cIndexes []int
@@ -70,7 +71,8 @@ var retrycachep = util.NewPool(func(capacity int) *retrycache {
 })
 
 type clusterClient struct {
-	slots  [16384]conn
+	pslots [16384]conn
+	rslots []conn
 	opt    *ClientOption
 	conns  map[string]conn
 	connFn connFn
@@ -91,6 +93,11 @@ func newClusterClient(opt *ClientOption, connFn connFn) (client *clusterClient, 
 		retry:  !opt.DisableRetry,
 		aws:    len(opt.InitAddress) == 1 && strings.Contains(opt.InitAddress[0], "amazonaws.com"),
 	}
+
+	if opt.ReplicaOnly && opt.SendToReplicas != nil {
+		return nil, ErrReplicaOnlyConflictsWithSendToReplicas
+	}
+
 	client.connFn = func(dst string, opt *ClientOption) conn {
 		cc := connFn(dst, opt)
 		cc.SetOnCloseHook(func(err error) {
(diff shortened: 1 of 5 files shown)


cluster.go
```diff
@@ -455,16 +502,58 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last
...
+				} else {
+					p = c.pslots[rand.Intn(int(cmds.InitSlot))]
+				}
+
+				if p == nil {
+					return nil, 0
+				}
+
+				count.m[p]++
+				connIndexes[i] = p
```
rueian, line 537: This part looks inconsistent with the previous implementation where we only do
```go
init = true
continue
```
because we would like to defer slot selection when we have commands with no slot in DoMulti().

rueian: For example, when pipelining the `MULTI`, `SET k v`, `EXIPRE k 1000`, `EXEC` through DoMulti, we would choose the slot by `SET k v` and `EXIPRE k 1000`.

proost: there is test cases for this:

```
	t.Run("DoMulti Single Slot Read Operation And Write Operation + Init Slot", func(t *testing.T) {
		c1 := client.B().Get().Key("K1{a}").Build()
		c2 := client.B().Set().Key("K2{a}").Value("V2{a}").Build()
		c3 := client.B().Info().Build()
		resps := client.DoMulti(context.Background(), c1, c2, c3)
		if v, err := resps[0].ToString(); err != nil || v != "GET K1{a}" {
			t.Fatalf("unexpected response %v %v", v, err)
		}
		if v, err := resps[1].ToString(); err != nil || v != "SET K2{a} V2{a}" {
			t.Fatalf("unexpected response %v %v", v, err)
		}
		if v, err := resps[2].ToString(); err != nil || v != "INFO" {
			t.Fatalf("unexpected response %v %v", v, err)
		}
	})
```

Above case, "INFO" command is ignored. so i add connection to init slot command

cluster.go
```diff
@@ -423,7 +466,11 @@ func (c *clusterClient) Do(ctx context.Context, cmd Completed) (resp RedisResult
 
 func (c *clusterClient) do(ctx context.Context, cmd Completed) (resp RedisResult) {
 retry:
-	cc, err := c.pick(ctx, cmd.Slot())
+	isSendToReplicas := false
+	if c.opt.SendToReplicas != nil {
+		isSendToReplicas = c.opt.SendToReplicas(cmd)
+	}
```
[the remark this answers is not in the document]
proost: Thanks! I change it. [f7197a8](https://github.com/redis/rueidis/pull/402/commits/f7197a8efc893604e35ef4a7dce36c86979f2682)

cluster.go
```diff
@@ -447,6 +490,14 @@ process:
 	return resp
 }
 
+func (c *clusterClient) isSendToReplicas(cmd Completed) bool {
+	isSendToReplicas := false
+	if c.opt.SendToReplicas != nil {
+		isSendToReplicas = c.opt.SendToReplicas(cmd)
+	}
+	return isSendToReplicas
```
rueian, line 498:
```suggestion
	if c.opt.SendToReplicas != nil {
		return c.opt.SendToReplicas(cmd)
	}
	return false
```

cluster.go
```diff
@@ -711,12 +835,29 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
...
+			if p == nil {
+				return nil
+			}
+			count.m[p]++
+		}
+	} else {
+		for _, cmd := range multi {
+			var p conn
+			isSendToReplicas := c.opt.SendToReplicas(Completed(cmd.Cmd))
+			if isSendToReplicas {
```
rueian, line 850:
```suggestion
			if c.opt.SendToReplicas(Completed(cmd.Cmd)) {
```

cluster.go
```diff
@@ -725,11 +866,27 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
...
+			cc := c.pslots[cmd.Cmd.Slot()]
+			re := retries.m[cc]
+			re.commands = append(re.commands, cmd)
+			re.cIndexes = append(re.cIndexes, i)
+		}
+	} else {
+		for i, cmd := range multi {
+			var cc conn
+			isSendToReplicas := c.opt.SendToReplicas(Completed(cmd.Cmd))
+			if isSendToReplicas {
```
rueian, line 880:
```suggestion
]			if c.opt.SendToReplicas(Completed(cmd.Cmd)) {
```

cluster.go
```diff
@@ -455,16 +506,58 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last
...
+
+				count.m[p]++
+				connIndexes[i] = p
+
+				continue
+			}
+
+			var p conn
+			isSendToReplicas := c.opt.SendToReplicas(cmd)
+			if isSendToReplicas {
```
rueian, line 548:
```suggestion
			if c.opt.SendToReplicas(cmd) {
```

cluster.go
rueian: I don't get what you mean by providing the test case.

I mean, previously we intentionally did not select connections for `cmd.Slot() == cmds.InitSlot`. Therefore, I expect you would also write the same code of
```go
if cmd.Slot() == cmds.InitSlot {
	init = true
	continue
}
```
here.

cluster.go
```diff
@@ -455,16 +506,58 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last
 	defer c.mu.RUnlock()
 
 	count := conncountp.Get(len(c.conns), len(c.conns))
-	for _, cmd := range multi {
-		if cmd.Slot() == cmds.InitSlot {
-			init = true
-			continue
+	var connIndexes []conn
```
rueian, line 509: I would like to avoid this `connIndexes` allocation. I think we can assume the `SendToReplicas` is cheap for calling multiple times.

cluster.go
proost: It is possible if multiple connections is selected and there is "InitSlot" command, then "InitSlot" command is ignored. 

For exmaple,
"GET K1{a}" command uses connection "A", "SET K2{a} V2{a}" command uses connection "B", and there is "InitSlot" command likes "INFO" together. original code ignore "INFO" because none of connections is assigned.

So if you want to changed to original code, above ignoring command is what you intended?

cluster.go
proost: [600ecb0](https://github.com/redis/rueidis/pull/402/commits/600ecb060dcf12a25048d89d10783a8598158770) 
I change it

cluster.go
proost: [600ecb0](https://github.com/redis/rueidis/pull/402/commits/600ecb060dcf12a25048d89d10783a8598158770)

also change it.

cluster.go
proost: [600ecb0](https://github.com/redis/rueidis/pull/402/commits/600ecb060dcf12a25048d89d10783a8598158770)

cluster.go
proost: [600ecb0](https://github.com/redis/rueidis/pull/402/commits/600ecb060dcf12a25048d89d10783a8598158770)

cluster.go
rueian: I see the problem. But, we should only pick one connection if "InitSlot" commands are involved, otherwise commands like `MULTI` and `EXEC` will fail.

proost: OK. How about this? 
[eac17d5](https://github.com/redis/rueidis/pull/402/commits/eac17d5f2ab9afedfeab468c8529d1343365d72f)
When "InitSlot" command exist among multiple commands, use one of primary node only.

cluster.go
```diff
@@ -608,7 +714,21 @@ func fillErrs(n int, err error) (results []RedisResult) {
...
+			if cmd.Slot() == cmds.InitSlot {
+				continue
+			}
+
+			isSendToReplicas = isSendToReplicas && c.opt.SendToReplicas(cmd)
+			if !isSendToReplicas {
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
proost: This part also changed it in here. [eac17d5](https://github.com/redis/rueidis/pull/402/commits/eac17d5f2ab9afedfeab468c8529d1343365d72f)

cluster.go
rueian: > OK. How about this? [eac17d5](https://github.com/redis/rueidis/pull/402/commits/eac17d5f2ab9afedfeab468c8529d1343365d72f) When "InitSlot" command exist among multiple commands, use one of primary node only.

Hi @proost,

Sure. Using one of the primary nodes only sounds good to me.

However, the [eac17d5](https://github.com/redis/rueidis/pull/402/commits/eac17d5f2ab9afedfeab468c8529d1343365d72f) makes the _pickMulti becomes too complex, IMHO. How about we refactor it a little bit to first detect if there are "InitSlot" commands? If yes, we then just go to the original logic.

proost: [1178a7c](https://github.com/redis/rueidis/pull/402/commits/1178a7cfa1763dd2333525147020c5bf2ae9f708)

Sure! I changed it more simple

cluster.go
```diff
@@ -358,26 +388,37 @@ func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]g
...
+				break
+			}
+		} else {
+			p = c.pslots[slot]
 		}
 	} else {
-		p = c.slots[slot]
+		switch {
+		case slot == cmds.InitSlot:
+			p = c.pslots[rand.Intn(int(cmds.InitSlot))]
```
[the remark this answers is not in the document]
rueian: Hi @proost,

Your pconns reminds me that the original implementation is wrong. The original implementation is possible to choose a replica connection for “InitSlot” commands.

However, I wonder if there is a simpler solution than introducing pconns.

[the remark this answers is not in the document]
proost: Yes, I agree that introducing "pconns" makes code more complex. but when choose primary node connection randomly among available primary node connections, pconn is simple & efficient solution.
How about replace "conns" with "pconns" and "rconns" like i did to replace "slots" to "pslots" and "rslots"?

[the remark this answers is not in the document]
rueian: Thank you for the `pconns` and `rconns` proposal.

But, after reviewing the access pattern of the `conns` field, I think we still need this unified map. If we replace it with `pconns` and `rconns`, many places will become much more complex, for example:
https://github.com/redis/rueidis/blob/4532630306ad234bc53cd97441e9f208fa6663c5/cluster.go#L387-L390
Here, we must check both fields to find an existing connection.

Another example is here:
https://github.com/redis/rueidis/blob/4532630306ad234bc53cd97441e9f208fa6663c5/cluster.go#L222-L232
where we reuse living connections and find which connections to remove after refreshing the cluster information. If we split the `conns` into two fields, this logic will become far more complex, especially when handling living connections switching their roles from one to the other.

I think the simplest solution to this problem could be extending the `conns` field to be a `map[string]struct{conn, bool}` where the new bool flag indicates it is a primary connection or not.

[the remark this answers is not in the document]
proost: Nice. I changed it [3139c13](https://github.com/redis/rueidis/pull/402/commits/3139c13ca43b983f16a136238fa71deb1bd2e0a6)

cluster.go
```diff
@@ -82,15 +84,26 @@ type clusterClient struct {
 	aws    bool
 }
 
+// NOTE: clusterConnection and conn must be initialized at the same time
+type clusterConnection struct {
+	conn              conn
+	isPrimaryNodeConn bool
```
rueian, line 90:
```suggestion
	replica bool
```
I think it is better to consider a conn not a replica connection by default whereas the current isPrimaryNodeConn treats a conn not a primary connection by default.

cluster.go
```diff
@@ -226,31 +253,54 @@ func (c *clusterClient) _refresh() (err error) {
 		if _, ok := conns[addr]; ok {
 			conns[addr] = cc
```
rueian, line 254: The role of a connection can changed overtime, therefore we should not override the whole field from the old map, we should keep the new role instead.

cluster.go
```diff
@@ -82,15 +84,26 @@ type clusterClient struct {
...
+	isPrimaryNodeConn bool
+}
+
 func newClusterClient(opt *ClientOption, connFn connFn) (client *clusterClient, err error) {
 	client = &clusterClient{
 		cmd:    cmds.NewBuilder(cmds.InitSlot),
 		opt:    opt,
 		connFn: connFn,
-		conns:  make(map[string]conn),
+		conns:  make(map[string]*clusterConnection),
```
rueian, line 98:
```suggestion
		conns:  make(map[string]connrole),
```
clusterConnection is too long for me. And I think it is better to use struct directly.

proost: [3bb4fb5](https://github.com/redis/rueidis/pull/402/commits/3bb4fb58e43879fa87824249b8325a4751323c7d) OK. I changed it.

cluster.go
proost: Also changed it too. [3bb4fb5](https://github.com/redis/rueidis/pull/402/commits/3bb4fb58e43879fa87824249b8325a4751323c7d)

cluster.go
proost: Oh, i missed that point. [d115a78](https://github.com/redis/rueidis/pull/402/commits/d115a789988ccb28a17add44fb5422930d2a4442) Thank you!

rueian:
Hi @proost,

Thank you for your hard work! It looks good to me.

PR #404, opened 2023-11-10
Add Normalized Vector Tests for `TestVectorString32`/`64`

h0rv:
Running into some issues parsing vector embeddings in a project. They vectors are normalized and was wondering if that was the issue after looking at the tests for the function. Looks like it is not but adding these tests for better coverage.


h0rv:
Looks like the tests timed out

rueian:
Thanks @h0rv!

PR #418, opened 2023-12-10
feat: let clusterclient respect the option.PipelineMultiplex

### LINKED ISSUE #417: PipelineMultiplex for cluster
Currently `PipelineMultiplex` has no effect for cluster. For cluster, you create 1 tcp connection per redis instance. Would it be possible to increase that value? 

I have issues on the client side with timeout issues. Each client is making around 15k to 20k requests a second. Cluster side is 14 redis instances. Each call is an exec call, that contains 2 redis ops. 

```
	redis, err = rueidis.NewClient(rueidis.ClientOption{
		InitAddress:  <ips>,
		Dialer:       net.Dialer{KeepAlive: -1}, // To disable pings
		DisableRetry: true,
		DisableCache: true,
		ShuffleInit:  true,
		MaxFlushDelay: 20 * time.Microsecond,
	})
```
Using version v1.0.23

rueian:
Resolves #417

### DIFF
--- a/rueidis.go
+++ b/rueidis.go
@@ -317,11 +317,8 @@ func NewClient(option ClientOption) (client Client, err error) {
 		option.PipelineMultiplex = singleClientMultiplex(option.PipelineMultiplex)
 		return newSentinelClient(&option, makeConn)
 	}
-	pmbk := option.PipelineMultiplex
-	option.PipelineMultiplex = 0 // PipelineMultiplex is meaningless for cluster client
-
 	if option.ForceSingleClient {
-		option.PipelineMultiplex = singleClientMultiplex(pmbk)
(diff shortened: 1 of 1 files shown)


salarali:
Would also need to update the documentation:
https://github.com/redis/rueidis/blob/3851dec72cb1f34c5eb39a85df8ad952b28f9508/rueidis.go#L128C54-L128C54

rueian:
> Would also need to update the documentation: https://github.com/redis/rueidis/blob/3851dec72cb1f34c5eb39a85df8ad952b28f9508/rueidis.go#L128C54-L128C54

Sure! Good catch.

salarali (approved):
Thanks for the quick changes.

PR #422, opened 2023-12-12
Feat: Align `go-redis` `TimeseriesCmdable`

unknowntpo:
This PR implemented adapter for `go-redis` `TimeseriesCmdable`.

### DIFF
--- a/message.go
+++ b/message.go
@@ -1130,28 +1130,6 @@ func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
 	return geoLocations, nil
 }
 
-// AsAnyAnyMap converts RedisMessage to map[any]any.
-// we don't use (*RedisMessage).ToAny() on map type because it will convert ele
-// to map[string]any
-func (m *RedisMessage) AsAnyAnyMap() (map[any]any, error) {
-	if !m.IsMap() {
-		panic(fmt.Sprintf("redis message type %s is not a RESP3 map", typeNames[m.typ]))
-	}
-	_m, err := m.ToMap()
-	if err != nil {
-		return nil, err
-	}
-	anyMap := make(map[any]any, len(_m))
-	for k, e := range _m {
-		_e, err := e.ToAny()
-		if err != nil {
-			return nil, err
-		}
-		anyMap[k] = _e
-	}
-	return anyMap, nil
-}
-
 // ToMap check if message is a redis RESP3 map response, and return it
 func (m *RedisMessage) ToMap() (map[string]RedisMessage, error) {
 	if m.IsMap() {
--- a/message.go
+++ b/message.go
@@ -1130,6 +1130,28 @@ func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
 	return geoLocations, nil
 }
 
+// AsAnyAnyMap converts RedisMessage to map[any]any.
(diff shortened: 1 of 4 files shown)


rueidiscompat/adapter.go
```diff
@@ -3513,6 +3546,792 @@ func (c *Compat) TDigestTrimmedMean(ctx context.Context, key string, lowCutQuant
...
+		}
+		if options.BucketTimestamp != nil {
+			cmds.TsMrangeAggregationBucketduration(_cmd).Buckettimestamp(str(options.BucketTimestamp))
+		}
+		if options.Empty {
+			cmds.TsMrangeAggregationBuckettimestamp(_cmd).Empty()
+		}
+		cmds.TsMrangeTotimestamp(_cmd).Filter(filterExpr...)
+		if options.GroupByLabel != nil {
+			// FIXME: Wrong API definition: REDUCE
```
unknowntpo, line 4196: Note that our builder API is wrong.

rueidiscompat/adapter.go
```diff
@@ -3513,6 +3546,792 @@ func (c *Compat) TDigestTrimmedMean(ctx context.Context, key string, lowCutQuant
...
+		}
+		if options.BucketTimestamp != nil {
+			cmds.TsMrevrangeAggregationBucketduration(_cmd).Buckettimestamp(str(options.BucketTimestamp))
+		}
+		if options.Empty {
+			cmds.TsMrevrangeAggregationBuckettimestamp(_cmd).Empty()
+		}
+		cmds.TsMrevrangeTotimestamp(_cmd).Filter(filterExpr...)
+		if options.GroupByLabel != nil {
+			// FIXME: Wrong API definition: REDUCE
```
unknowntpo, line 4295: Note that our builder API is wrong.

rueidiscompat/command.go
```diff
@@ -3170,3 +3170,305 @@ type TDigestMergeOptions struct {
...
+func newMapStringInterfaceCmd(res rueidis.RedisResult) *MapStringInterfaceCmd {
+	cmd := &MapStringInterfaceCmd{}
+	m, err := res.AsMap()
+	if err != nil {
+		cmd.err = err
+		return cmd
+	}
+	strIntMap := make(map[string]any, len(m))
+	for k, ele := range m {
+		var v any = rueidis.Nil
```
unknowntpo, line 3311: This `rueidis.Nil` is added to make tests pass, should we use `rueidis.Nil` or let it be `nil` if `ele.IsNil()` is true ?

rueian: I think `nil` is more reasonable.

message.go
```diff
@@ -1130,6 +1130,28 @@ func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
 	return geoLocations, nil
 }
 
+// AsAnyAnyMap converts RedisMessage to map[any]any.
+// we don't use (*RedisMessage).ToAny() on map type because it will convert ele
+// to map[string]any
+func (m *RedisMessage) AsAnyAnyMap() (map[any]any, error) {
```
rueian, line 1136: `map[any]any` is too hard to use. I don't think we should have this function.

unknowntpo: Okay, I'll remove it.

PR #426, opened 2023-12-15
fix: do not ignore error of `url.Parse`

j178:
If the error is ignored, there will be a nil pinter dereference panic when accessing `u.Scheme`.


rueian:
Hi @j178, good catch and thank you for the test case!

PR #428, opened 2023-12-17
OTEL tweaks

ash2k:
**First commit**: it's good to avoid mutable public fields because they encourage poor user behavior. It's possible to configure the histogram buckets via options, no need to make the default value mutable.

**Second commit**: it's better to fail early if OTEL library returned an error rather than discard an error, store `nil` in the field and then panic with NPE later. There will be no way to find out why the field was `nil` because the error was discarded. Also, it'd be better if `WithClient()` just returned an error.

**Third commit**: `Second()` does something weird instead of just division. It loses precision that way. OpenTelementry libraries just use division, e.g. https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8e44cea6d4fe181343500b6cb680bd501b3d24aa/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L343-L344

Overall I think it's not the best choice to have both `WithClient()` and `NewClient()`. Why not have a single function and let the user configure if they want metrics, traces, both, and with what options. If we want to keep two functions, names are really not clear. They should be `WithTracing()` and `WithMetrics()` or something like that.

But IMO it's better to have a single function. If someone doesn't want e.g. metrics, they can provide a NOOP `MeterProvider`. Same with tracing - there is a NOOP `TracerProvider`.

### DIFF
--- a/rueidisotel/metrics.go
+++ b/rueidisotel/metrics.go
@@ -38,7 +38,10 @@ func WithHistogramOption(histogramOption HistogramOption) Option {
 // - rueidis_dial_conns: number of active connections
 // - rueidis_dial_latency: dial latency in seconds
 func NewClient(clientOption rueidis.ClientOption, opts ...Option) (rueidis.Client, error) {
-	oclient := newClient(opts...)
+	oclient, err := newClient(opts...)
+	if err != nil {
+		return nil, err
+	}
 
 	if clientOption.DialFn == nil {
 		clientOption.DialFn = defaultDialFn
@@ -84,7 +87,7 @@ func NewClient(clientOption rueidis.ClientOption, opts ...Option) (rueidis.Clien
 	return oclient, nil
 }
 
-func newClient(opts ...Option) *otelclient {
+func newClient(opts ...Option) (*otelclient, error) {
 	cli := &otelclient{}
 	for _, opt := range opts {
 		opt(cli)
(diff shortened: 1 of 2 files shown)


rueidisotel/metrics.go
```diff
@@ -44,10 +38,13 @@ func WithHistogramOption(histogramOption HistogramOption) Option {
 // - rueidis_dial_conns: number of active connections
 // - rueidis_dial_latency: dial latency in seconds
 func NewClient(clientOption rueidis.ClientOption, opts ...Option) (rueidis.Client, error) {
-	oclient := newClient(opts...)
+	oclient, err := newClient(opts...)
+	if err != nil {
+		panic(err)
```
rueian, line 43:
```suggestion
		return nil, err
```

rueian:
Hi @ash2k,

> Third commit: Second() does something weird instead of just division. It loses precision that way. OpenTelementry libraries just use division.

Thank you for the reference. Good to know this fun fact.

> But IMO it's better to have a single function. If someone doesn't want e.g. metrics, they can provide a NOOP `MeterProvider`. Same with tracing - there is a NOOP `TracerProvider`.

Totally agreed. I forgot to mark `WithClient` as deprecated when introducing the new `NewClient`. Would you like to help me mark `WithClient` as deprecated?

ash2k:
Done!

rueian:
Thanks @ash2k!

PR #430, opened 2023-12-19
fix: send readonly command

proost:
I missed to send "READONLY" command for replica connection.

I want to integrate "READONLY" command with pipeline constructor, but it quite hard. has any idea?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -70,12 +70,14 @@ var retrycachep = util.NewPool(func(capacity int) *retrycache {
 	}
 })
 
+type clusterConnFn func(dst string, opt *ClientOption, isReplicaConn bool) conn
+
 type clusterClient struct {
 	pslots [16384]conn
 	rslots []conn
 	opt    *ClientOption
 	conns  map[string]connrole
-	connFn connFn
+	connFn clusterConnFn
 	sc     call
 	mu     sync.RWMutex
 	stop   uint32
@@ -92,23 +94,33 @@ type connrole struct {
 
 func newClusterClient(opt *ClientOption, connFn connFn) (client *clusterClient, err error) {
 	client = &clusterClient{
-		cmd:    cmds.NewBuilder(cmds.InitSlot),
-		opt:    opt,
-		connFn: connFn,
-		conns:  make(map[string]connrole),
-		retry:  !opt.DisableRetry,
-		aws:    len(opt.InitAddress) == 1 && strings.Contains(opt.InitAddress[0], "amazonaws.com"),
+		cmd:   cmds.NewBuilder(cmds.InitSlot),
+		opt:   opt,
+		conns: make(map[string]connrole),
+		retry: !opt.DisableRetry,
+		aws:   len(opt.InitAddress) == 1 && strings.Contains(opt.InitAddress[0], "amazonaws.com"),
 	}
 
 	if opt.ReplicaOnly && opt.SendToReplicas != nil {
 		return nil, ErrReplicaOnlyConflict
 	}
 
-	client.connFn = func(dst string, opt *ClientOption) conn {
+	client.connFn = func(dst string, opt *ClientOption, isReplicaConn bool) conn {
 		cc := connFn(dst, opt)
 		cc.SetOnCloseHook(func(err error) {
 			client.lazyRefresh()
 		})
+		if opt.SendToReplicas != nil && isReplicaConn {
+			timeout := opt.Dialer.Timeout
+			if timeout <= 0 {
+				timeout = DefaultDialTimeout
+			}
+
+			ctx, cancel := context.WithTimeout(context.Background(), timeout)
+			defer cancel()
+
+			cc.Do(ctx, cmds.NewCompleted([]string{"READONLY"})) // ignore error
+		}
 		return cc
 	}
 
@@ -129,7 +141,7 @@ func (c *clusterClient) init() error {
 	}
 	results := make(chan error, len(c.opt.InitAddress))
 	for _, addr := range c.opt.InitAddress {
-		cc := c.connFn(addr, c.opt)
+		cc := c.connFn(addr, c.opt, false)
 		go func(addr string, cc conn) {
 			if err := cc.Dial(); err == nil {
 				c.mu.Lock()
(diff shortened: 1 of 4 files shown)


rueian:
Hi @proost, I think we should reuse this block: https://github.com/redis/rueidis/blob/65f705a8f7d7feab5d39dbd6725081f471aa1e6d/pipe.go#L273-L275

One more thing needs to be considered: A replica can be promoted to primary, meaning we should reset the 'READONLY' state with the `READWRITE` command.

proost:
@rueian 
yes, you are right. replica can be promoted to primary. but it is hard to notice that replica is promoted. 
I thinkg that adding `ReplicaConnectionReadOnly` to `rueidis.ClientOption` and sending "READONLY" command in `_refresh` can be better. because use know when replica connection can be read-only connection. how do you think change like this?

rueian:
Hi @proost,

I don’t think we need a new option. How about we modify the ReplicaOnly to true only for replica connections from the refresh function? And if we find a replica connection is promoted when refreshing, then we open a new one without ReplicaOnly.

proost:
@rueian 

I change it. is okay to use `ReplicaOnly`? because it is breaking change.

cluster.go
```diff
@@ -222,10 +217,42 @@ func (c *clusterClient) _refresh() (err error) {
...
+				timeout := c.opt.Dialer.Timeout
+				if timeout <= 0 {
+					timeout = DefaultDialTimeout
+				}
+
+				ctx, cancel := context.WithTimeout(context.Background(), timeout)
+				defer cancel()
+
+				cc.Do(ctx, cmds.NewCompleted([]string{"READONLY"})) // ignore error
+			}(cc)
```
rueian, line 250: Hi @proost,

Thank you for the changes, but this goroutine looks wired to me and I don't expect there will be breaking change by reusing the `ReplicaOnly`. I am afraid we have some misunderstanding here.

What I previously suggested is that the `pipe` will send `READONLY` command already if its `ReplicaOnly` is true.
So I think we can reuse this behavior by modifying the `c.connFn(addr, c.opt)` in the refresh function, and then the above goroutine is not necessary.

rueian: Given that the signature of the `connFn` is `func(dst string, opt *ClientOption) conn` which accepts a ClientOption pointer, I think we can copy the `c.opt` to `c.optReplica` with the `ReplicaOnly` set to true and then we can use `c.connFn(addr, c.optReplica)` for replica connections.

proost: [98040d7](https://github.com/redis/rueidis/pull/430/commits/98040d775392220820aea58a2738889c1607ad53)

@rueian 
I'm sorry about serious misunderstanding. I figured out.

cluster.go
```diff
@@ -225,7 +220,15 @@ func (c *clusterClient) _refresh() (err error) {
 	for master, g := range groups {
 		conns[master] = connrole{conn: c.connFn(master, c.opt), replica: false}
 		for _, addr := range g.nodes[1:] {
-			conns[addr] = connrole{conn: c.connFn(addr, c.opt), replica: true}
+			var cc conn
+			if c.opt.ReplicaOnly {
+				opt := *c.opt
```
rueian, line 225: I am still a little confused. So now, are we requiring users to set `ReplicaOnly` as well even if they use `SendToReplicas`?

I thought we would only set `ReplicaOnly` transparently for users here to the `opt`.

proost: fix it. [3ec3548](https://github.com/redis/rueidis/pull/430/commits/3ec3548e8207c497029f37acad869db5e0d3e5f5) 

when `SendToReplicas` is configured and `ReplicaOnly` is true,
- connection to master: do not send "READONLY"
- connection to replica: send "READONLY"

when `SendToReplicas` is configured and `ReplicaOnly` is false,
- connection to master: do not send "READONLY"
- connection to replica: do not send "READONLY"

when `SendToReplicas` is not configured and `ReplicaOnly` is true,
- connection to master: send "READONLY"

rueian: Why we don't send "READONLY" to replicas when SendToReplicas is configured and ReplicaOnly is false?

Isn't it the original issue you want to fix?

cluster.go
```diff
@@ -223,7 +218,11 @@ func (c *clusterClient) _refresh() (err error) {
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
 	for master, g := range groups {
-		conns[master] = connrole{conn: c.connFn(master, c.opt), replica: false}
+		opt := *c.opt
+		if c.opt.SendToReplicas != nil {
+			opt.ReplicaOnly = false
+		}
+		conns[master] = connrole{conn: c.connFn(master, &opt), replica: false}
```
rueian, line 225: The current code above will allocate a copy of `c.opt` on the heap every time.
We should do it better by preparing the copied `opt` once in advance.

cluster.go
proost: [cd6a674](https://github.com/redis/rueidis/pull/430/commits/cd6a674ec54ce392de2ff6e0d268fc6cb877e344)

Yes, it doesn't matter HOW. i just want to send "READONLY" command to replica connections automatically. but if you think read-only mode as a default is better than i changed it.

cluster.go
proost: [cd6a674](https://github.com/redis/rueidis/pull/430/commits/cd6a674ec54ce392de2ff6e0d268fc6cb877e344)

yes. you're right.

cluster.go
```diff
@@ -222,10 +217,18 @@ func (c *clusterClient) _refresh() (err error) {
...
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
+	masterOpt := *c.opt
+	if c.opt.SendToReplicas != nil {
+		masterOpt.ReplicaOnly = false
+	}
+	replicaOpt := *c.opt
+	if c.opt.SendToReplicas != nil {
+		replicaOpt.ReplicaOnly = !c.opt.ReplicaOnly
+	}
```
rueian, line 227: It is better now. But it still allocates `masterOpt` and `replicaOpt` in every refresh.
We can make it even better by storing a `c.replicaOpt` in the `newClusterClient`.

rueian: >  i just want to send "READONLY" command to replica connections automatically.

Agreed. So, how about this:

when `SendToReplicas` is configured and `ReplicaOnly` is true,

* return `ErrReplicaOnlyConflict`

when SendToReplicas is configured and ReplicaOnly is false,

* connection to master: use `c.connFn(addr, c.opt)`
* connection to replica: use `c.connFn(addr, c.replicaOpt)`

when SendToReplicas is not configured and ReplicaOnly is true,

* connection to all nodes: use `c.connFn(addr, c.opt)`, we don't need `c.replicaOpt` in this case.

proost: [5702ce3](https://github.com/redis/rueidis/pull/430/commits/5702ce3e8deb82dd5848d1adba36b07842e94e57)

Good! i changed it.

cluster.go
```diff
@@ -241,7 +254,8 @@ func (c *clusterClient) _refresh() (err error) {
 
 	c.mu.RLock()
 	for addr, cc := range c.conns {
-		if fresh, ok := conns[addr]; ok {
+		fresh, ok := conns[addr]
+		if ok && cc.replica == fresh.replica {
```
rueian, line 258:
```suggestion
		if ok && (cc.replica == fresh.replica || c.opt.SendToReplicas == nil) {
```

rueian: We can still reuse the conn if c.opt.SendToReplicas is not configured.

proost: [87e2b39](https://github.com/redis/rueidis/pull/430/commits/87e2b39036c247c1106d95207277848d3c7b1f2c)

rueian: Hi @proost, thank you very much! It looks great!

PR #436, opened 2023-12-28
Check for too big option.PipelineMultiplex value

atercattus:
If we use `ClientOption.PipelineMultiplex` in the wrong way (too big), we will get a panic inside rueidis:

https://go.dev/play/p/QE8VYHE6hJX
```
package main

func foo(v int) int {
	return 1 << v
}

func main() {
	println(foo(0), foo(2), foo(50), foo(70), foo(100), foo(200))
}

>> 1 4 1125899906842624 0 0 0
```

In this case, we will get [zero for multiplex value](https://github.com/redis/rueidis/blob/a85dd76338d31ff6e92e55d698dd536e5ea6f10a/mux.go#L97) and zero-sized `wire`, `mu`, `sc` slices. Which will lead to panic [here](https://github.com/redis/rueidis/blob/a85dd76338d31ff6e92e55d698dd536e5ea6f10a/mux.go#L142) for example.

```
goroutine 59 [running]:
github.com/redis/rueidis.(*mux)._pipe(0x0?, 0x0?)
        /home/X/go/pkg/mod/github.com/redis/rueidis@v1.0.25/mux.go:142 +0x71e
github.com/redis/rueidis.(*mux).Dial(0x0?)
        /home/X/go/pkg/mod/github.com/redis/rueidis@v1.0.25/mux.go:188 +0x1b
github.com/redis/rueidis.(*clusterClient).init.func1({0xc00010f640, 0x19}, {0x17ce2d0, 0xc00001b180})
        /home/X/go/pkg/mod/github.com/redis/rueidis@v1.0.25/cluster.go:134 +0x48
created by github.com/redis/rueidis.(*clusterClient).init
        /home/X/go/pkg/mod/github.com/redis/rueidis@v1.0.25/cluster.go:133 +0x9d
```

### DIFF
--- a/rueidis.go
+++ b/rueidis.go
@@ -30,6 +30,8 @@ const (
 	DefaultReadBuffer = 1 << 19
 	// DefaultWriteBuffer is the default value of bufio.NewWriterSize for each connection, which is 0.5MiB
 	DefaultWriteBuffer = 1 << 19
+	// MaxPipelineMultiplex is the maximum meaningful value for ClientOption.PipelineMultiplex
+	MaxPipelineMultiplex = 8
 )
 
 var (
@@ -46,6 +48,8 @@ var (
 	// ErrReplicaOnlyNotSupported means ReplicaOnly flag is not supported by
 	// current client
 	ErrReplicaOnlyNotSupported = errors.New("ReplicaOnly is not supported for single client")
+	// ErrWrongPipelineMultiplex means wrong value for ClientOption.PipelineMultiplex
+	ErrWrongPipelineMultiplex = errors.New("ClientOption.PipelineMultiplex must not be bigger than MaxPipelineMultiplex")
 )
 
 // ClientOption should be passed to NewClient to construct a Client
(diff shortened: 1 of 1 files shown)


mux.go
```diff
@@ -95,7 +95,8 @@ func newMux(dst string, option *ClientOption, init, dead wire, wireFn wireFn) *m
 	var multiplex int
 	if option.PipelineMultiplex >= 0 {
 		multiplex = 1 << option.PipelineMultiplex
-	} else {
+	}
+	if multiplex == 0 {
```
bald2b, line 99: Is this good way? Because we just replace wrong value with default, but user never knows about problem.
I think better panic with explanation why

bald2b (changes requested):

rueian:
Thanks @atercattus! And I agree with @bald2b. It will be better to return an error to users.

Actually, I don't think setting `PipelineMultiplex` to a value larger than 2 will help. The Redis documentation also suggests that ["Using more than 8 threads is unlikely to help much".](https://redis.io/docs/management/config-file/)

atercattus:
We set too big a value by mistake and got unexpected panic :)

I think, we have two ways:
- change a signature for `newMux()` to `newMux(...) (*mux, error)` and for all functions that call it
- update docs for a given behavior.

Now I'm leaning towards the second option. What do you think?

rueian:
> We set too big a value by mistake and got unexpected panic :)
> 
> I think, we have two ways:
> 
> * change a signature for `newMux()` to `newMux(...) (*mux, error)` and for all functions that call it
> * update docs for a given behavior.
> 
> Now I'm leaning towards the second option. What do you think?

I see, but changing the signature of `newMux` is too much. It requires too many changes.

I think we can just validate the value of `PipelineMultiplex` in the `NewClient`.

rueian:
We can put a hard limit of 8, resulting in 256 connections to a node.

rueian (approved):

atercattus:
Could you make a new tag? :)

rueian:
Sure. I will do it later.

PR #446, opened 2024-01-12
fix wrong block tag setting

### LINKED ISSUE #445: IsReadOnly() of Redis streams readonly command with BLOCK should return true
Current implementation overrides readonly flag with block flag.
readonly flag: https://github.com/redis/rueidis/blob/main/internal/cmds/gen_stream.go#L1039
block flag: https://github.com/redis/rueidis/blob/main/internal/cmds/gen_stream.go#L1050

Block tag should be set like blow?
```go
	c.cf |= int16(blockTag)
```

tamayika:
closes #445 

I don't know well about timeseries but I think the same fix should be applied.

### DIFF
--- a/hack/cmds/gen.go
+++ b/hack/cmds/gen.go
@@ -803,7 +803,7 @@ func printBuilder(w io.Writer, parent, next goStruct) {
 
 	for _, cmd := range next.BuildDef.Command {
 		if cmd == "BLOCK" {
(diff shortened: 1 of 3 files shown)


rueian:
> closes #445
> 
> I don't know well about timeseries but I think the same fix should be applied.

Yes, it should be applied as well. LGTM, thanks!

Issue #444, opened 2024-01-11
Support for pushing metrics to prometheus


DestroyerAlpha:
First of all thanks for constantly developing this library based on inputs from the community. I am glad to see that we have instrumented connection level metrics via rueidisotel. I was wondering if it would be possible to have a way to push these metrics to prometheus as well? Something like https://github.com/redis/go-redis/tree/master/extra/redisprometheus.

Would be happy to work on this, if needed, as well.

rueian:
Hi @DestroyerAlpha,

Does an otel prometheus exporter fit your needs?
https://github.com/open-telemetry/opentelemetry-go/blob/main/example/prometheus/main.go

DestroyerAlpha:
We are on go1.18 so probably would have to upgrade the go version for that as well, since it requires go1.20. Do we see a future plan to integrate it inhouse?

rueian:
> Do we see a future plan to integrate it inhouse?

Sorry, I didn’t get you. What did you mean by that? Or what kind of functionalities you need to integrate with your in-house infrastructure?

DestroyerAlpha:
I meant do we plan to export prom metric internally from rueidis instead of using an exporter?

rueian:
@DestroyerAlpha,

Oh, I see. Yes, we can have a `rueidisprom` subpackage if we keep the core rueidis depending on fewer packages.

So, there are two things we can do:

1. Introduce the `rueidisprom` subpackage with its own go.mod separated from the root one.
2. Let `rueidisotel` have its own go.mod (https://github.com/redis/rueidis/issues/447).

Would you like to help the first one?

DestroyerAlpha:
Sure I can help with the first one

DestroyerAlpha:
Added a PR for the same @rueian

rueian:
Hi @DestroyerAlpha,

I would like to release the next v1.0.28 recently. Is there any chance of bringing the real implementation rueidisprom subpackage to life? Or I may need to revert it when releasing the next version.

DestroyerAlpha:
@rueian Hey would raise a PR for this ASAP.

PR #448, opened 2024-01-13
Introduce rueidisprom subpackage

DestroyerAlpha:
Introduce a new subpackage for adding support for prometheus metrics in Rueidis.
Add a separate go.mod file for it.
#444


rueian:
Hi @DestroyerAlpha, would you also add the implementation?

DestroyerAlpha:
@rueian How about doing it in a separate PR? We can merge this

rueian:
Ok

PR #458, opened 2024-01-26
fix: rueidislock deadlock caused by the absence of peers

### LINKED ISSUE #457: occurs deadlock when using rueidislock
`m.mu.Lock()` occurs deadlock when executing `locker.waitgate` once.
image

But It didn't occurs again.

rueidis version is `v1.0.22-go1.18`

rueian:
fixes #457 

There is a rare chance that when receiving `bulking invalidation` and all peers are aborted, the following line https://github.com/redis/rueidis/blob/33e9233955925a4e66fe2d8613443c4ca95ef5df/rueidislock/lock.go#L324 blocks indefinitely and causes deadlock reported in the #457.

This PR removes the unnecessary notification to the `g.ch` from the `bulking invalidation` and makes the mutex never be blocked.

### DIFF
--- a/rueidislock/lock.go
+++ b/rueidislock/lock.go
@@ -296,6 +296,7 @@ func (m *locker) try(ctx context.Context, cancel context.CancelFunc, name string
 		if released := atomic.AddInt32(&released, 1); released >= m.majority {
 			cancel()
 			if released == m.totalcnt {
+				close(done)
 				m.mu.Lock()
 				if g.w--; g.w == 0 {
 					if m.gates[name] == g {
@@ -308,7 +309,6 @@ func (m *locker) try(ctx context.Context, cancel context.CancelFunc, name string
 					}
 				}
 				m.mu.Unlock()
-				close(done)
 			}
 		}
 	}
--- a/rueidislock/lock.go
+++ b/rueidislock/lock.go
@@ -5,7 +5,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"math/rand"
-	"reflect"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


rueidislock/lock.go
```diff
@@ -234,10 +232,6 @@ func (m *locker) onInvalidations(messages []rueidis.RedisMessage) {
 	if messages == nil {
 		m.mu.RLock()
 		for _, g := range m.gates {
-			select {
-			case g.ch <- struct{}{}:
```
rueian, line 238: This may cause the below `g.ch <- struct{}{}` below stuck.

rueidislock/lock.go
```diff
@@ -304,28 +296,19 @@ func (m *locker) try(ctx context.Context, cancel context.CancelFunc, name string
...
 			cancel()
 			if released == m.totalcnt {
-				if atomic.LoadInt32(&failures) >= m.majority {
-					timer := time.NewTimer(m.interval)
-					cases := make([]reflect.SelectCase, 0, m.totalcnt+1)
-					cases = append(cases, reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(timer.C)})
-					for i := int32(0); i < m.totalcnt; i++ {
-						cases = append(cases, reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(g.csc[i])})
-					}
-					reflect.Select(cases)
```
rueian, line 314: This is no longer needed because we already clear each of the`g.csc` channel before entering the `monitoring` function.

rueidislock/lock.go
```diff
@@ -304,28 +296,19 @@ func (m *locker) try(ctx context.Context, cancel context.CancelFunc, name string
...
 				if g.w--; g.w == 0 {
 					if m.gates[name] == g {
 						delete(m.gates, name)
 					}
-				} else {
-					if g, ok := m.gates[name]; ok {
-						g.ch <- struct{}{}
+				} else if m.gates != nil {
+					select {
+					case g.ch <- struct{}{}:
```
rueian, line 307: Notify one peer if any without blocking.

PR #459, opened 2024-01-27
Feat: Align `go-redis` `JSONCmdable`

unknowntpo:
This PR implemented adapter for `go-redis` `JSONCmdable`.

### DIFF
--- a/rueidiscompat/command.go
+++ b/rueidiscompat/command.go
@@ -524,8 +524,8 @@ func newSliceCmd(res rueidis.RedisResult, isJSONObjKeys bool, keys ...string) *S
 		return cmd
 	}
 	vals := make([]any, len(arr))
-	for i, v := range arr {
-		if isJSONObjKeys {
+	if isJSONObjKeys {
+		for i, v := range arr {
 			// for JSON.OBJKEYS
 			if v.IsNil() {
 				continue
@@ -537,11 +537,14 @@ func newSliceCmd(res rueidis.RedisResult, isJSONObjKeys bool, keys ...string) *S
 				return cmd
 			}
 			vals[i] = arr
-		} else {
-			// keep the old behavior the same as before (don't handle error while parsing v as string)
-			if s, err := v.ToString(); err == nil {
-				vals[i] = s
-			}
+		}
+		cmd.SetVal(vals)
+		return cmd
+	}
+	for i, v := range arr {
+		// keep the old behavior the same as before (don't handle error while parsing v as string)
+		if s, err := v.ToString(); err == nil {
+			vals[i] = s
 		}
 	}
 	cmd.SetVal(vals)
--- a/rueidiscompat/command.go
+++ b/rueidiscompat/command.go
@@ -3515,10 +3515,6 @@ func newJSONCmd(res rueidis.RedisResult) *JSONCmd {
 	cmd := &JSONCmd{}
 	msg, err := res.ToMessage()
 	if err != nil {
-		if err == rueidis.Nil {
-			cmd.SetVal("")
-			return cmd
-		}
 		cmd.SetErr(err)
 		return cmd
 	}
@@ -3548,6 +3544,9 @@ func newJSONCmd(res rueidis.RedisResult) *JSONCmd {
 		for i, e := range arr {
 			anyE, err := e.ToAny()
 			if err != nil {
+				if err == rueidis.Nil {
+					continue
+				}
 				cmd.SetErr(err)
 				return cmd
 			}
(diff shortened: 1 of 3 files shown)


rueidiscompat/command.go
```diff
@@ -509,44 +510,49 @@ func newStatusCmd(res rueidis.RedisResult) *StatusCmd {
...
+func newSliceCmd(res rueidis.RedisResult, isJSONObjKeys bool, keys ...string) *SliceCmd {
+	cmd := &SliceCmd{keys: keys}
+	arr, err := res.ToArray()
+	if err != nil {
+		cmd.SetErr(err)
+		return cmd
+	}
+	vals := make([]any, len(arr))
+	for i, v := range arr {
+		if isJSONObjKeys {
```
rueian, line 529: It will be better to move `if isJSONObjKeys {` to the outside of the loop.

rueidiscompat/command.go
```diff
@@ -509,44 +510,49 @@ func newStatusCmd(res rueidis.RedisResult) *StatusCmd {
...
+			// for JSON.OBJKEYS
+			if v.IsNil() {
+				vals[i] = nil
```
rueian, line 532: This is unnecessary, right?

rueidiscompat/command.go
```diff
@@ -509,44 +510,49 @@ func newStatusCmd(res rueidis.RedisResult) *StatusCmd {
...
+				continue
+			}
+			// convert to any which underlying type is []any
+			arr, err := v.ToAny()
```
rueian, line 536:
```suggestion
			vals[i], err = v.ToAny()
```

rueidiscompat/command.go
```diff
@@ -3452,3 +3462,168 @@ type TSMGetOptions struct {
...
+func (cmd *JSONCmd) Result() (string, error) {
+	return cmd.Val(), nil
+}
+
+func newJSONCmd(res rueidis.RedisResult) *JSONCmd {
+	cmd := &JSONCmd{}
+	msg, err := res.ToMessage()
+	if err != nil {
+		if err == rueidis.Nil {
+			cmd.SetVal("")
```
rueian, line 3518: why cmd.SetVal("") when err == rueidis.Nil ?

rueidiscompat/command.go
```diff
@@ -3452,3 +3462,168 @@ type TSMGetOptions struct {
...
+	cmd := &IntPointerSliceCmd{}
+	arr, err := res.ToArray()
+	if err != nil {
+		cmd.SetErr(err)
+		return cmd
+	}
+	intPtrSlice := make([]*int64, len(arr))
+	for i, e := range arr {
+		if e.IsNil() {
+			intPtrSlice[i] = nil
```
rueian, line 3589: This also seems unnecessary.

rueidiscompat/command.go
```diff
@@ -3452,3 +3462,168 @@ type TSMGetOptions struct {
...
+	cmd := &JSONSliceCmd{}
+	arr, err := res.ToArray()
+	if err != nil {
+		cmd.SetErr(err)
+		return cmd
+	}
+	anySlice := make([]any, len(arr))
+	for i, e := range arr {
+		if e.IsNil() {
+			anySlice[i] = nil
```
rueian, line 3617: This also seems unnecessary.

rueidiscompat/command.go
unknowntpo: It's necessary because the return value might be `Nil`

e.g. 

```
redis> JSON.SET doc $ '{"a":[3], "nested": {"a": {"b":2, "c": 1}}}'
OK
redis> JSON.OBJKEYS doc $..a
1) (nil)
2) 1) "b"
   2) "c"
```

Ref: [`JSON.OBJKEYS`](https://redis.io/commands/json.objkeys/)

rueidiscompat/command.go
unknowntpo: I think I'm wrong.

There are two command using `JSONCmd` as returned value: `JSON.NUMINCRBY`, `JSON.GET`:

> `JSON.NUMINCRBY` returns a bulk string reply specified as a stringified new value for each path, or `nil`, if the matching JSON value is not a number.

> `JSON.GET` returns a bulk string representing a JSON array of string replies. Each string is the JSON serialization of each JSON value that matches a path.

According to the documentation, only `JSON.NUMINCRBY` will return `nil`, and when it's nil, we should return `rueidis.Nil` as `err`, not empty string.

rueidiscompat/command.go
unknowntpo: Do you mean `intPtrSlice[i] = nil` is unnecessary because element of intPtrSlice has default value `nil` ?

rueian: Yes.

rueidiscompat/command.go
unknowntpo: I get your point, `vals[i]` is `nil` by default, no need to set it, I'll fix that.

PR #485, opened 2024-03-01
add IsBusyGroup() to rueidis.RedisError

jhq0113:
When executing the `XGROUP CREATE <key> <group> <id> ` command, if group already exists, Redis will return BUSYGROUP..., Generally, the caller BUSYGROUP needs to distinguish from other errors, and adding this judgment can facilitate differentiation from other errors


rueian:
Cool! Thanks @jhq0113!

Would you like to also add a global helper `IsRedisBusyGroup`? I guess that could help further streamline the developer experience:

```go
err := client.Do(context.Background(), client.B().XgroupCreate().Key("").Group("").Id("").Build()).Error()
if rueidis.IsRedisBusyGroup(err) {
	// ...
}
```

jhq0113:
Yes, I do

rueian (approved):

PR #491, opened 2024-03-08
add redis-cell  command CL.THROTTLE

qi-hua:
add redis-cell  command CL.THROTTLE


rueian:
Thank you so much @qi-hua! I was not aware of this command before.

PR #492, opened 2024-03-09
feat: improve client instantiation examples

### LINKED ISSUE #435: Add documentation regarding Replica nodes
We seem to have added support for Replica Reads in ClusterClient. The documentation for how to properly set it up is still missing. Adding it would help consumers from integrating it effectively.

erdemtuna:
Resolves #435.
- I restructured the document so that client instantiation is placed at the top. In the old version, the section was at the very bottom, which was inconvenient for a new reader.
- I also improved the URL build options text by looking at `rueidis/url.go`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ A fast Golang Redis client that does auto pipelining and supports server-assiste
 * Pub/Sub, Sharded Pub/Sub, Streams
 * Redis Cluster, Sentinel, RedisJSON, RedisBloom, RediSearch, RedisTimeseries, etc.
 
+---
+
 ## Getting Started
 
 ```golang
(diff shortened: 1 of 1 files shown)


rueian:
Hi @erdemtuna, thank you very much. The content of these two sections looks good to me.

However, I put them at the bottom of the README intentionally because I like to prioritize the command builder and the auto pipelining. I believe the getting started section is already sufficient for new readers and they can find more details later at the bottom.

If we really need to put these initialization sections to the top, I think we had better further shorten their paragraph and merge them into the getting started section, or we can just leave them at the bottom.

erdemtuna:
Hi @rueian, thank you for the feedback. I will try to come up with another alternative based on your concerns.

erdemtuna:
@rueian what do you think about the new version?

rueian:
Thanks! That looks good!

PR #493, opened 2024-03-10
feat: rueidisprob package and standard bloom filter

proost:
discussion: https://github.com/redis/rueidis/discussions/486

standard bloom filter for `rueidisprob`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,14 @@ jobs:
             for n in {1..5}; do ../dockertest.sh $list && break; done
           no_output_timeout: 15m
 
+      - run: # test ./rueidisprob/go.mod
+          command: |
+            cd $CIRCLE_WORKING_DIRECTORY/rueidisprob
+            list=$(go list ./... | circleci tests split --split-by=timings)
+            echo "Test Packages: $list"
+            for n in {1..5}; do ../dockertest.sh $list && break; done
+          no_output_timeout: 15m
+
       - store_test_results:
           path: .
       - run: curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x codecov && ./codecov -t ${CODECOV_TOKEN}
--- a/rueidisprob/bloomfilter.go
+++ b/rueidisprob/bloomfilter.go
@@ -29,12 +29,12 @@ for i=2, numElements+1 do
     table.insert(bitfieldArgs, '1')
 end
 
-local setBits = redis.call('BITFIELD', unpack(bitfieldArgs))
+local bitset = redis.call('BITFIELD', unpack(bitfieldArgs))
 
 local counter = 0
 local oneBits = 0
-for i=1, #setBits do
-	oneBits = oneBits + setBits[i]
+for i=1, #bitset do
+	oneBits = oneBits + bitset[i]
 
 	if i % hashIterations == 0 then
 		if oneBits ~= hashIterations then
@@ -49,12 +49,12 @@ return redis.call('INCRBY', counterKey, counter)
 `
 
 	existsMultiScript = `
-local numElements = tonumber(#ARGV)
+local hashIterations = tonumber(ARGV[1])
+local numElements = tonumber(#ARGV) - 1
 local filterKey = KEYS[1]
-local result = {}
 
 local bitfieldArgs = { filterKey }
-for i=1, numElements do
+for i=2, numElements+1 do
 	local index = tonumber(ARGV[i])
 
 	table.insert(bitfieldArgs, 'GET')
@@ -62,7 +62,25 @@ for i=1, numElements do
 	table.insert(bitfieldArgs, index)
 end
 
-return redis.call('BITFIELD', unpack(bitfieldArgs))
+local bitset = redis.call('BITFIELD', unpack(bitfieldArgs))
+
+local result = {}
+local oneBits = 0
+for i=1, #bitset do
+	oneBits = oneBits + bitset[i]
+
+	if i % hashIterations == 0 then
+		if oneBits == hashIterations then
+			table.insert(result, 1)
+		else
+			table.insert(result, 0)
+		end
+
+		oneBits = 0
+	end
+end
+
+return result
 `
 
 	resetScript = `
@@ -249,7 +267,11 @@ func (c *bloomFilter) ExistsMulti(ctx context.Context, keys []string) ([]bool, e
 
 	indexes := c.indexes(keys)
 
-	resp := c.existsMultiScript.Exec(ctx, c.client, c.existsMultiKeys, indexes)
+	var args []string
+	args = append(args, c.hashIterationString)
+	args = append(args, indexes...)
+
+	resp := c.existsMultiScript.Exec(ctx, c.client, c.existsMultiKeys, args)
 	if resp.Error() != nil {
 		return nil, resp.Error()
 	}
@@ -260,22 +282,8 @@ func (c *bloomFilter) ExistsMulti(ctx context.Context, keys []string) ([]bool, e
 	}
 
 	result := make([]bool, len(keys))
-	resultIdx := 0
-	oneBits := 0
-	iterations := int(c.hashIterations)
 	for i, v := range is {
-		if v == 1 {
-			oneBits++
-		}
-
-		if (i+1)%iterations == 0 {
-			if oneBits == iterations {
-				result[resultIdx] = true
-			}
-
-			resultIdx++
-			oneBits = 0
-		}
+		result[i] = v == 1
 	}
 	return result, nil
 }
--- a/rueidisprob/bloomfilter.go
+++ b/rueidisprob/bloomfilter.go
@@ -238,10 +238,11 @@ func (c *bloomFilter) AddMulti(ctx context.Context, keys []string) error {
 
 func (c *bloomFilter) indexes(keys []string) []string {
 	allIndexes := make([]string, 0, len(keys)*int(c.hashIterations))
+	size := uint64(c.size)
 	for _, key := range keys {
-		indices := indexes([]byte(key), c.hashIterations, c.size)
-		for _, index := range indices {
-			allIndexes = append(allIndexes, strconv.FormatUint(index, 10))
+		h1, h2 := hash([]byte(key))
+		for i := uint(0); i < c.hashIterations; i++ {
+			allIndexes = append(allIndexes, strconv.FormatUint(index(h1, h2, i, size), 10))
 		}
 	}
 	return allIndexes
--- a/rueidisprob/bloomfilter.go
+++ b/rueidisprob/bloomfilter.go
@@ -144,10 +144,12 @@ func NewBloomFilter(
 	}
 	hashIterations := numberOfHashFunctions(size, expectedNumberOfItems)
 
+	// NOTE: https://redis.io/docs/reference/cluster-spec/#hash-tags
+	namespace := "{" + name + "}"
 	return &bloomFilter{
 		client:         client,
-		name:           name,
-		counter:        name + ":c",
+		name:           namespace,
+		counter:        namespace + ":c",
 		hashIterations: hashIterations,
 		size:           size,
 	}, nil
--- a/rueidisprob/bloomfilter.go
+++ b/rueidisprob/bloomfilter.go
@@ -1,0 +1,308 @@
+package rueidisprob
+
+import (
+	"context"
+	"errors"
+	"math"
+	"strconv"
+
+	"github.com/redis/rueidis"
+)
+
+const (
+	// NOTE: https://redis.io/docs/data-types/bitmaps/
+	maxSize = 1 << 32
+)
+
+const (
+	addMultiScript = `
+local count = tonumber(ARGV[1])
+local numElements = tonumber(#ARGV) - 1
+local filterKey = KEYS[1]
+local counterKey = KEYS[2]
+
+for i = 2, numElements + 1 do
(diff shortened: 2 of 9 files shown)


rueian:
Hi @proost,

There are a couple of things we can address and discuss about:

1. Please add a test entry to the .circleci/config.yml, for example https://github.com/redis/rueidis/blob/b08dc62ed45c49f607e44d3af5f9bfc4e92fe21f/.circleci/config.yml#L61-L67
2. The filter key and the counter key should be located in the same Redis slot by using a [tag](https://redis.io/docs/reference/cluster-spec/#hash-tags) so that these Lua scripts can work in cluster mode.
3. The current implementation increments the counter no matter what, but that is inaccurate. Can we check the SETBIT results before incrementing the count? In other words, if the sum of the return values of SETBIT equals to the number of hash iterations, we should not increment the counter.
4. We use SETBIT one by one for each index. Can we make it a batch for a key with BITFIELD + multiple SETs?
5. We should not use EVAL every time. That hurts performance. Can we leverage the `rueidis.NewLuaScript`?

proost:
@rueian 
Yes, you're right. I was wrong many points.

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,344 @@
...
+end
+
+local setBits = redis.call('BITFIELD', unpack(bitfieldArgs))
+
+local counter = 0
+local oneBits = 0
+for i=1, #setBits do
+	if setBits[i] == 1 then
+		oneBits = oneBits + 1
+	end
```
rueian, line 39:
```suggestion
	oneBits = oneBits + setBits[i]
```
Could we just do this?

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,344 @@
...
+	return &bloomFilter{
+		client:              client,
+		name:                bfName,
+		counter:             counterName,
+		hashIterations:      hashIterations,
+		hashIterationString: strconv.FormatUint(uint64(hashIterations), 10),
+		size:                size,
+		addMultiScript:      rueidis.NewLuaScript(addMultiScript),
+		addMultiKeys:        []string{bfName, counterName},
+		existsMultiScript:   rueidis.NewLuaScript(existsMultiScript),
```
rueian, line 191:
```suggestion
		existsMultiScript:   rueidis.NewLuaScriptReadOnly(existsMultiScript),
```

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,344 @@
...
+		existsMultiKeys:     []string{bfName},
+	}, nil
+}
+
+func numberOfBits(n uint, r float64) uint {
+	if r == 0 {
+		r = math.SmallestNonzeroFloat64
```
rueian, line 198: Why don't we just reject any r which is smaller than math.SmallestNonzeroFloat64?

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,344 @@
...
+local bitfieldArgs = { filterKey }
+for i=1, numElements do
+	local index = tonumber(ARGV[i])
+
+	table.insert(bitfieldArgs, 'GET')
+	table.insert(bitfieldArgs, 'u1')
+	table.insert(bitfieldArgs, index)
+end
+
+return redis.call('BITFIELD', unpack(bitfieldArgs))
```
rueian, line 67: I am curious about whether would it be better if we calculate key existences here and return a bool array only instead of returning every bits? I am not sure which one will be better.

rueidisprob/bloomfilter.go
proost: [299d097](https://github.com/redis/rueidis/pull/493/commits/299d09754639f4f78bdb8c04906755a382c0467f)

rueidisprob/bloomfilter.go
proost: `EVAL_RO` and `EVALSHA_RO` support since from 7.0.0. is it too narrow to support various redis version?

rueidisprob/bloomfilter.go
proost: [e7cd23c](https://github.com/redis/rueidis/pull/493/commits/e7cd23c9e2374c581482087955d2c9ac19c8b7b9)

rueidisprob/bloomfilter.go
proost: there is trade-off. if we calculate on redis, we should use more CPU, but it can reduce bandwidth per request. In this situation, I prefer use less CPU. because bit array size is small event if keys and hash iteration size is large. but if you think bandwidth is more valuable than CPU, i will change it.

rueian: We can leave this for future evaluation. I may be wrong, but my guess is that the calculation on redis is cheap because it just loops through the return of BITFIELD and sums up bits for every hash iteration. If we do that, we can save the bandwidth by hash iteration times. This requires more evaluation for verification.

I also have another idea: Can we use a normal GET to fetch the whole bitmap and do the calculation on the client side? By that, we can leverage client-side caching.

rueidisprob/bloomfilter.go
rueian: Okay, that is too narrow. Never mind.

rueidisprob/bloomfilter.go
proost: [7491a90](https://github.com/redis/rueidis/pull/493/commits/7491a90298299bc8f7903b0f6369ed8b54267fb3)

OK. I change it.

Still, I don't think it is good to depend on Lua script heavily. scaling out servers is easy, but scaling out redis is not easy. 
I found another implementation of bloom filter using redis: https://github.com/igrigorik/bloomfilter-rb

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,347 @@
...
+	var args []string
+	args = append(args, c.hashIterationString)
+	args = append(args, indexes...)
+
+	resp := c.existsMultiScript.Exec(ctx, c.client, c.existsMultiKeys, args)
+	if resp.Error() != nil {
+		return nil, resp.Error()
+	}
+
+	is, err := resp.AsIntSlice()
```
rueian, line 279:
```suggestion
	is, err := resp.ToArray()
```

`AsIntSlice` will allocate a new slice which is costly if it is a temporary storage to a bool slice. Unfortunately, we don't have `AsBoolSlice` yet. We can instead use `ToArray` to avoid allocation.

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,347 @@
...
+	if len(keys) == 0 {
+		return nil
+	}
+
+	indexes := c.indexes(keys)
+
+	var args []string
+	args = append(args, c.hashIterationString)
+	args = append(args, indexes...)
+
```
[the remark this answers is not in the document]
proost: [8941769](https://github.com/redis/rueidis/pull/493/commits/8941769769a987aa27969ee2f1bdc40319f7ec58)

rueidisprob/bloomfilter.go
proost: [d037dad](https://github.com/redis/rueidis/pull/493/commits/d037dadc6557908a437bdca34b87112d59ac6d1d)

Oh, Thank you!

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,347 @@
...
+local oneBits = 0
+for i=1, #bitset do
+	oneBits = oneBits + bitset[i]
+
+	if i % hashIterations == 0 then
+		if oneBits == hashIterations then
+			table.insert(result, 1)
+		else
+			table.insert(result, 0)
+		end
```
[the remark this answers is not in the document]
proost: lua `false` is nil in redis. It can be confused.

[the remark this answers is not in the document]
rueian: Returning Nils is actually good I think. It requires fewer bytes to transmit. A nil message takes 3 bytes to transmit, but an integer 0 takes 4 bytes.

rueidisprob/bloomfilter.go
```diff
@@ -0,0 +1,352 @@
...
+		return resp.Error()
+	}
+
+	return nil
+}
+
+func (c *bloomFilter) indexes(keys []string) []string {
+	allIndexes := make([]string, 0, len(keys)*int(c.hashIterations))
+	for _, key := range keys {
+		indices := indexes([]byte(key), c.hashIterations, c.size)
```
rueian, line 246: Can we reuse the `indices` for every key to save some allocations? Or how about letting the `indexes` function write to the `allIndexes` directly?

rueidisprob/bloomfilter.go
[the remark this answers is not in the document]
proost: [56f7d9d](https://github.com/redis/rueidis/pull/493/commits/56f7d9d228d613c6500e3a0dc1367a8bd53e39a2). I changed it.

rueidisprob/bloomfilter.go
proost: [56fc836](https://github.com/redis/rueidis/pull/493/commits/56fc836c835bd589c7bbcc61d063e85835f59cb2)

rueidisprob/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/redis/rueidis/rueidisprob
+
+go 1.22.0
```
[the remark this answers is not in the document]
proost: [4ed476f](https://github.com/redis/rueidis/pull/493/commits/4ed476fab06c336fb3a46f6908df27a3af96de0d)

rueian:
Hi @proost, thanks! This looks good to me. I also made a small benchmark between the latest commit and the previous commit d118842c3649:

```
goos: darwin
goarch: arm64
pkg: cmd/bloom
    │ d118842c3649    │        feat-rueidisprob         │
    │     sec/op      │   sec/op     vs base            │
-10       48.08µ ± 0%   48.12µ ± 0%  ~ (p=0.183 n=20)

    │ d118842c3649    │            feat-rueidisprob            │
    │      B/op       │     B/op      vs base                  │
-10      8.749Ki ± 1%   3.621Ki ± 0%  -58.61% (p=0.000 n=20)

    │ d118842c3649     │           feat-rueidisprob           │
    │    allocs/op     │ allocs/op   vs base                  │
-10        90.00 ± 0%   74.00 ± 0%  -17.78% (p=0.000 n=20)
```

The CPU loading of the Redis server in these two cases was both ~97% and ~98%. The latest feat-rueidisprob had slightly heavy CPU loading on the Redis, but it had -58.61% and -17.78% reduction in `B/op` and `allocs/op`.

Benchmark Code:
```go
package main

import (
	"context"
	"math/rand/v2"
	"strconv"
	"testing"

	"github.com/redis/rueidis"
	"github.com/redis/rueidis/rueidisprob"
)

func Benchmark(b *testing.B) {
	c, err := rueidis.NewClient(rueidis.ClientOption{InitAddress: []string{"127.0.0.1:6379"}})
	if err != nil {
		panic(err)
	}
	defer c.Close()

	bf, err := rueidisprob.NewBloomFilter(c, "bf", 1000000, 0.01)
	if err != nil {
		panic(err)
	}

	if err := bf.Add(context.Background(), "1"); err != nil {
		panic(err)
	}

	keys := make([]string, 10)
	for i := range keys {
		keys[i] = strconv.Itoa(rand.IntN(10000))
	}

	b.ReportAllocs()
	b.ResetTimer()
		for pb.Next() {
			if _, err := bf.ExistsMulti(context.Background(), keys); err != nil {
				panic(err)
			}
		}
	})
	b.StopTimer()
}

```

proost:
@rueian 
[a2df9bb](https://github.com/redis/rueidis/pull/493/commits/a2df9bba11cefe16b578f7e9ae2847cc660dfc35)
I want to add this package to main README file. is it ok?

rueian:
> @rueian [a2df9bb](https://github.com/redis/rueidis/pull/493/commits/a2df9bba11cefe16b578f7e9ae2847cc660dfc35) I want to add this package to main README file. is it ok?

Sure. Please go ahead.

.circleci/config.yml
```diff
@@ -66,6 +66,14 @@ jobs:
             for n in {1..5}; do ../dockertest.sh $list && break; done
           no_output_timeout: 15m
 
+      - run: # test ./rueidisprob/go.mod
+          command: |
+            cd $CIRCLE_WORKING_DIRECTORY/rueidisprob
+            list=$(go list ./... | circleci tests split --split-by=timings)
+            echo "Test Packages: $list"
+            for n in {1..5}; do ../dockertest.sh $list && break; done
```
rueian, line 74: I found that test failures may be caused by not yet initialized redis cluster. Adding a `sleep 5` after `docker-compose up -d` in the `dockertest.sh` could be helpful.

rueian (commented):
I found that test failures may be caused by not yet initialized redis cluster. Adding a sleep 5 after docker-compose up -d in the dockertest.sh could be helpful.

.circleci/config.yml
proost: [cfcfad5](https://github.com/redis/rueidis/pull/493/commits/cfcfad548c2dfaa251dec08cc18c711352a7f87d)

Yes. some time is needed to cluster configuration. deterministic approach is required, later we can do it.

rueian (approved):
@proost LGTM. Thank you for this great work!

proost:
@rueian 
Thank you for accepting proposal! I will send Counting Bloom Filter Design to Discussion soon.

PR #518, opened 2024-03-28
fix rueidisprob readme

trim21:


rueidisprob/README.md
```diff
@@ -17,40 +17,43 @@ Example:
...
-    }
-    fmt.Println(exists) // true
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	bf, err := rueidisprob.NewBloomFilter(client, "bloom_filter", 1000, 0.01)
+
+	err = bf.Add(context.TODO(), "hello")
```
rueian, line 37:
```suggestion
	err = bf.Add(context.Background(), "hello")
```

rueian: Thanks @trim21! Is there a specific reason for using `context.TODO()`? Or a `context.Background()` is good and clear.

trim21: it's shorter

trim21: I don't think Background and TODO have any difference, I can change it to Background if you prefer

rueian: I understand that they have no difference except for their semantics. I prefer Background over TODO because the latter looks so unfinished.

PR #519, opened 2024-03-31
feat: add counting bloom filter

proost:
Previous Discussion: https://github.com/redis/rueidis/discussions/510

There are 2 different points from origin design.

1. `Count` returns all inserted items. not distinct items.
2. `RemoveMulti` deduplicate keys. count in Counting Bloom Filter can't be negative. because of these nature, `RemoveMulti` operation can be complicated and performance can be bad. so add constraints to avoid difficulties.

### DIFF
--- a/rueidisprob/countingbloomfilter.go
+++ b/rueidisprob/countingbloomfilter.go
@@ -291,14 +291,12 @@ func (f *countingBloomFilter) ExistsMulti(ctx context.Context, keys []string) ([
 	args = append(args, indexes...)
 
 	resp := f.existsMultiScript.Exec(ctx, f.client, f.existsMultiKeys, args)
-	if resp.Error() != nil {
-		return nil, resp.Error()
-	}
 
 	arr, err := resp.AsBoolSlice()
 	if err != nil {
 		return nil, err
 	}
+
 	return arr, nil
 }
 
@@ -358,16 +356,12 @@ func (f *countingBloomFilter) Count(ctx context.Context) (uint, error) {
 			Key(f.counter).
 			Build(),
 	)
-	if resp.Error() != nil {
-		if rueidis.IsRedisNil(resp.Error()) {
+	count, err := resp.AsUint64()
+	if err != nil {
+		if rueidis.IsRedisNil(err) {
 			return 0, nil
 		}
 
-		return 0, resp.Error()
-	}
-
-	count, err := resp.AsUint64()
-	if err != nil {
 		return 0, err
 	}
 
--- a/rueidisprob/countingbloomfilter.go
+++ b/rueidisprob/countingbloomfilter.go
@@ -125,7 +125,6 @@ type CountingBloomFilter interface {
 	Add(ctx context.Context, key string) error
 
 	// AddMulti adds one or more items to the Counting Bloom Filter.
-	// If there are duplicate keys, they are deduplicated.
 	// NOTE: If keys are too many, it can block the Redis server for a long time.
 	AddMulti(ctx context.Context, keys []string) error
 
@@ -236,19 +235,10 @@ func (f *countingBloomFilter) AddMulti(ctx context.Context, keys []string) error
 		return nil
 	}
 
-	deduplicatedKeys := make([]string, 0, len(keys))
-	keySet := make(map[string]struct{}, len(keys))
-	for _, key := range keys {
-		if _, ok := keySet[key]; !ok {
-			keySet[key] = struct{}{}
-			deduplicatedKeys = append(deduplicatedKeys, key)
-		}
-	}
-
-	indexes := f.indexes(deduplicatedKeys)
+	indexes := f.indexes(keys)
 
 	args := make([]string, 0, len(indexes)+1)
-	args = append(args, strconv.Itoa(len(deduplicatedKeys)))
+	args = append(args, strconv.Itoa(len(keys)))
 	args = append(args, indexes...)
 
 	resp := f.addMultiScript.Exec(ctx, f.client, f.addMultiKeys, args)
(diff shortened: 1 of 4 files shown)


rueidisprob/countingbloomfilter.go
```diff
@@ -0,0 +1,375 @@
...
+			return 0, nil
+		}
+
+		return 0, resp.Error()
+	}
+
+	count, err := resp.AsUint64()
+	if err != nil {
+		return 0, err
+	}
```
rueian, line 372:
```suggestion
	count, err := resp.AsUint64()
	if err != nil {
		if rueidis.IsRedisNil(err) {
			return 0, nil
		}
		return 0, err
	}
```
There is no need to use `resp.Error()` separately. The  `resp.AsUint64()` should also return the same err.

rueian:
> Count returns all inserted items. not distinct items.

I guess, a method for getting the distinct count is still needed for users to know whether this filter is close to full or not.

And instead of `Exists`, should we provide a method for getting the minimum count of a key? Otherwise, users may just use the standard bloom filter.

rueidisprob/countingbloomfilter.go
proost: [7069d90](https://github.com/redis/rueidis/pull/519/commits/7069d905ebc9d7bcc0eed498d6755ef8735d752f)

proost:
@rueian 
> And instead of Exists, should we provide a method for getting the minimum count of a key? 
Do you want to remove `Exists` or Adding a new method? According to [definition](https://en.wikipedia.org/wiki/Counting_Bloom_filter), Counting Bloom Filter can insert same key several times.

rueian:
I feel the `Esists` is not necessary if we have a method to get the minimum count of a key, but we can keep it anyway. It is still a handy method.

rueidisprob/countingbloomfilter.go
```diff
@@ -0,0 +1,369 @@
...
+	args := make([]string, 0, len(indexes)+1)
+	args = append(args, strconv.Itoa(len(keys)))
+	args = append(args, indexes...)
+
+	resp := f.addMultiScript.Exec(ctx, f.client, f.addMultiKeys, args)
+	if resp.Error() != nil {
+		return resp.Error()
+	}
+
+	return nil
```
rueian, line 249:
```suggestion
	return resp.Error()
```
I am not sure if this kind of pattern can be optimized by the go compiler but we can just shortcut it by hand.

proost:
@rueian 
I understand what you said. If we had a method that returns minimum count of a key, it is count-min sketch. not counting bloom filter. do you think count-min sketch is more proper?

rueian:
>  do you think count-min sketch is more proper? or do you want to integrate with both data structures?

The wiki you referenced above also says that count-min sketch and counting bloom filter are essentially the same data structure. Could you elaborate more on their differences?

I have no preference between count-min sketch and counting bloom filter. I just want this filter to have more functionalities than the previous standard bloom filter.

proost:
> The wiki you referenced above also says that count-min sketch and counting bloom filter are essentially the same data structure.

what i understand is both data structures almost same logic
1. use hash function to given key to reduce space usage.
2. each hashed value store count.

but what i know purpose, functionalities, internal data structure is different.
In terms of purpose, Count-Min Sketch store and get item's frequency but Counting Bloom Filter check whether item is in or not.
In terms of internal data structure, Count-Min Sketch uses 2d array, but Counting Bloom Filter use 1d array.
In terms of functionalities, Count-Min Sketch explicitly supports inserting same data multiple, like this `update(data, delta)`. but Counting Bloom Filter doesn't. `AddMulti` is just batch function to use `Add` conveniently in this PR, but not originally created function in the paper.

proost:
> I have no preference between count-min sketch and counting bloom filter. I just want this filter to have more functionalities than the previous standard bloom filter.

I see. But this implementation may have bigger error rate than count-min sketch(This is uncertain.)

rueian:
> but Counting Bloom Filter check whether item is in or not.

This makes it look the same as the original bloom filter, except that it could still report `exists=true` even after removing an item (since the item could have been added multiple times). From this perspective, it seems that having a method to retrieve the minimum count of an item is still reasonable.

rueidisprob/countingbloomfilter.go
proost: [bc2f494](https://github.com/redis/rueidis/pull/519/commits/bc2f494e049ece15d4fd12635c18d55c207c7138)

proost:
@rueian 
[abda0c8](https://github.com/redis/rueidis/pull/519/commits/abda0c8a33f7864b466e9a92d45948e1797f1646)
Adding `ItemMinCount` and `ItemMinCountMulti`. But considering purpose of Counting Bloom Filter, leaving left `Exists` and `ExistsMulti` makes sense to me.

rueidisprob/countingbloomfilter.go
```diff
@@ -0,0 +1,411 @@
...
+	if len(keys) == 0 {
+		return nil
+	}
+
+	deduplicatedKeys := make([]string, 0, len(keys))
+	keySet := make(map[string]struct{}, len(keys))
+	for _, key := range keys {
+		if _, ok := keySet[key]; !ok {
+			keySet[key] = struct{}{}
+			deduplicatedKeys = append(deduplicatedKeys, key)
```
rueian, line 309: Should we do this deduplication? It feels to me that duplication should be allowed.

proost: [f959dc3](https://github.com/redis/rueidis/pull/519/commits/f959dc3203e4538e83bbb5d8c35e410a2d720b3e)

To mitigate complex logic, i added deduplication. But I remove it it is more complex and can downgrade performance.

rueian: I see. It is indeed complex but looks good to me now.

rueian (approved):

rueian:
Merged. Thank you @proost!

PR #522, opened 2024-04-01
feat: Add a helper generic function DecodeSliceOfJSON

### LINKED ISSUE #521: Add DecodeSliceOfJSON to handle []Struct results easier
(init idea: Add *RedisMessage.ScanSlice(dst any) to handle []Struct results easier)

When values of key1, key2 are both instances of X, code below returns an error of course because DecodeJSON handles only string

```go
type X struct {
	Name string
}

xs := make([]*X, 0)
if err := redisClient.Do(ctx, s.redisClient.B().Mget().Key("key1", "key2").Build()).DecodeJSON(xs); err != nil {
    return err
}
```

I could not find any method to scan a slice of struct so I would like to add one, like "ScanSlice". 
Scanning slice of struct in golang is not always easy so providing a scanning-slice method would be helpful.

some libraries, for example, sqlx.DB has a method "SelectContext" to scan a slice of struct and this is really helpful for us.

```go
xs := make([]*X, 0)
if err := sqlx.DB.SelectContext(ctx, &xs, q, args...); err != nil { return err }

// -> xs = []&X{{Name: "name1"}, {Name: "name2"}, ....}
```

sgasho:
added a helper function DecodeSliceOfJSON to struct scan a redis result.
code provided in https://github.com/redis/rueidis/issues/521#issuecomment-2029762737 did not work so I changed a little.

## Related Issue
close #521

### DIFF
--- a/helper.go
+++ b/helper.go
@@ -182,16 +182,16 @@ func JsonMSet(client Client, ctx context.Context, kvs map[string]string, path st
 }
 
 // DecodeSliceOfJSON is a helper that struct-scans each RedisMessage into dest, which must be a slice.
-func DecodeSliceOfJSON[T any](resp RedisResult, dest *[]T) error {
-	arr, err := resp.ToArray()
+func DecodeSliceOfJSON[T any](result RedisResult, dest *[]T) error {
+	values, err := result.ToArray()
 	if err != nil {
 		return err
 	}
 
-	ts := make([]T, len(arr))
-	for i, a := range arr {
+	ts := make([]T, len(values))
+	for i, v := range values {
 		var t T
-		if err = a.DecodeJSON(&t); err != nil && !IsRedisNil(err) {
+		if err = v.DecodeJSON(&t); err != nil && !IsRedisNil(err) {
 			return err
 		}
 		ts[i] = t
--- a/helper.go
+++ b/helper.go
@@ -182,19 +182,22 @@ func JsonMSet(client Client, ctx context.Context, kvs map[string]string, path st
 }
 
 // DecodeSliceOfJSON is a helper that struct-scans each RedisMessage into dest, which must be a slice of pointer.
-func DecodeSliceOfJSON[T any](result RedisResult, dest *[]*T) error {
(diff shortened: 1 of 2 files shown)


helper.go
```diff
@@ -181,6 +181,25 @@ func JsonMSet(client Client, ctx context.Context, kvs map[string]string, path st
...
+		return err
+	}
+
+	ts := make([]T, len(values))
+	for i, v := range values {
+		var t T
+		if err = v.DecodeJSON(&t); err != nil && !IsRedisNil(err) {
+			return err
+		}
+		ts[i] = t
```
rueian, line 197:
```suggestion
func DecodeSliceOfJSON[T any](result RedisResult, dest *[]*T) error {
	values, err := result.ToArray()
	if err != nil {
		return err
	}

	ts := make([]*T, len(values))
	for i, v := range values {
		var t T
		if err = v.DecodeJSON(&t); err != nil && !IsRedisNil(err) {
			return err
		}
		ts[i] = &t
```

Considering the case of Nil, I think accepting a `*[]*T` is still a better approach.

rueian:
image
The error handling parts is not tested. Could you also add some tests for them?

rueian:
Sorry @sgasho, I didn't notice that your original version can actually work with both `*[]T` and `*[]*T`. So, I now believe your original version is superior. Could you change it back?

rueian (approved):

helper.go
```diff
@@ -181,6 +181,28 @@ func JsonMSet(client Client, ctx context.Context, kvs map[string]string, path st
...
+	if err != nil {
+		return err
+	}
+
+	ts := make([]T, len(values))
+	for i, v := range values {
+		var t T
+		if err = v.DecodeJSON(&t); err != nil {
+			if IsRedisNil(err) {
+				continue
```
ash2k, line 196: @rueian @sgasho If this happens, then `ts[i]` will be an empty value. Is this the expected behavior? Probably not.

ash2k: Ah, I see there is a test for this behavior, so probably it is a desired behavior. Sorry for the noise.

PR #523, opened 2024-04-02
Add WithDBStatement to add db.statement attribute to rueidisotel

chkp-omris:
Add tracer option to add the whole command's statement to db.statement attribute.

Need this for integration with APM and couldn't find another way that doesn't involve copying a lot of code from `otelClient`'s implementation.

I mostly use non-multi operations and they seem a bit more complex and opinionated so I didn't implement it on my own.

### DIFF
--- a/rueidisotel/trace.go
+++ b/rueidisotel/trace.go
@@ -17,6 +17,7 @@ var (
 	name   = "github.com/redis/rueidis"
 	kind   = trace.WithSpanKind(trace.SpanKindClient)
 	dbattr = attribute.String("db.system", "redis")
+	dbstmt = attribute.Key("db.statement")
 )
 
 var _ rueidis.Client = (*otelclient)(nil)
@@ -50,6 +51,13 @@ func WithTracerProvider(provider trace.TracerProvider) Option {
 	}
 }
 
+// WithDBStatement tells the tracing hook to add raw redis commands to db.statement attribute.
+func WithDBStatement(enabled bool) Option {
+	return func(o *otelclient) {
+		o.dbStmtEnabled = enabled
+	}
+}
+
 type otelclient struct {
 	client          rueidis.Client
 	meterProvider   metric.MeterProvider
@@ -61,6 +69,7 @@ type otelclient struct {
 	mAttrs          metric.MeasurementOption
 	tAttrs          trace.SpanStartEventOption
 	histogramOption HistogramOption
+	dbStmtEnabled   bool
 }
 
 func (o *otelclient) B() rueidis.Builder {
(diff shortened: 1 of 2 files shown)


rueian:
Hi @chkp-omris,

Unlike other database systems, capturing the full redis command into `db.statement` is probably a terrible idea. The full command could not only be huge but also contain credentials that can't be leaked to APM.

Could you just capture at most two tokens from a command? For example, the `db.statement` of `SET mykey mysecret` command could be just `SET mykey`.

chkp-omris:
@rueian thanks for the quick response.
I agree it's not a great idea, I saw that `go-redis` implements this so thought `rueidis` can have that too.
I looked at `go-redis`'s implementation again, I missed that it does truncate/limit the statement.
`go-redis` [puts](https://github.com/redis/go-redis/blob/master/extra/redisotel/tracing.go#L127) the result of [`CmdString`](https://github.com/redis/go-redis/blob/master/extra/rediscmd/rediscmd.go#L13) and [`CmdsString`](https://github.com/redis/go-redis/blob/master/extra/rediscmd/rediscmd.go#L19) to `db.statement` attribute, which do limit the string's size.

I think it's reasonable to have the same limitation in `rueidis`'s implementation, what do you think?

rueian:
32 bytes capping is good, but I think we should also take only the first two tokens to avoid capturing any credentials.

chkp-omris:
@rueian First of all I hope you and your loved ones are safe after the earthquake in Taiwan.

> 32 bytes capping is good, but I think we should also take only the first two tokens to avoid capturing any credentials.

It depends on the user's use case.
In my case it isn't an issue, and of course for some it is an issue, but that's why this option is off by default.
I think it would be better to add this functionality and expand it for other use cases when requested.

Also, I suggest applying the 32 bytes cap to each token in the command, as there could be flags/options given that are important.
For example, when using `FT.SEARCH` the options given to it can change its meaning quite a lot.
We can cap the number of tokens it will take of the command as well to like 10.

Let me know what you think and I'll implement it.

chkp-omris:
> So, instead of `WithDBStatement(enabled bool)`, how about `WithDBStatement(stmtFn func(command []string) (stmt string))`?
> 
> In this way, users have full responsibility for the statement being sent.

Yea that's a good idea, done

rueidisotel/trace.go
```diff
@@ -83,6 +99,10 @@ func (o *otelclient) DoMulti(ctx context.Context, multi ...rueidis.Completed) (r
 
 func (o *otelclient) DoStream(ctx context.Context, cmd rueidis.Completed) (resp rueidis.RedisResultStream) {
 	ctx, span := o.start(ctx, first(cmd.Commands()), sum(cmd.Commands()), o.tAttrs)
+	if o.dbStmtFunc != nil {
+		span.SetAttributes(dbstmt.String(o.dbStmtFunc(cmd.Commands())))
```
rueian, line 103: Hi @chkp-omris, LGTM! But these SetAttributes are not covered by tests

image

Would you mind to add tests for them?

chkp-omris: @rueian Sure, I added the option to the unit tests for the client

rueian (approved):
👍

PR #524, opened 2024-04-03
docs: DecodeSliceOfJSON in README.md

sgasho:
related: #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,19 +490,19 @@ DecodeSliceOfJSON is useful when you would like to scan the results of an array 
 
 ```golang
 type User struct {
-	Name string
+	Name string `json:"name"`
 }
 
 // Set some values
-if err = client.Do(ctx, client.B().Set().Key("user1").Value(`{"Name": "name1"}`).Build()).Error(); err != nil {
+if err = client.Do(ctx, client.B().Set().Key("user1").Value(`{"name": "name1"}`).Build()).Error(); err != nil {
 	return err
 }
-if err = client.Do(ctx, client.B().Set().Key("user2").Value(`{"Name": "name2"}`).Build()).Error(); err != nil {
+if err = client.Do(ctx, client.B().Set().Key("user2").Value(`{"name": "name2"}`).Build()).Error(); err != nil {
 	return err
 }
 
 // Scan MGET results into []*User
-users := make([]*User, 0) // []User is also being scannable
+var users []*User // or []User is also scannable
 if err := rueidis.DecodeSliceOfJSON(client.Do(ctx, client.B().Mget().Key("user1", "user2").Build()), &users); err != nil {
 	return err
 }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -502,4 +582,4 @@ go generate
 ### Testing
 
 Please use the [./dockertest.sh](./dockertest.sh) script for running test cases locally.
-And please try your best to have 100% test coverage on code changes.
\ No newline at end of file
+And please try your best to have 100% test coverage on code changes.
```
sgasho, line 585: I have no idea the differences....

README.md
```diff
@@ -484,6 +484,86 @@ client.Do(ctx, client.B().FtSearch().Index("idx").Query("@f:v").Build()).AsFtSea
...
+	return err
+}
+
+
+users := make([]*User, 0)
+if err := rueidis.DecodeSliceOfJSON(client.Do(ctx, client.B().Mget().Key("user1").Build()), &users); err != nil {
+	return err
+}
+// users = [{name:name1}]
+```
```
sgasho, line 549: I wonder if this section is necessary...
for some experienced golang developer don't need it but some people like beginners or not familiar with golang might do.

README.md
```diff
@@ -484,6 +484,86 @@ client.Do(ctx, client.B().FtSearch().Index("idx").Query("@f:v").Build()).AsFtSea
...
+
+DecodeSliceOfJSON is useful when you would like to scan the results of an array into a slice of a specific struct.
+
+```golang
+type User struct {
+	Name string
+}
+
+// Set some values
+if err = client.Do(ctx, client.B().Set().Key("user1").Value(`{"Name": "name1"}`).Build()).Error(); err != nil {
```
rueian, line 497:
```suggestion
if err = client.Do(ctx, client.B().Set().Key("user1").Value(`{"name": "name1"}`).Build()).Error(); err != nil {
```

README.md
```diff
@@ -484,6 +484,86 @@ client.Do(ctx, client.B().FtSearch().Index("idx").Query("@f:v").Build()).AsFtSea
...
+	return err
+}
+if err = client.Do(ctx, client.B().Set().Key("user2").Value(`{"Name": "name2"}`).Build()).Error(); err != nil {
```
rueian, line 500:
```suggestion
if err = client.Do(ctx, client.B().Set().Key("user2").Value(`{"name": "name2"}`).Build()).Error(); err != nil {
```

README.md
rueian: Indeed, I think this section is not necessary, and scanning into private fields is also not a common approach. Actually using `json` tags is preferred in the above case.

README.md
```diff
@@ -484,6 +484,54 @@ client.Do(ctx, client.B().FtSearch().Index("idx").Query("@f:v").Build()).AsFtSea
...
+	return err
+}
+
+users := make([]*User, 0)
+if err := rueidis.DecodeSliceOfJSON(client.Do(ctx, client.B().Mget().Key("user1").Build()), &users); err != nil {
+	return err
+}
+// -> Error: invalid character 'u' looking for beginning of value
+// in this case, use client.Do(ctx, client.B().Mget().Key("user1").Build()).AsStrSlice()
+```
```
rueian, line 533: Hi @sgasho,

I understand that you want to point out an incorrect usage, but I am afraid that the warning is not obvious enough.

Users would have to look at this example very carefully, including the final comments, to know that this is a counterexample. Could you try to add more warnings around the code? For example, adding something like `!!!!!! DO NOT DO THIS !!!!!!`

rueian (approved):
Thanks @sgasho!

Issue #526, opened 2024-04-07
Feature: Add the Unit field to rueidiscompat BitCount


rueian:
go-redis recently added the BitCount.Unit field to its Cmdable implementation. We had better to add it to rueidiscompat too.

Ref: https://github.com/redis/go-redis/issues/2956

rueian:
Closed by https://github.com/redis/rueidis/pull/538

PR #538, opened 2024-04-26
feat: add unit field to bitcount

SoulPancake:
fixes 
- #526


SoulPancake:
@rueian 
Do we need to also do it for 
`func (c CacheCompat) BitCount(ctx context.Context, key string, bitCount *BitCount) *IntCmd`
for this to work?

rueian:
> CacheCompat) BitCount(ctx context.Context, ke

Yes, we should support the Unit there too.

rueidiscompat/adapter.go
```diff
@@ -1093,14 +1093,36 @@ func (c *Compat) SetBit(ctx context.Context, key string, offset int64, value int
...
+
 func (c *Compat) BitCount(ctx context.Context, key string, bitCount *BitCount) *IntCmd {
-	var resp rueidis.RedisResult
+	ret := &IntCmd{}
+
 	if bitCount == nil {
-		resp = c.client.Do(ctx, c.client.B().Bitcount().Key(key).Build())
-	} else {
-		resp = c.client.Do(ctx, c.client.B().Bitcount().Key(key).Start(bitCount.Start).End(bitCount.End).Build())
+		resp := c.client.Do(ctx, c.client.B().Bitcount().Key(key).Bit().Build()) // Default to BIT
```
rueian, line 1105: Hi @SoulPancake,

We can't use .Bit() as default because it is only supported on Redis >= 7. We should keep the original behavior.

rueidiscompat/adapter.go
```diff
@@ -1093,14 +1093,36 @@ func (c *Compat) SetBit(ctx context.Context, key string, offset int64, value int
...
-	return newIntCmd(resp)
+
+	builder := c.client.B().Bitcount().Key(key).Start(bitCount.Start).End(bitCount.End)
+
+	// Default to BIT unless specified otherwise
+	var resp rueidis.RedisResult
+	switch bitCount.Unit {
+	case BitCountIndexByte:
+		resp = c.client.Do(ctx, builder.Byte().Build()) // Handling for BYTE
+	case BitCountIndexBit, "": // Default to BIT if not specified
```
rueian, line 1117: Same here. we can't use .Bit() as default in the case of "".

rueian:
Hi @SoulPancake, would you like to continue on this?

SoulPancake:
yes give me some time I will address it tonight @rueian

SoulPancake:
@rueian I was thinking whether I should continue here or on valkey-go
You can suggest

rueian:
Hi @SoulPancake,

Here is ok. I will sync commits between them.

SoulPancake:
Will that make me a contributor there too? As I would like to be so @rueian

rueian:
> Will that make me a contributor there too? As I would like to be so @rueian

Yes of course

SoulPancake:
I have some doubts
The command slice looks alright to me
image

But still I am getting an `ERR syntax error` for this
However, it works fine for BIT

@rueian

SoulPancake:
Further, In the container that I am using, I am able to get the response too
image

rueidiscompat/command.go
```diff
@@ -1959,6 +1959,7 @@ type SetArgs struct {
 
 type BitCount struct {
 	Start, End int64
+	Unit       *string // Stores BIT or BYTE
```
rueian, line 1962: Hi @SoulPancake,

Could we keep using `string` instead of `*string`? We'd like to make the interface as close to go-redis'.

rueian:
> I have some doubts The command slice looks alright to me image
> 
> But still I am getting an `ERR syntax error` for this However, it works fine for BIT
> 
> @rueian

Do you mean here? The RESP2 test suite is tested against Redis 5. 
image

You need to wrap your new tests in a if block:

```go
		if resp3 {
			bitCount = adapter.BitCount(ctx, "key", &BitCount{
				Start: 1,
				End:   1,
				Unit:  "BYTE",
			})
			Expect(bitCount.Err()).NotTo(HaveOccurred())
			Expect(bitCount.Val()).To(Equal(int64(6)))	
		}
```

SoulPancake:
Please review @rueian :)

rueian:
LGTM. Thanks @SoulPancake!

Issue #513, opened 2024-03-23
Fix code scanning alert - Incorrect conversion between integer types


rueian:
Tracking issue for:
- [ ] https://github.com/redis/rueidis/security/code-scanning/5

moonorange:
@rueian I'm running into a 404 error when trying to access the link. Would you mind filling me in with more details about the issue?

rueian:
Hi @moonorange, here is the content of the link

<img width="932" alt="image" src="https://github.com/redis/rueidis/assets/2727535/7fd1b1d5-d5a8-4c3d-9d31-f59d79306234">

rueian:
Hi @proost,

CodeQL found the `uint(uint64)` conversion issue in 3 lines of rueidisprob.

<img width="922" alt="image" src="https://github.com/redis/rueidis/assets/2727535/140e5057-7775-4989-a72e-66cfd5f7a008">

Could we just return uint64 without conversion?

proost:
@rueian changing to uint64 is easy, but it is breaking change. 
If we care backward-compatibility, we should deprecate old methods and adding new methods. If not, change to uint64.

rueian:
Yes, it is a breaking change, but I believe it is acceptable since most people don’t explicitly declare its return value as uint. If they don’t do that they will not be break. It they do that they can also easily fix the break.

PR #541, opened 2024-05-01
fix: incorrect conversion

proost:
fix: https://github.com/redis/rueidis/issues/513

returning uint64 to avoid incorrect conversion


rueian:
Thanks @proost!

Issue #366, opened 2023-09-09
`panic`  when we try to get the redis result as some specific datatype


smit245:
I am looking into this library to use it in one of my projects and i found that when we do `client.Do` operation and it returns `RedisResult` and when we use `redisResult.AsBool()` this return bool value of the result or error but when i looked into the method i found that when result is not actually bool the program will panic as i found at [message.go#615](https://github.com/redis/rueidis/blob/c2ec5fb10d4817e4f1d96a9491760e67b49a7efe/message.go#L615).

I found this and i found it is a bad practice.
The method is already sending the error so we can easily return error why panic?
Is there any reason?

I also looked at the [Command Response Cheatsheet](https://github.com/redis/rueidis#command-response-cheatsheet) but still, mistakes happen and for that program should not panic instead error should be returned and it will be handled by the library user and then he will find out that response of redis is not in the data type they wished for.

Please  answer if it's a valid question

Otherwise i found this package amazing and loved the command builder you guys have provided also looking into other features

rueian:
Hi @smit245,

As you mentioned, using functions like `AsBool()` to parse a mismatched response data type is a programming mistake. That is exactly the reason why we make it panic instead of returning an error.

creker:
I was also burned by this multiple times. Often, this happens in DoMulti calls where the order of the results is important. Testing doesn't help much because mocked rueidis is controlled by the programmer. It might return wrong result types that happen to match the expectations of the code you're testing.

I think this should be changed to returning an error. It is a programming mistake, but it happens at runtime and causes outages in the worst case if panic is not recovered. At least with errors it allows code to continue to run and defers the problem to metrics/logs to notify the user about buggy behaviour. 

And it's simply not idiomatic. Library code should panic only in very rare cases. It makes sense to panic in constructors to terminate the application before it even launches. Standard library even does this and it's ok usually. But panicking later is bad practice in almost all cases.

kevinxmorales:
I've also run into issues with panics during development, where I wasn't sure what type method to use on RedisResult values. I do think the panic's shouldn't be too bad of a problem once you figure it out in development and I personally wouldn't rely on this proposal.
I was thinking making this behavior configurable could be done by adding a command before the final type method
Something like this
Add a `noPanic` boolean that determines if it should not panic (false by default, keep current behavior)
```go
type RedisResult struct {
	err error
	val RedisMessage

	noPanic bool
}
```
add a NP (No Panic) method to RedisResult that sets the noPanic value to true
[...]

PR #555, opened 2024-06-03
feat: Panic -> Error migration

SoulPancake:
This is to address #366 

This is in draft state with some idea I understood from the current proposal thread.
We can discuss and decide on what specifically we want to do here, and how much of it has to stay backwards compatible and in what all way and whether we want to introduce some breaking major changes too.

@rueian Let's continue the discussion further on this

### DIFF
--- a/message.go
+++ b/message.go
@@ -571,8 +571,7 @@ func (m *RedisMessage) ToString() (val string, err error) {
 	}
 	if m.IsInt64() || m.values != nil {
 		typ := m.typ
-		return "",
-			fmt.Errorf("redis message type %s is not a string", typeNames[typ])
+		return "", fmt.Errorf("redis message type %s is not a string", typeNames[typ])
 	}
 	return m.string, m.Error()
 }
@@ -646,8 +645,7 @@ func (m *RedisMessage) AsBool() (val bool, err error) {
 		return
 	default:
 		typ := m.typ
-		return false,
-			fmt.Errorf("redis message type %s is not a int, string or bool", typeNames[typ])
+		return false, fmt.Errorf("redis message type %s is not a int, string or bool", typeNames[typ])
 	}
 }
 
@@ -936,7 +934,7 @@ func (m *RedisMessage) AsMap() (map[string]RedisMessage, error) {
 		return nil, err
 	}
 	if (m.IsMap() || m.IsArray()) && len(m.values)%2 == 0 {
-		return toMap(m.values), nil
+		return toMap(m.values)
 	}
 	typ := m.typ
 	return nil, fmt.Errorf("redis message type %s is not a map/array/set or its length is not even", typeNames[typ])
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -18,6 +18,29 @@ const messageStructSize = int(unsafe.Sizeof(RedisMessage{}))
...
+var defaultErrorXRangeEntrySliceMap = make(map[string][]XRangeEntry)
+var defaultErrorZScoreSliceMap = make(map[string][]ZScore)
+var defaultErrorZScore = ZScore{}
+var defaultErrorRedisMessageMap = make(map[string]RedisMessage)
+var defaultErrorScanEntry = ScanEntry{}
+var defaultErrorStrMap = make(map[string]string)
+var defaultErrorStrMapSlice = make([]map[string]string, 0)
+var defaultErrorIntMap = make(map[string]int64)
+var defaultErrorKeyValues = KeyValues{}
+var defaultErrorKeyZScores = KeyZScores{}
```
SoulPancake, line 39: On retrospect this might not be a great idea at all

Maybe I misunderstood the proposal a little

rueian: We don't need to pre-define these default error values. We should just return a zero value of the corresponding type as well as the error. For example:

`return defaultErrorRedisMessageSlice, fmt.Errorf("redis message type %s is not a array", typeNames[typ])`

should be just

`return nil, fmt.Errorf("redis message type %s is not a array", typeNames[typ])`

SoulPancake: @rueian Zero values for the non-nullable ones? 

It will not compile in those cases when we send back a nil for non-nullable one I guess
Okay let me do try this refactor

SoulPancake: Also, is this the only file where we require changes for this? @rueian 

Any other changes we require other than the tests

rueian: Yes, the `message.go` is the only file that needs to be changed other than the tests.

message.go
```diff
@@ -571,7 +571,8 @@ func (m *RedisMessage) ToString() (val string, err error) {
 	}
 	if m.IsInt64() || m.values != nil {
 		typ := m.typ
-		panic(fmt.Sprintf("redis message type %s is not a string", typeNames[typ]))
+		return "",
+			fmt.Errorf("redis message type %s is not a string", typeNames[typ])
```
rueian, line 575: Please keep them on the same line.

message.go
```diff
@@ -645,7 +646,8 @@ func (m *RedisMessage) AsBool() (val bool, err error) {
 		return
 	default:
 		typ := m.typ
-		panic(fmt.Sprintf("redis message type %s is not a int, string or bool", typeNames[typ]))
+		return false,
+			fmt.Errorf("redis message type %s is not a int, string or bool", typeNames[typ])
```
rueian, line 650: Please keep them on the same line.

message.go
```diff
@@ -1319,7 +1321,9 @@ func toMap(values []RedisMessage) map[string]RedisMessage {
 			continue
 		}
 		typ := values[i].typ
-		panic(fmt.Sprintf("redis message type %s as map key is not supported", typeNames[typ]))
+		return map[string]RedisMessage{
+			"error": {typ: typeSimpleErr, string: fmt.Sprintf("redis message type %s as map key is not supported", typeNames[typ])},
+		}
```
rueian, line 1326: We should return an error as the second return value here, instead of embedding it into the map.

rueian (approved):
Cool! LGTM! Thank you @SoulPancake!

PR #562, opened 2024-06-12
Update build.yml and aside.go

JayChavan-23:
Updated build.yml with new go-version variable with values including 1.21 and 1.22 to the matrix
and
Added an option to ClientOption to use Lua scripts for acquiring locks, enhancing compatibility with older Redis versions that do not support the SET NX GET command.


JayChavan-23:
Do I have to do something in order to get this test cases passed or is it normal?

PR #563, opened 2024-06-12
Updated build.yml

JayChavan-23:
Added new go-version variable with values including 1.21 and 1.22 to the matrix.


rueian:
Thank you @JayChavan-23!

Issue #559, opened 2024-06-10
Remove the`oklog/ulid` dependency from the `rueidisaside` submodule


rueian:
The `github.com/oklog/ulid/v2` is the only dependency used by the `rueidisaside` submodule and it uses the dependency only to generate random strings:

https://github.com/redis/rueidis/blob/0124d4a3cc7c3a8d5f61aa6033481e160fd37f10/rueidisaside/aside.go#L116

It would be great if we could replace the `github.com/oklog/ulid/v2` with a self-made random string generator function, and then we can merge the `rueidisaside` submodule back to the root `rueidis`.

psotou:
I always use the implementation described at 7 in the accepted answer of this StackOverflow question https://stackoverflow.com/questions/22892120

psotou:
> Hi @psotou, thank you for sharing! The masking method is really cool.

It is indeed! If you want I can find the time to try to work out something using the masking method (or masking improved).

rueian:
Sure. Also, we just need a random string function that produces a fixed length of strings, such as 24 or 32. So it should be much easier to find an optimized version. For example, how about this one?

```go
import (
	"encoding/binary"
	"encoding/hex"
	"math/rand/v2"
	"unsafe"
)

func random() string {
	b := make([]byte, 32)
	binary.LittleEndian.PutUint64(b[16:], rand.Uint64())
	binary.LittleEndian.PutUint64(b[24:], rand.Uint64())
	hex.Encode(b, b[16:])
	return unsafe.String(unsafe.SliceData(b), len(b))
}
```

neelpatel04466:
Here's the updated code implementing the custom random string generator function as discussed:

```go
package rueidisaside

import (
    "encoding/binary"
    "encoding/hex"
    "math/rand"
    "unsafe"
)

const PlaceholderPrefix = "your_prefix_"  // Replace with the actual prefix if needed

func generateRandomString() string {
    b := make([]byte, 32)
    binary.LittleEndian.PutUint64(b[16:], rand.Uint64())
    binary.LittleEndian.PutUint64(b[24:], rand.Uint64())
    hex.Encode(b, b[16:])
    return unsafe.String(unsafe.SliceData(b), len(b))
}

func someFunction() {
    id := PlaceholderPrefix + generateRandomString()
    // Rest of your function logic
}

psotou:
if it's all the same to you @rueian, I'd advocate for a 24-byte rand string. Based on your proposal,
```go
func random() string {
[...]

PR #573, opened 2024-06-17
enhancement: remove oklog/ulid dependency for rand string generation

psotou:
Custom rand string generator to replace `oklog/ulid` dependency. This PR solves issue #559


rueian:
Hi @psotou, could you also help us remove the `go.mod` and `go.sum` from the `rueidisaside` as well?

psotou:
> Hi @psotou, could you also help us remove the `go.mod` and `go.sum` from the `rueidisaside` as well?

Hi there. Shall I move everything to the root of the project?

rueian:
No need. Removing those `go.*` files is enough just like the `rueidislock`.

Issue #567, opened 2024-06-14
Fallback MGetCache to MGet when DisableCache=true


jo-me:
Hi,

I'm playing around with rueidis as drop-in replacement for redis-go, but I'm seeing some strange performance in a (local) comparison.
On my dev machine I have redis running in docker and a simple API service that checks redis based on the IDs given in the API requests. 
All tests were done with Rueidis v1.0.38 and go-redis v9.5.3 using go 1.22.3.

For my simple test I'm setting 5 values in redis with~2KB each and hammer my API with k6 to see how many requests per second I get when I request all 5 elements over and over.

Without any caching involved, I have around 50 rps (have to recreate the 5 elements).
**With go-redis I have around 25000 rps**

Now, when switching to ruedis I tried out 3 different configurations:
**MGetCache with client option DisableCache:true:     **~500 rps****
MGetCache with client option DisableCache:false:   ~50000rps
DoMultiStream:   ~30000 rps

So, with client-side caching the rps are considerably higher as expected (although the difference is not that much probably due to redis being on the same host), but with _disabled_ client-side cache, there seems to be something strange going on.
Aside from the DisableCache option the code is exactly the same.

To my understanding Mget and Mgetcache (without client-side caching) should behave identically and I would have expected a similar result as with go-redis.

Any idea what could be the issue?

Thanks!

Jochen Mehlhorn <jochen.mehlhorn@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH

[Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)

rueian:
Hi @jo-me,

Thank you for sharing your test results. Unlike `MGet` helper, `MGetCache` with `DisableCache=true` currently sends each key separately to Redis.

Could you also test `MGet`? I think it would be similar to go-redis. If that is the case, we can fallback `MGetCache` to `MGet` when `DisableCache=true` in the future.

jo-me:
Hmm, 
I repeated the test this morning with the same setup as described above but another two variants for the no caching code path:

`client.Do(ctx, c.client.B().Mget().Key(keys...).Build()).AsStrSlice()`
and
`rueidis.MGet(client, ctx, keys)`

but both were on par with the MgetCache variant with DisableCache=true
I also tried it without the DisableCache client option in case it somehow influences the Do/Mget commands, but it did non make any difference.

My client initialization is like this btw:

```
[...]

PR #574, opened 2024-06-18
feat: fallback MGetCache to MGet

SoulPancake:
Solves for #567

### DIFF
--- a/helper.go
+++ b/helper.go
@@ -60,14 +60,11 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 func isCacheDisabled(client Client) bool {
 	switch c := client.(type) {
 	case *singleClient:
-		if c.DisableCache {
-			return true
-		}
-		return false
+		return c.DisableCache
 	case *sentinelClient:
-		return (c.mOpt != nil && c.mOpt.DisableCache) || (c.sOpt != nil && c.sOpt.DisableCache)
+		return (c.mOpt != nil && c.mOpt.DisableCache)
 	case *clusterClient:
-		return (c.opt != nil && c.opt.DisableCache) || (c.rOpt != nil && c.rOpt.DisableCache)
+		return (c.opt != nil && c.opt.DisableCache)
 	}
 	return false
 }
--- a/helper.go
+++ b/helper.go
@@ -60,15 +60,14 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 func isCacheDisabled(client Client) bool {
 	switch c := client.(type) {
 	case *singleClient:
-		// what do we do here ? @rueian
+		if c.DisableCache {
+			return true
+		}
 		return false
 	case *sentinelClient:
-		if c.mOpt != nil && c.mOpt.DisableCache {
-			return true
-		}
-		if c.sOpt != nil && c.sOpt.DisableCache {
-			return true
-		}
+		return (c.mOpt != nil && c.mOpt.DisableCache) || (c.sOpt != nil && c.sOpt.DisableCache)
+	case *clusterClient:
+		return (c.opt != nil && c.opt.DisableCache) || (c.rOpt != nil && c.rOpt.DisableCache)
 	}
 	return false
 }
--- a/helper.go
+++ b/helper.go
@@ -62,9 +62,9 @@ func isCacheDisabled(client Client) bool {
 	case *singleClient:
 		return c.DisableCache
 	case *sentinelClient:
-		return (c.mOpt != nil && c.mOpt.DisableCache)
+		return c.mOpt != nil && c.mOpt.DisableCache
 	case *clusterClient:
-		return (c.opt != nil && c.opt.DisableCache)
+		return c.opt != nil && c.opt.DisableCache
 	}
 	return false
 }
--- a/helper.go
+++ b/helper.go
@@ -46,12 +46,31 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 	if len(keys) == 0 {
 		return make(map[string]RedisMessage), nil
 	}
+	if isCacheDisabled(client) {
+		return MGet(client, ctx, keys)
+	}
 	cmds := mgetcachecmdsp.Get(len(keys), len(keys))
 	defer mgetcachecmdsp.Put(cmds)
 	for i := range cmds.s {
 		cmds.s[i] = CT(client.B().Get().Key(keys[i]).Cache(), ttl)
 	}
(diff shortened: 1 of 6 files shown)


helper.go
```diff
@@ -54,6 +57,22 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 	return doMultiCache(client, ctx, cmds.s, keys)
 }
 
+func isCacheDisabled(client Client) bool {
+	switch c := client.(type) {
+	case *singleClient:
+		// what do we do here ? @rueian
```
SoulPancake, line 63: Is this the right approach to this? @rueian 
Can you suggest?

rueian: We may copy the `DisableCache` to a new bool field in the `singleClient` and then check the field here.

helper.go
```diff
@@ -54,6 +57,22 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
...
+		return false
+	case *sentinelClient:
+		if c.mOpt != nil && c.mOpt.DisableCache {
+			return true
+		}
+		if c.sOpt != nil && c.sOpt.DisableCache {
+			return true
+		}
+	}
```
rueian, line 72: We should also do something similar for the `clusterClient`.

sentinel.go
```diff
@@ -177,7 +177,11 @@ func (c *sentinelClient) Dedicate() (DedicatedClient, func()) {
 
 func (c *sentinelClient) Nodes() map[string]Client {
 	conn := c.mConn.Load().(conn)
-	return map[string]Client{conn.Addr(): newSingleClientWithConn(conn, c.cmd, c.retry)}
+	disableCache := false
+	if (c.mOpt != nil && c.mOpt.DisableCache) || (c.sOpt != nil && c.sOpt.DisableCache) {
```
rueian, line 181: We don't need to consider `c.sOpt` here. That is basically a copy of `c.mOpt`.

helper.go
```diff
@@ -54,6 +57,21 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 	return doMultiCache(client, ctx, cmds.s, keys)
 }
 
+func isCacheDisabled(client Client) bool {
+	switch c := client.(type) {
+	case *singleClient:
+		if c.DisableCache {
+			return true
+		}
+		return false
```
rueian, line 66:
```suggestion
		return c.DisableCache
```

helper.go
```diff
@@ -54,6 +57,21 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
...
+	case *sentinelClient:
+		return (c.mOpt != nil && c.mOpt.DisableCache) || (c.sOpt != nil && c.sOpt.DisableCache)
```
rueian, line 68:
```suggestion
		return (c.mOpt != nil && c.mOpt.DisableCache)
```

helper.go
```diff
@@ -54,6 +57,21 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
...
+	case *clusterClient:
+		return (c.opt != nil && c.opt.DisableCache) || (c.rOpt != nil && c.rOpt.DisableCache)
```
rueian, line 70:
```suggestion
		return (c.opt != nil && c.opt.DisableCache)
```

helper.go
```diff
@@ -46,6 +46,9 @@ func MGetCache(client Client, ctx context.Context, ttl time.Duration, keys []str
 	if len(keys) == 0 {
 		return make(map[string]RedisMessage), nil
 	}
+	if isCacheDisabled(client) {
+		return MGet(client, ctx, keys)
```
rueian, line 50: We need tests on this.

SoulPancake: Will update the tests and cover these conditions

helper_test.go
```diff
@@ -108,6 +125,35 @@ func TestMGetCache(t *testing.T) {
 			}
 		})
 	})
+	// TODO: @SoulPancake I have some doubts about this @rueian
```
SoulPancake, line 128: I have some doubts here, Can you suggest how we can cover the sentinelClient scenario? @rueian

rueian: Here is an example:  https://github.com/redis/rueidis/blob/63951b5b27a90738fe251a48305f65954df3fec9/sentinel_test.go#L664

helper.go
SoulPancake: I have covered that case, can you please take a look again? @rueian

helper_test.go
```diff
@@ -18,6 +18,23 @@ func TestMGetCache(t *testing.T) {
 		if err != nil {
 			t.Fatalf("unexpected err %v", err)
 		}
+		disabledCacheClient, err := newSingleClient(&ClientOption{InitAddress: []string{""}, DisableCache: true}, m, func(dst string, opt *ClientOption) conn {
```
rueian, line 21: Hi @SoulPancake, we still need a test for the `newClusterClient`.

helper_test.go
```diff
@@ -55,13 +72,49 @@ func TestMGetCache(t *testing.T) {
...
+				return &redisresults{s: []RedisResult{newErrResult(context.Canceled), newErrResult(context.Canceled)}}
+			},
 		}
 		client, err := newClusterClient(&ClientOption{InitAddress: []string{":0"}}, func(dst string, opt *ClientOption) conn {
 			return m
 		})
 		if err != nil {
 			t.Fatalf("unexpected err %v", err)
 		}
+		disabledCacheClient, err := newClusterClient(&ClientOption{InitAddress: []string{":0"}, DisableCache: true}, func(dst string, opt *ClientOption) conn {
```
SoulPancake, line 85: I have some doubts about this, and it is timing out, Can you take a look? @rueian

rueian: Hi @SoulPancake, when the `DisableCache` is set, you should mock the m.DoMultiFn instead of the m.DoMultiCacheFn.

SoulPancake: Done, Can you please review @rueian

helper_test.go
```diff
@@ -55,13 +72,49 @@ func TestMGetCache(t *testing.T) {
 			DoFn: func(cmd Completed) RedisResult {
 				return slotsResp
 			},
+			DoMultiFn: func(cmd ...Completed) *redisresults {
+				return &redisresults{s: []RedisResult{newErrResult(context.Canceled), newErrResult(context.Canceled)}}
```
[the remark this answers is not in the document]
SoulPancake: removed

rueian (approved):

rueian:
Thank you @SoulPancake!

Issue #579, opened 2024-06-22
Send `DISCARD` automatically when a dedicated connection is recycled


rueian:
Following the discussion https://github.com/redis/rueidis/discussions/571, we had better send a `DISCARD` command when a dedicated connection is recycled by adding the command here:
https://github.com/redis/rueidis/blob/5fe0175fe338cca5685968b73963e0eb159affb3/pipe.go#L751-L759

Doing this will prevent unintentional bugs and other accidents while improving developer experiences.

PR #580, opened 2024-06-24
feat: send DISCARD command when conn recycled

SoulPancake:
Fixes - #579

### DIFF
--- a/internal/cmds/cmds.go
+++ b/internal/cmds/cmds.go
@@ -76,6 +76,12 @@ var (
 		cs: newCommandSlice([]string{"UNSUBSCRIBE", "+sentinel", "+slave", "-sdown", "+sdown", "+switch-master", "+reboot"}),
 		cf: noRetTag,
 	}
+
+	// DiscardCmd is predefined DISCARD
+	DiscardCmd = Completed{
+		cs: newCommandSlice([]string{"DISCARD"}),
+		cf: noRetTag,
+	}
 )
 
 // ToBlock marks the command with blockTag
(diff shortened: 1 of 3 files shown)


SoulPancake:
@rueian Can you please take a look?

internal/cmds/cmds.go
```diff
@@ -76,6 +76,12 @@ var (
 		cs: newCommandSlice([]string{"UNSUBSCRIBE", "+sentinel", "+slave", "-sdown", "+sdown", "+switch-master", "+reboot"}),
 		cf: noRetTag,
 	}
+
+	// DiscardCmd is predefined DISCARD
+	DiscardCmd = Completed{
+		cs: newCommandSlice([]string{"DISCARD"}),
+		cf: noRetTag,
```
rueian, line 83:
```suggestion
```

We can’t set the noRetTag on it. The tag will trigger a special handling for PUBSUB commands.

rueian (approved):

pipe_test.go
```diff
@@ -4129,6 +4129,11 @@ func TestPipe_CleanSubscriptions_6(t *testing.T) {
 			{typ: '+', string: "punsubscribe"},
 			{typ: '_'},
 			{typ: ':', integer: 2},
+		}},
+		RedisMessage{typ: '>', values: []RedisMessage{
```
rueian, line 4133: @SoulPancake, I found this reply is incorrect. DISCARD should reply a OK message instead of a push message.

rueian (changes requested):

pipe_test.go
SoulPancake: My bad, Addressed @rueian

rueian (approved):

Issue #487, opened 2024-03-03
Need to improve the test coverage on rueidiscompat


rueian:
Currently, we have poor test coverage on rueidiscompat:

<img width="1592" alt="image" src="https://github.com/redis/rueidis/assets/2727535/06e48a5e-cb1a-4051-aaf6-1c773b04c580">

Details of which lines are not covered can be found here: https://app.codecov.io/gh/redis/rueidis/tree/main/rueidiscompat

We hope that we can improve the coverage to be close to 100%.

PR #583, opened 2024-06-27
feat: 5% increase in command.go coverage

SoulPancake:
part of #487

### DIFF
--- a/message.go
+++ b/message.go
@@ -1386,3 +1386,8 @@ func (m *prettyRedisMessage) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal(obj)
 }
+
+// NewMockResult is a mock method to send back RedisResult with err for test coverage
+func NewMockResult(val RedisMessage, err error) RedisResult {
+	return RedisResult{val: val, err: err}
+}
--- a/rueidiscompat/command_test.go
+++ b/rueidiscompat/command_test.go
@@ -29,6 +29,8 @@ package rueidiscompat
 import (
 	"errors"
 	"fmt"
+	"github.com/redis/rueidis"
+	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


message.go
```diff
@@ -1386,3 +1386,8 @@ func (m *prettyRedisMessage) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal(obj)
 }
+
+// NewMockResult is a mock method to send back RedisResult with err for test coverage
+func NewMockResult(val RedisMessage, err error) RedisResult {
```
SoulPancake, line 1391: I have mixed thoughts about this
Because even though this will provide us a way to cover roughly 20 more lines of code in ruediscompat command, we should not probably expose this as a part of rueidis package

But the only alternative to cover most of the other parts inside command would be to introduce interfaces instead of passing concrete structs and a large refactor. Otherwise, it seems difficult to achieve near-100% coverage.

What do you think? @rueian

rueidiscompat/command_test.go
```diff
@@ -29,6 +29,8 @@ package rueidiscompat
 import (
 	"errors"
 	"fmt"
+	"github.com/redis/rueidis"
```
rueian, line 32: The import should be moved to below.

rueidiscompat/command_test.go
```diff
@@ -29,6 +29,8 @@ package rueidiscompat
...
+	"github.com/stretchr/testify/assert"
```
rueian, line 33: Could we avoid introducing a new test dependency?

message.go
rueian: > we should not probably expose this as a part of rueidis package

Agreed. Actually, we don't need this function but we can use the same hack of the `mock` package instead:
https://github.com/redis/rueidis/blob/45cdf7eb946fed65e41ec78a654860d44b79a143/mock/result.go#L22-L25

SoulPancake: Oh wow
This is a great use of unsafePtrs

rueidiscompat/go.mod
```diff
@@ -8,13 +8,15 @@ require (
 	github.com/onsi/ginkgo/v2 v2.15.0
 	github.com/onsi/gomega v1.31.1
 	github.com/redis/rueidis v1.0.40
+	github.com/redis/rueidis/mock v1.0.40
```
rueian, line 11: Please also add a local replace directive for this.

rueidiscompat/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/redis/rueidis/rueidiscompat
 go 1.20
 
 replace github.com/redis/rueidis => ../
+replace github.com/redis/rueidis/mock => ../mock
```
rueian, line 6: We should run `go mod tidy` again after adding the replace directive.

rueian (approved):

rueian:
LGTM. Thanks @SoulPancake!

Issue #489, opened 2024-03-04
feature request: adding an option to disable dial on creating client


mehdipourfar:
When we initialize a redis client with go-redis, it does not return any error. I think this is a very good feature. Imagine I use redis in my web service only as a cache. If it goes down, it is still ok and I can use the main repository for fetching my data. If it goes up during the service runtime, I will use that cache again. But this package returns nil and error if it cannot connect to redis on initialization and I don't want to interrupt my service start up because my redis server is out of reach.

rueian:
Hi @mehdipourfar,

If redis servers go down after the initialization, rueidis will try reconnection automatically and can work as you want.

rueidis has a good reason to return an error at initialization, allowing users to verify whether their config for establishing a Redis connection is correct or not. I generally recommend users to fail fast and fix fast.

Indeed, this approach will interrupt service startup if their config is correct but Redis server are not ready. However, there is another purpose of initialization which I think is the true blocker to your request: choosing the correct client implementation.

Currently, rueidis will connect to the server to check if it is a Redis cluster. If yes, then the initialization returns a cluster client implementation, otherwise returns a standalone client implementation.

If we let this behavior be configurable, then we can return a concrete client implementation instead of a nil even when the initialization has failed.

mehdipourfar:
Thanks for your reply. What happens if we try to not `Dial` in `newSentinelClient` and `newSingleClient`?

rueian:
You will have no chance to know whether your config is wrong or not before sending your first redis request.

rueian:
I think we should keep the Dail() even if we make the auto-detection configurable. Users can just ignore the error returned by the initialization if they don't care about it.

mehdipourfar:
That would solve my problem if we don't return nil as client.

erdemtuna:
Hey @rueian, can we clarify the description and expected change of this task for a potential implementation?

rueian:
Hi @erdemtuna,

The goal is to make the `rueidis.NewClient` return a concrete client implementation even if there is an error occurs. However, which concrete client implementation to use is currently queried from an online redis instance. This is the first blocker that we are unable to return a concrete implementation. We need a method to let users specify which implementation to use explicitly.

Since we already have the option `ForceSingleClient`, a potential implementation starts from adding a new one `ForceClusterClient`:
1. When `ForceClusterClient` is true, let `rueidis.NewClient` return a `clusterClient` even if there is an error occurs.
[...]

PR #584, opened 2024-07-01
feat: Option to disable dial on creating client

SoulPancake:
Approach to #489

### DIFF
--- a/rueidis.go
+++ b/rueidis.go
@@ -367,8 +367,6 @@ func NewClient(option ClientOption) (client Client, err error) {
 		option.PipelineMultiplex = singleClientMultiplex(option.PipelineMultiplex)
 		singleClt, err := newSingleClient(&option, nil, makeConn)
 		if err != nil {
-			// Handle the error gracefully
-			// TODO: @SoulPancake how to ensure the client can work after Redis is back
 			return &singleClient{
 				cmd:          cmds.NewBuilder(cmds.NoSlot),
 				conn:         nil,
--- a/rueidis.go
+++ b/rueidis.go
@@ -4,9 +4,11 @@ package rueidis
 //go:generate go run hack/cmds/gen.go internal/cmds hack/cmds/*.json
 
 import (
+	"container/list"
 	"context"
 	"crypto/tls"
 	"errors"
+	"github.com/redis/rueidis/internal/cmds"
 	"math"
 	"net"
 	"runtime"
@@ -166,6 +168,9 @@ type ClientOption struct {
 	//  ForceSingleClient force the usage of a single client connection, without letting the lib guessing
 	//  if redis instance is a cluster or a single redis instance.
 	ForceSingleClient bool
+	//  ForceClusterClient forces the usage of a cluster client connection, without letting the lib guess
+	//  if the Redis instance is a cluster or a single Redis instance.
+	ForceClusterClient bool
 
 	// ReplicaOnly indicates that this client will only try to connect to readonly replicas of redis setup.
 	ReplicaOnly bool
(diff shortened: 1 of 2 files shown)


SoulPancake:
Hi @rueian I decided to begin on this
However, I have a few doubts about the requirement and have some questions.
1. https://github.com/redis/rueidis/issues/489#issuecomment-1988702413
In this how do we go about testing #4 in a failsafe way.
2. How does the auto-retry work? Do we mean simply returning the structs of these clients without the dial failure being involved like I have done in the current commit? If yes, then how do we update it when the retry is successful?

Thanks in advanced. Let me know which direction I can take for this

rueidis.go
```diff
@@ -342,14 +347,49 @@ func NewClient(option ClientOption) (client Client, err error) {
...
+			// TODO: @SoulPancake how to ensure the client can work after Redis is back
+			return &sentinelClient{
+				cmd:       cmds.NewBuilder(cmds.NoSlot),
+				mOpt:      &option,
+				sOpt:      newSentinelOpt(&option),
+				connFn:    makeConn,
+				sentinels: list.New(),
+				retry:     !option.DisableRetry,
+				replica:   option.ReplicaOnly,
+			}, nil
```
rueian, line 362: The `err` should still be returned.

rueidis.go
```diff
@@ -342,14 +347,49 @@ func NewClient(option ClientOption) (client Client, err error) {
...
+		singleClt, err := newSingleClient(&option, nil, makeConn)
+		if err != nil {
+			// Handle the error gracefully
+			// TODO: @SoulPancake how to ensure the client can work after Redis is back
+			return &singleClient{
+				cmd:          cmds.NewBuilder(cmds.NoSlot),
+				conn:         nil,
+				retry:        !option.DisableRetry,
+				DisableCache: option.DisableCache,
+			}, nil
```
rueian, line 377: The `err` should still be returned.

rueidis.go
```diff
@@ -342,14 +347,49 @@ func NewClient(option ClientOption) (client Client, err error) {
...
+		if client == (*clusterClient)(nil) && option.ForceClusterClient {
+			// Return a clusterClient instance if ForceClusterClient is enabled
+			return &clusterClient{
+				cmd:    cmds.NewBuilder(cmds.InitSlot),
+				connFn: makeConn,
+				opt:    &option,
+				conns:  make(map[string]connrole),
+				retry:  !option.DisableRetry,
+				aws:    len(option.InitAddress) == 1 && strings.Contains(option.InitAddress[0], "amazonaws.com"),
+			}, nil
```
rueian, line 391: The `err` should still be returned.

rueidis_test.go
```diff
@@ -197,38 +197,84 @@ func TestFallBackSingleClient(t *testing.T) {
...
+				mock.Expect("CLIENT", "SETINFO", "LIB-VER", LibVer).
+					ReplyError("UNKNOWN COMMAND")
+				mock.Expect("PING").ReplyString("OK")
+				mock.Close()
+				close(done)
+			}()
+
+			_, port, _ := net.SplitHostPort(ln.Addr().String())
+
+			dialFn := func(addr string, dialer *net.Dialer, tlsConfig *tls.Config) (net.Conn, error) {
```
SoulPancake, line 239: @rueian Do you think this is how I should do it?
In principle it should work but I am not too sure why it is timing out, maybe I'm missing something

rueian: The `dailFn` looks good to me but you must invoke `client.Do` to trigger the `dailFn` again after the `NewClient`. Otherwise,  the `dailFn` is only called once.

SoulPancake: I am slightly stuck here
I should client.Do to trigger the dialFn right after the NewClient call? @rueian

rueian: Yes. I believe so. There is no other background goroutine that will dial for a connection.

SoulPancake:
This has been stale because of other work, so I am closing this now, might re-open a different PR rebasing from latest master
@rueian

PR #588, opened 2024-07-21
fix: remove unnecessary aws configuration endpoint specialization

rueian:
The endpoint is just a DNS alias to any nodes. There is no need for a special treatment.


proost (approved):
Nice! Thank you!

jinwoo1225 (approved):
yay

PR #589, opened 2024-07-21
enhancement: mitigate refresh storm

proost:
what i found is when got "MOVED" to specific slot, many refreshes are possible. so we update "pslots" and "rslots" when got "MOVED".

"ASK" also triggers many refreshes but it should not change "pslots" and "rslots". because importing node may not serve some items which are migrated yet.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -444,7 +444,7 @@ func (c *clusterClient) pick(ctx context.Context, slot uint16, toReplica bool) (
 	return p, nil
 }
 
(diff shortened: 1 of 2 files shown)


rueian (approved):

rueian:
Cool! Thanks @proost! This looks good to me. I think we can merge it. However, I should remind you that this enhancement may be hard to trigger. It only updates ONE slot and only when a new node is added.

proost:
@rueian
Yes, i agree. This situation is very unlikely happen.

PR #595, opened 2024-07-30
fix: too large arguments

proost:
I got error, when `redis.call` parameters are too large.
```
user_script:15: too many results to unpack script: 73a336266d816e4bcf60d6cf47daafcad23ad693, on @user_script:15
```

`LUAI_MAXCSTACK` in luaconf.h limits stack slots. script can run very long, but it should be avoided by user not rueidisprob.


rueian:
LGTM.

rueian:
BTW, would it be better if we group, for example, 4 bits into one BITFIELD call?

PR #598, opened 2024-08-05
feat: add enable read operation

proost:
[Discussion](https://github.com/redis/rueidis/discussions/597)

As we discussed before, supporting replica when using bloom filter.

### DIFF
--- a/rueidisprob/bloomfilter.go
+++ b/rueidisprob/bloomfilter.go
@@ -1,12 +1,13 @@
 package rueidisprob
 
 import (
+	"bufio"
 	"context"
 	"errors"
+	"github.com/redis/rueidis"
 	"math"
 	"strconv"
-
-	"github.com/redis/rueidis"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


rueian:
Hi @proost,

The docker used in Github Actions seems to be updated and the `docker-compose` command has gone. Please rebase this PR to the latest main branch.

rueidisprob/bloomfilter.go
```diff
@@ -170,6 +212,32 @@ func NewBloomFilter(
...
+		if resp.Error() != nil {
+			return nil, resp.Error()
+		}
+
+		info, err := resp.ToString()
+		if err != nil {
+			return nil, err
+		}
+
+		if parseRedisMajorVersion(info) < 7 {
```
rueian, line 232: Users may run a redis cluster with mixed versions and a single INFO request is not enough for checking the version of each node. Actually, I think this check is unnecessary. Stating the version requirement in the comment of WithEnableReadOperation is enough.

rueidisprob/bloomfilter.go
```diff
@@ -1,12 +1,13 @@
 package rueidisprob
 
 import (
+	"bufio"
 	"context"
 	"errors"
+	"github.com/redis/rueidis"
```
rueian, line 7: Keep this at the last import line.

rueidisprob/bloomfilter.go
proost: > Users may run a redis cluster with mixed versions and a single INFO request is not enough for checking the version of each node. 

Oh, i missed. thank you.

[8832748](https://github.com/redis/rueidis/pull/598/commits/8832748a2a85c64cc0e3e2b102d559fd11484710)

rueian (approved):

rueian:
Merged. But I forgot that the `parseRedisMajorVersion` is still there. Could you send another PR to remove it?

Issue #587, opened 2024-07-18
Error updating the cluster architecture


ajzach:
Hello, I am having an issue with one of our clusters (AWS Elasticache) that has autoscaling configured. The cluster has a minimum of 10 and a maximum of 15. During the day, the cluster scaled up to 13 nodes, and the client correctly detected the new nodes. However, it then scaled back down to 10, and we started seeing errors where the client was trying to resolve the domain of the nodes that had been removed, causing errors. I tested adding the 11th node back, and the client detected it, but once I removed it again, the errors reappeared. It seems like the client is not updating the cluster architecture internally. We are using version 1.0.35.

rueian:
Hi @ajzach, did the errors disappear eventually?

ajzach:
> Hi @ajzach, did the errors disappear eventually?

Hello @rueian, no, we saw the errors for hours, they disappeared when I added the nodes back to the cache.

proost:
I also experienced a similar issue. When i scale down nodes in cluster, client doesn't update correctly. currently i digging it now.

rueian:
Hi @ajzach,

I have just tested AWS Elasticache by deleting shards manually, but I didn't see the DNS error you described and rueidis updated the cluster nodes successfully.
<img width="1570" alt="image" src="https://github.com/user-attachments/assets/ba68ca18-8bbe-49f6-9a89-c34e7825eebe">

Additionally, I found the AWS Elasticache responds to the `CLUSTER SHARDS` with IP addresses instead of domain names.
> the client was trying to resolve the domain of the nodes that had been removed, causing errors.

So this looks weird to me. Do you manually put the domain names of nodes into the `rueidis.ClientOption.InitAddress`? In the case of AWS Elasticache, the InitAddress should be just the configuration endpoint provided by AWS.

ajzach:
To connect to Redis, I just use the endpoint. I performed the same tests adding and removing nodes, including failover, and everything worked correctly. This error is not very common; currently, we have more than 100 applications using Rueidis, and only 2 have reported this problem. It seems that the error occurs with more prolonged use of the client. Some condition causes the client to stop updating the cluster architecture internally and to retain in memory nodes that were deleted.

rueian:
Hi @ajzach,

[...]

PR #615, opened 2024-08-25
fix: set timeout on refreshing cluster slots to avoid being stuck

rueian:
Related to https://github.com/redis/rueidis/issues/587


tituscarl:
Hi @rueian, If it's okay, can I contact you on email or linkedIn?  I have some questions for cluster support. Thanks!

Issue #620, opened 2024-08-27
Rueidis making `CLIENT SETINFO` commands that are not supported by ElastiCache Redis


rahulgkatre:
We have observed that our application is making `CLIENT SETINFO` commands to the Redis server hosted on AWS ElastiCache. However, these commands are not supported by ElastiCache Redis.

The `CLIENT SETINFO` command is used to set arbitrary information for the current connection. While this command is available in open-source Redis version 7.2 and up, it is not supported by ElastiCache Redis (which is on "7.1" or 7.0).

Steps to Reproduce:
1. Create an ElastiCache Redis instance (e.g. serverless) 
2. Connect to the ElastiCache Redis instance using a Rueidis client
3. The ElastiCache metric `CommandAuthorizationFailures` should increase

Potential fix:
The Rueidis client has an option for `ClientSetInfo` but there is no option to disable it entirely for support with older versions of Redis. An option can be added to the client options type to disable the execution of this command.

rueian:
Instead of another option, I would prefer this:

```go
var DisableClientSetInfo = make([]string)
...
rueidis.NewClient(rueidis.ClientOption{
	ClientSetInfo: rueidis.DisableClientSetInfo,
})
```

rahulgkatre:
I've been trying your suggested fix out, diff here: https://github.com/rahulgkatre/rueidis/pull/1/files

The test I added for this change passes locally but in the workflow there are failures related to data races.

rahulgkatre:
Thanks for the help, workflows passed and I've opened the PR to this repo.

PR #622, opened 2024-08-29
Add option to disable making `CLIENT SETINFO` command on initialization

rahulgkatre:
This pull request addresses #620 and adds a `DisableClientSetInfo` constant to be used for the `ClientSetInfo` field of `ClientOption`. 

When this constant is used, the `CLIENT SETINFO` command and relevant paths are skipped during new connection initialization, in order to be compatible with older versions of Redis (<7.2) that do not have the `CLIENT SETINFO` command.

- Add DisableClientSetInfo path and test

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -168,10 +168,14 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps, nobg 
 	if option.ClientNoEvict {
 		init = append(init, []string{"CLIENT", "NO-EVICT", "ON"})
 	}
+
+	addClientSetInfoCmds := true
 	if len(option.ClientSetInfo) == 2 {
 		init = append(init, []string{"CLIENT", "SETINFO", "LIB-NAME", option.ClientSetInfo[0]}, []string{"CLIENT", "SETINFO", "LIB-VER", option.ClientSetInfo[1]})
-	} else {
+	} else if option.ClientSetInfo == nil {
 		init = append(init, []string{"CLIENT", "SETINFO", "LIB-NAME", LibName}, []string{"CLIENT", "SETINFO", "LIB-VER", LibVer})
+	} else {
+		addClientSetInfoCmds = false
 	}
 
 	timeout := option.Dialer.Timeout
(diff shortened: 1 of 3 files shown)


pipe.go
```diff
@@ -249,16 +260,28 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps, nobg
...
 			defer resultsp.Put(resp)
-			for i, r := range resp.s[:len(resp.s)-2] { // skip error checking on the last CLIENT SETINFO
+
+			count := len(resp.s)
+			if addClientSetInfoCmds {
+				// skip error checking on the last CLIENT SETINFO
+				count -= 2
+			}
+
+			for i, r := range resp.s[:len(resp.s)-2] {
```
rueian, line 284:
```suggestion
			for i, r := range resp.s[:count] {
```

rahulgkatre: Thanks for catching this

pipe_test.go
```diff
@@ -491,6 +491,36 @@ func TestNewPipe(t *testing.T) {
...
+				ReplyString("OK")
+		}()
+		if err != nil {
+			t.Fatalf("pipe setup failed: %v", err)
+		}
+		p.Close()
+		mock.Close()
+		n1.Close()
+		n2.Close()
+	})
```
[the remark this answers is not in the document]
rahulgkatre: Added a similar test to `TestNewRESP2Pipe`

rahulgkatre:
Is there a way I can create a release of this change to use in my project?

rueian:
> Is there a way I can create a release of this change to use in my project?

I can tag a new alpha version tomorrow.

reagankan:
Hello @rueian, have you had a chance to create an alpha version? We need to consume this ASAP, preferably today or monday/tuesday Sept 2,3. Thanks in advance.

rueian:
Hi @rahulgkatre, @reagankan,

I have tagged the v1.0.46-alpha.1. You can change your go.mod to that.

I would also like to know why you need this. I thought the `CLIENT SETINFO` should be in a fire-and-forget manner and should not block you.

PR #625, opened 2024-08-30
Support iter.Seq2 for *FieldValue

### LINKED ISSUE #355: Support Go 1.23 range over func in internal/cmds
I know Builder respects Redis command argument order, but this is a little annoying in some situation.

For example, please consider maxlen is set only when limit > 0 for XADD.
Currently I have to write like blow.
```go
func xadd(ctx context.Context, conn rueidis.Client, key string, limit int, attributes map[string]string) error {
	xadd := conn.B().Xadd().Key(key)
	var completed rueidis.Completed
	if limit > 0 {
		fieldValue := xadd.Maxlen().Almost().Threshold(strconv.Itoa(limit)).Id("*").FieldValue()
		for k, v := range attributes {
			fieldValue = fieldValue.FieldValue(k, v)
		}
		completed = fieldValue.Build()
	} else {
		fieldValue := xadd.Id("*").FieldValue()
		for k, v := range attributes {
			fieldValue = fieldValue.FieldValue(k, v)
		}
		completed = fieldValue.Build()
	}
	return conn.Do(ctx, completed).Error()
}
```

If XaddFieldValue is exported, I can write like below.

```go
func xadd(ctx context.Context, conn rueidis.Client, key string, limit int, attributes map[string]string) error {
	xadd := conn.B().Xadd().Key(key)
	var fieldValue rueidis.XaddFieldValue
	if limit > 0 {
		fieldValue = xadd.Maxlen().Almost().Threshold(strconv.Itoa(limit)).Id("*").FieldValue()
	} else {
		fieldValue = xadd.Id("*").FieldValue()
	}
	for k, v := range attributes {
		fieldValue = fieldValue.FieldValue(k, v)
	}
	return conn.Do(ctx, fieldValue.Build()).Error()
}
```

tamayika:
closes #355

I added iter.Seq2 support for `HSET`, `HMSET`, `XADD` which have `FieldValue()` method.
We must support `< go 1.23`, so we must split file to other files.
It's hard to support auto generation because current implementation does not support code splitting.
So I hand-written the implementation.
When go version in `go.mod` is `go 1.23`, I think we can support auto generation by adding flags to `*.json` and its structs.

### Consideration

I named `FieldValues`, but this is maybe not good to support auto generation in the future because we must consider plural form correctly.
Is `FieldValueSeq` better?

### DIFF
--- a/internal/cmds/iter_test.go
+++ b/internal/cmds/iter_test.go
@@ -4,10 +4,16 @@ package cmds
 
 import (
 	"maps"
+	"testing"
 )
 
 func iter0(s Builder) {
-	s.Hmset().Key("1").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
-	s.Hset().Key("1").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
-	s.Xadd().Key("1").Id("*").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
+	s.Hmset().Key("1").FieldValue().FieldValueIter(maps.All(map[string]string{"1": "1"})).Build()
+	s.Hset().Key("1").FieldValue().FieldValueIter(maps.All(map[string]string{"1": "1"})).Build()
+	s.Xadd().Key("1").Id("*").FieldValue().FieldValueIter(maps.All(map[string]string{"1": "1"})).Build()
+}
+
+func TestIter(t *testing.T) {
+	var s = NewBuilder(InitSlot)
+	t.Run("0", func(t *testing.T) { iter0(s) })
 }
--- a/internal/cmds/iter_test.go
+++ b/internal/cmds/iter_test.go
@@ -1,0 +1,13 @@
+//go:build go1.23
+
+package cmds
+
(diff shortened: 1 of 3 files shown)


rueian:
> Is `FieldValueSeq` better?

I think your original `FieldValueIter` is better.

internal/cmds/iter_test.go
```diff
@@ -0,0 +1,13 @@
...
+package cmds
+
+import (
+	"maps"
+)
+
+func iter0(s Builder) {
+	s.Hmset().Key("1").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
+	s.Hset().Key("1").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
+	s.Xadd().Key("1").Id("*").FieldValue().FieldValues(maps.All(map[string]string{"1": "1"})).Build()
```
rueian, line 12: We need to call this function somewhere in tests.

tamayika: thanks, I fixed it.

rueian (approved):

PR #638, opened 2024-09-17
feat: add clutser topology awareness

proost:
discussion: https://github.com/redis/rueidis/discussions/637

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -73,7 +73,7 @@ func newClusterClient(opt *ClientOption, connFn connFn) (*clusterClient, error) 
 		return nil, err
 	}
 
-	if err := client.refresh(context.Background(), true); err != nil {
+	if err := client.refresh(context.Background()); err != nil {
 		return client, err
 	}
 
@@ -121,16 +121,12 @@ func (c *clusterClient) init() error {
 	return es[0]
 }
 
-func (c *clusterClient) refresh(ctx context.Context, forceRefresh bool) (err error) {
-	return c.sc.Do(ctx, func() error {
-		return c._refresh(forceRefresh)
-	})
+func (c *clusterClient) refresh(ctx context.Context) (err error) {
+	return c.sc.Do(ctx, c._refresh)
 }
 
 func (c *clusterClient) lazyRefresh() {
-	c.sc.LazyDo(time.Second, func() error {
-		return c._refresh(false)
-	})
+	c.sc.LazyDo(time.Second, c._refresh)
 }
 
 type clusterslots struct {
@@ -156,7 +152,7 @@ func getClusterSlots(c conn, timeout time.Duration) clusterslots {
 	return clusterslots{reply: c.Do(ctx, cmds.ShardsCmd), addr: c.Addr(), ver: v}
 }
 
-func (c *clusterClient) _refresh(forceRefresh bool) (err error) {
+func (c *clusterClient) _refresh() (err error) {
 	c.mu.RLock()
 	results := make(chan clusterslots, len(c.conns))
 	pending := make([]conn, 0, len(c.conns))
@@ -208,39 +204,19 @@ func (c *clusterClient) _refresh(forceRefresh bool) (err error) {
 
 	var removes []conn
 
-	if !forceRefresh {
-		shouldRefresh := false
-
-		c.mu.RLock()
-		// check if the current topology is different from the new one
-		for addr, cc := range c.conns {
-			fresh, ok := conns[addr]
-			if ok && (cc.replica == fresh.replica || c.rOpt == nil) {
-				conns[addr] = connrole{
-					conn:    cc.conn,
-					replica: fresh.replica,
-				}
-			} else {
-				shouldRefresh = true
-				removes = append(removes, cc.conn)
+	c.mu.RLock()
+	for addr, cc := range c.conns {
+		fresh, ok := conns[addr]
+		if ok && (cc.replica == fresh.replica || c.rOpt == nil) {
+			conns[addr] = connrole{
+				conn:    cc.conn,
+				replica: fresh.replica,
 			}
-		}
-		// check if the new topology is different from the current one
-		if !shouldRefresh {
-			for addr, cc := range conns {
-				old, ok := c.conns[addr]
-				if !ok || old.replica != cc.replica {
-					shouldRefresh = true
-					break
-				}
-			}
-		}
-		c.mu.RUnlock()
-
-		if !shouldRefresh {
-			return nil
+		} else {
+			removes = append(removes, cc.conn)
 		}
 	}
+	c.mu.RUnlock()
 
 	pslots := [16384]conn{}
 	var rslots []conn
@@ -425,7 +401,7 @@ func (c *clusterClient) _pick(slot uint16, toReplica bool) (p conn) {
 
 func (c *clusterClient) pick(ctx context.Context, slot uint16, toReplica bool) (p conn, err error) {
 	if p = c._pick(slot, toReplica); p == nil {
-		if err := c.refresh(ctx, false); err != nil {
+		if err := c.refresh(ctx); err != nil {
 			return nil, err
 		}
 		if p = c._pick(slot, toReplica); p == nil {
@@ -607,7 +583,7 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, last 
 func (c *clusterClient) pickMulti(ctx context.Context, multi []Completed) (*connretry, uint16, bool, error) {
 	conns, slot, toReplica := c._pickMulti(multi)
 	if conns == nil {
-		if err := c.refresh(ctx, false); err != nil {
+		if err := c.refresh(ctx); err != nil {
 			return nil, 0, false, err
 		}
 		if conns, slot, toReplica = c._pickMulti(multi); conns == nil {
@@ -887,7 +863,7 @@ func (c *clusterClient) _pickMultiCache(multi []CacheableTTL) *connretrycache {
 func (c *clusterClient) pickMultiCache(ctx context.Context, multi []CacheableTTL) (*connretrycache, error) {
 	conns := c._pickMultiCache(multi)
 	if conns == nil {
-		if err := c.refresh(ctx, false); err != nil {
+		if err := c.refresh(ctx); err != nil {
 			return nil, err
 		}
 		if conns = c._pickMultiCache(multi); conns == nil {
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -18,19 +18,21 @@ import (
...
+	opt         *ClientOption
+	rOpt        *ClientOption
+	conns       map[string]connrole
+	connFn      connFn
+	sc          call
+	mu          sync.RWMutex
+	stop        uint32
+	cmd         Builder
+	retry       bool
+	refreshStop chan struct{}
```
rueian, line 35:
```suggestion
	stopCh chan struct{}
```
The channel can be used by other goroutines in the future. Let's not limit its naming to the current purpose.

cluster.go
```diff
@@ -1018,6 +1088,8 @@ func (c *clusterClient) Nodes() map[string]Client {
 }
 
 func (c *clusterClient) Close() {
+	close(c.refreshStop)
```
rueian, line 1091: Use the `c.stop` atomic flag to avoid closing the channel more than once.

cluster.go
```diff
@@ -358,6 +381,53 @@ func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]g
...
+			// check if the current topology is different from the new one
+			if !isChanged {
+				for addr := range c.conns {
+					if _, ok := conns[addr]; !ok {
+						isChanged = true
+						break
+					}
+				}
+			}
+			c.mu.RUnlock()
```
rueian, line 422: Why not put this check in the lazyRefresh? Then we can simply call lazyRefresh periodically here.

cluster.go
proost: [1181ffb](https://github.com/redis/rueidis/pull/638/commits/1181ffbe8041a47911e438c99bb2464e46a6d00e)

cluster.go
proost: [1181ffb](https://github.com/redis/rueidis/pull/638/commits/1181ffbe8041a47911e438c99bb2464e46a6d00e)

cluster.go
proost: I don't understand. this check logic is for refresh only topology is changed. if you want to make a new method, i added it. [027671b](https://github.com/redis/rueidis/pull/638/commits/027671b8a08e824c58cc35a069c8cf980c1abc49)

proost: Ok, how about this?

[b075e69](https://github.com/redis/rueidis/pull/638/commits/b075e699c15affca0f3595532edcf9120cf7b0ce)

cluster.go
```diff
@@ -358,6 +380,94 @@ func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]g
...
 
+func (c *clusterClient) runClusterTopologyRefreshment() {
+	ticker := time.NewTicker(c.opt.ClusterTopologyRefreshmentOption.ScanInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-c.stopCh:
+			return
+		case <-ticker.C:
+			c.conditionalRefresh()
```
rueian, line 391: Yes, now there is no duplicated `getClusterTopology` at the end. However, I still want all the refresh processes, including the new conditional one, to be gated by the `lazyRefresh` function to ensure at most one `getClusterTopology` per second and not to flood the servers. `lazyRefresh` serves the same purpose as your initial cool-down period proposal. Can we merge the `_refresh` and `conditionalRefresh` and then just use `lazyRefresh` here?
```suggestion
			c.lazyRefresh()
```

proost: [826de2c](https://github.com/redis/rueidis/pull/638/commits/826de2c76a7906b25727ee529ba91d28622d6457)

just call lazyRefresh

rueian: Yes, I am ok with this one, but why do you delete your detection logic?

rueidis.go
```diff
@@ -201,6 +204,14 @@ type SentinelOption struct {
 	ClientName string
 }
 
+// ClusterTopologyRefreshmentOption is the options for the cluster topology refreshment.
+// Cluster Refresh happens only when the cluster topology is not up-to-date.
+type ClusterTopologyRefreshmentOption struct {
```
rueian, line 209: ClusterTopologyRefreshmentOption is too limited. How about just ClusterOption?

cluster.go
proost: Oh, i thought you want to use original one. i changed [9fd1da9](https://github.com/redis/rueidis/pull/638/commits/9fd1da9a6bff4f2cf0e05e77ca66135c9aa50250)

rueidis.go
proost: Ok, [09a9403](https://github.com/redis/rueidis/pull/638/commits/09a9403301ee2aecab963f684a7c499f93a79039)

cluster.go
proost: [fa0b58a](https://github.com/redis/rueidis/pull/638/commits/fa0b58a22e8af7dc63d426d4bd0423c4c8182e25)

We must not merge `_refresh` and `conditionalRefresh`.

`_refresh` is called,
1. In a constructor. This case must init all conns, pslots, rslots. so should be all logic must be called.
2. When got error from redis cluster. This case must be refresh even if cluster is not modified. because a connection is band.

cluster.go
```diff
@@ -168,31 +212,16 @@ func (c *clusterClient) _refresh() (err error) {
...
-	}
-	// make sure InitAddress always be present
-	for _, addr := range c.opt.InitAddress {
-		if _, ok := conns[addr]; !ok {
-			conns[addr] = connrole{
-				conn: c.connFn(addr, c.opt),
-			}
-		}
-	}
+	return
```
rueian, line 219: These lines can be replaced with just a `return`.

cluster.go
```diff
@@ -262,6 +291,79 @@ func (c *clusterClient) _refresh() (err error) {
...
+			}
+		} else {
+			cc.conn = c.connFn(addr, c.opt)
+		}
+
+		conns[addr] = connrole{
+			conn:    cc.conn,
+			replica: cc.replica,
+		}
+	}
```
rueian, line 364: This loop costs nothing because `connFn` doesn't connect to the address immediately. So, you don't need to separate it from the loop above.

cluster.go
rueian: Would 1 be an issue if we made `_refresh` conditionally? I thought in that case the checking logic would be bypassed.

For 2 and 3, I don't get your points of not making `_refresh` conditionally. They are not related to the refresh process but are more related to how we act on the result of it.

proost: [c02c256](https://github.com/redis/rueidis/pull/638/commits/c02c2562a7276d3bf265651534aab1e7403bf0cf)

cluster.go
```diff
@@ -193,6 +202,40 @@ func (c *clusterClient) _refresh() (err error) {
...
+	// check if cluster client is initialized.
+	if !shouldRefresh {
+		for _, cc := range c.pslots {
+			if cc == nil {
+				shouldRefresh = true
+				break
+			}
+		}
+	}
```
rueian, line 232: Is this check really necessary? Also, it is valid that a living cluster has some missing slots.

cluster.go
```diff
@@ -193,6 +202,40 @@ func (c *clusterClient) _refresh() (err error) {
...
+	// check if the current topology is different from the new one
+	if !shouldRefresh {
+		for addr := range c.conns {
+			if _, ok := conns[addr]; !ok {
+				shouldRefresh = true
+				break
+			}
+		}
+	}
```
rueian, line 223: Can we merge this loop with the below `removes` loop (L241-L253)? So that we don't need to acquire Rlock and do hashmap access again.

proost: [9f886c5](https://github.com/redis/rueidis/pull/638/commits/9f886c503d27f212422fdedc97f1b3c050eff45c)

cluster.go
proost: Oh, yes. you are right. 
[9f886c5](https://github.com/redis/rueidis/pull/638/commits/9f886c503d27f212422fdedc97f1b3c050eff45c)

cluster.go
```diff
@@ -195,19 +208,39 @@ func (c *clusterClient) _refresh() (err error) {
...
 	var removes []conn
 
-	c.mu.RLock()
-	for addr, cc := range c.conns {
-		fresh, ok := conns[addr]
-		if ok && (cc.replica == fresh.replica || c.rOpt == nil) {
-			conns[addr] = connrole{
-				conn:    cc.conn,
-				replica: fresh.replica,
+	if !forceRefresh {
```
rueian, line 211: I think we are almost done. But is this `forceRefresh` flag necessary? What will happen if we don't use the flag?

proost: We need. Because if `conns` is same as `c.conns` after `init()` called, it doesn't update `pslots` and `rslots` in `_refresh`. we need forced refresh in `_refresh`

rueian: Yes, you are right. I overlooked that. However, that shows a problem with the current checking logic: slot changes are not considered. We need to consider them as well, otherwise, we will miss them.

To make the comparison easy, I think we may need to keep track of some simple hashes from the `parseSlots` and `parseShards`.

proost: >  slot changes are not considered.

Yes, I missed that point. how about just refresh always? just refreshing is more simpler and at this point, i don't think we can get many advantages using conditional refresh.

rueian: Ok, we can find a way to do that later.

proost: [3712efd](https://github.com/redis/rueidis/pull/638/commits/3712efd3a6d66db27019ae895c5bda0599ab5b3f)

rueidis.go
```diff
@@ -201,6 +204,14 @@ type SentinelOption struct {
 	ClientName string
 }
 
+// ClusterOption is the options for the redis cluster client.
+type ClusterOption struct {
+	// ScanInterval is the interval to scan the cluster topology.
+	// If the value is zero, refreshment will be disabled.
+	// Cluster topology cache refresh happens always in the background after successful scan.
+	ScanInterval time.Duration
```
rueian, line 212:
```suggestion
	ShardsRefreshInterval time.Duration
```

proost: [3c97bc8](https://github.com/redis/rueidis/pull/638/commits/3c97bc8f0a770ec8d630f7568fa8cf940aa4c477)

rueian (approved):

rueian:
@proost merged. Thanks a lot!

Issue #633, opened 2024-09-13
Feature: retry backoffs


rueian:
There are about 25 `goto retry` statements in rueidis. Such as:

https://github.com/redis/rueidis/blob/5135acdc65c815cdce8e8b0c12c1e5aa939504bd/cluster.go#L460-L462
https://github.com/redis/rueidis/blob/5135acdc65c815cdce8e8b0c12c1e5aa939504bd/client.go#L48-L50
https://github.com/redis/rueidis/blob/5135acdc65c815cdce8e8b0c12c1e5aa939504bd/sentinel.go#L66-L68

And so on.

It will be better to add backoffs before ALL these statements to avoid rapid retry storms and those backoffs should have the following properties:

1. Should be capped for at most 1 second.
2. Should not exceed the context deadline.
3. Backoff exponentially with jitter.

Related to https://github.com/valkey-io/valkey-doc/pull/164

proost:
It's good idea.

Maybe later someone want own retry logic. So default retry logic make as you said, but how about make a retrier function like this?

```
// Retrier is a function that will be called when a command is retryable.
// The function should return an error if the command should not be retried.
type Retrier func(ctx context.Context, attempts int) error 
```

rueian:
@proost sounds good, but the returned 'error' makes the Retrier confusing in terms of its responsibility.

How about just return a time.Duration and name the function RetryDelay?

proost:
Ok, I understand what's your intention. So Let's clear interface. 

How do you think?
```
// RetryDelay returns the delay that should be used before retrying the
// attempt. Will return error if the delay could not be determined or do not retry.
type RetryDelay func(attempts int, err error) (time.Duration, error)
```

If you agree with definition, I will make a feature.

rueian:
Returning an error still seems confusing to me. We can use negative duration to stop retrying. I think returning a duration only is much more clear.

One more thing. If the return duration exceeds the context deadline, we should not retry. So there might be some duplications of checking the context deadline in the old isRetryable function and the new RetryDelay that we had better keep in one place.

proost:
@rueian how about this?

```
// RetryDelay returns the delay that should be used before retrying the
// attempt. Will return negative delay if the delay could not be determined or do not retry.
type RetryDelay func(attempts int, err error) time.Duration
```

[...]

PR #643, opened 2024-10-03
feat: add backoff and jitter to retry

proost:
close: https://github.com/redis/rueidis/issues/633

Adding retry policy.

default is exponential backoff and jitter.

clean logic for retry handler is not easy. If have a idea, give me review.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -480,15 +480,15 @@ process:
 		resultsp.Put(results)
 		goto process
 	case RedirectRetry:
-		shouldRetry, err := c.retryHandler.WaitUntilNextRetry(
-			ctx, func() bool { return c.retry && cmd.IsReadOnly() }, attempts, resp.Error(),
-		)
-		if shouldRetry {
-			attempts++
-			goto retry
-		}
-		if err != nil {
-			return newErrResult(err)
+		if c.retry && cmd.IsReadOnly() {
+			shouldRetry, err := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
+			if err != nil {
+				return newErrResult(err)
+			}
 		}
 	}
 	return resp
@@ -613,9 +613,9 @@ func (c *clusterClient) doresultfn(
 		if mode != RedirectNone {
 			nc := cc
 			if mode == RedirectRetry {
-				if !isWaitingForRetry {
+				if c.retry && cm.IsReadOnly() && !isWaitingForRetry {
 					shouldRetry, errAbortingWaiting := c.retryHandler.WaitUntilNextRetry(
-						ctx, func() bool { return c.retry && cm.IsReadOnly() }, attempts, resp.Error(),
+						ctx, attempts, resp.Error(),
 					)
 					if errAbortingWaiting != nil {
 						select {
@@ -766,19 +766,16 @@ process:
 				goto process
 			}
 		case RedirectRetry:
-			shouldRetry, errAbortWaiting := c.retryHandler.WaitUntilNextRetry(
-				ctx,
-				func() bool { return c.retry && allReadOnly(multi) },
-				attempts,
-				resp.Error(),
-			)
-			if shouldRetry {
-				resultsp.Put(resps)
-				attempts++
-				goto retry
-			}
-			if errAbortWaiting != nil {
-				return fillErrs(len(multi), errAbortWaiting)
+			if c.retry && allReadOnly(multi) {
+				shouldRetry, err := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+				if shouldRetry {
+					resultsp.Put(resps)
+					attempts++
+					goto retry
+				}
+				if err != nil {
+					return fillErrs(len(multi), err)
+				}
 			}
 		}
 	}
@@ -805,15 +802,15 @@ process:
 		resultsp.Put(results)
 		goto process
 	case RedirectRetry:
-		shouldRetry, err := c.retryHandler.WaitUntilNextRetry(
-			ctx, func() bool { return c.retry }, attempts, resp.Error(),
-		)
-		if shouldRetry {
-			attempts++
-			goto retry
-		}
-		if err != nil {
-			return newErrResult(err)
+		if c.retry {
+			shouldRetry, err := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
+			if err != nil {
+				return newErrResult(err)
+			}
 		}
 	}
 	return resp
@@ -949,10 +946,8 @@ func (c *clusterClient) resultcachefn(
 		if mode != RedirectNone {
 			nc := cc
 			if mode == RedirectRetry {
-				if !isWaitingForRetry {
-					shouldRetry, errAbortingWaiting := c.retryHandler.WaitUntilNextRetry(
-						ctx, func() bool { return c.retry }, attempts, resp.Error(),
-					)
+				if c.retry && !isWaitingForRetry {
+					shouldRetry, errAbortingWaiting := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
 					if errAbortingWaiting != nil {
 						select {
 						case abortingRetryWaitingErrCh <- errAbortingWaiting:
@@ -1071,21 +1066,16 @@ retry:
 	}
 	err = cc.Receive(ctx, subscribe, fn)
 	if _, mode := c.shouldRefreshRetry(err, ctx); mode != RedirectNone {
-		shouldRetry, errAborting := c.retryHandler.WaitUntilNextRetry(
-			ctx,
-			func() bool {
-				return c.retry
-			},
-			attempts,
-			err,
-		)
-		if shouldRetry {
-			attempts++
-			goto retry
-		}
+		if c.retry {
+			shouldRetry, errAborting := c.retryHandler.WaitUntilNextRetry(ctx, attempts, err)
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
 
-		if errAborting != nil {
-			err = errAborting
+			if errAborting != nil {
+				err = errAborting
+			}
 		}
 	}
 ret:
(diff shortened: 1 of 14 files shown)


retry.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func newRetryer(retryDelay RetryDelay) *retryer {
+	return &retryer{RetryDelay: retryDelay}
+}
+
+func (r *retryer) WaitUntilNextRetry(
+	ctx context.Context, isRetryable retryableCheckFunc, attempts int, err error,
+) (bool, error) {
+	if !isRetryable() {
```
rueian, line 52: I don't like the `isRetryable retryableCheckFunc` to be passed in here. I think `isRetryable` should be checked before calling the `WaitUntilNextRetry`.

retry.go
```diff
@@ -0,0 +1,70 @@
...
+	delay := r.RetryDelay(attempts, err)
+	if delay < 0 {
+		return false, nil
+	}
+
+	timer := time.NewTimer(delay)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
```
rueian, line 65: We should not wait for the context deadline but check whether the deadline will be exceeded or not:

```go
	dl, ok := ctx.Deadline()
	if ok && time.Until(dl) > delay {
		time.Sleep(delay)
	}
```

proost: Is there no need to respect context cancel ?

retry.go
proost: [9e884c3](https://github.com/redis/rueidis/pull/643/commits/9e884c395cb0aa8c382e10c1dd92a514130c1efb). OK.

retry.go
rueian: It is better to respect the cancellation, but the returned `errAbortWaiting` complicates the code outside.

How about we do this, removing all the return values:

```go
	ctx := context.Background()
	if dl, ok := ctx.Deadline(); !ok || time.Until(dl) > delay {
		if ch := ctx.Done(); ch != nil {
			tm := time.NewTimer(delay)
			defer tm.Stop()
			select {
			case <-ch:
			case <-tm.C:
			}
		} else {
			time.Sleep(delay)
		}
	}
```

rueian: We still let it retry if the context is ended. The underlying pipe will check the ctx.Err().

retry.go
```diff
@@ -0,0 +1,62 @@
...
+// attempt. Will return negative delay if the delay could not be determined or do not retry.
+type RetryDelay func(attempts int, err error) time.Duration
+
+// defaultRetryDelay delays the next retry exponentially without considering the error.
+// max delay is 1 second.
+func defaultRetryDelay(attempts int, _ error) time.Duration {
+	base := util.FastRandFloat64()
+	backoff := uint64(1 << uint64(math.Min(float64(attempts), maxAttemptsShift)))
+	jitter := base * float64(backoff) * float64(time.Millisecond)
+	return time.Duration(math.Min(jitter, float64(defaultMaxRetryDelay)))
```
rueian, line 26: We should apply randomness only to the jitter part. The delay produced by the current implementation is not monotonically increasing.

Can we avoid using float64, including the `base`? We can use the built-in `min` and `max` functions.

rueian: For example:

```go
	backoff := time.Microsecond * 5 * (1 << min(20, attempts))
	jitter := time.Microsecond * time.Duration(rand.Int64N(1<<min(20, attempts)))
	return min(time.Second, backoff+jitter)
```

retry.go
proost: outside code of `WaitUntilNextRetry` can be complicated. 

I feel that this is bit more code style. when not returning error, caller should always check context is done or ,as you can said, `pipe` always will check and all contributors should know that. It is unclear to me, so that complicated but explicit code is more sense to me. I change it, but can you give me more detailed reason?

retry.go
proost: Well, actually above logic came from [here](https://aws.amazon.com/ko/blogs/architecture/exponential-backoff-and-jitter/).
> The delay produced by the current implementation is not monotonically increasing.

Intended part. To smooth multiple client calls.

retry.go
proost: [ea85114](https://github.com/redis/rueidis/pull/643/commits/ea8511464d7b71612c061b9366380764138303cc)

retry.go
proost: [d172784](https://github.com/redis/rueidis/pull/643/commits/d172784d4127a8955dcfe1ca9c3caf586fba7872)

retry.go
rueian: > pipe always will check and all contributors should know that. It is unclear to me

It is indeed implicit. But the reason for not returning an error is simple: adding a new error path to replace a `goto retry` statement seems doing too much.

cluster.go
```diff
@@ -479,8 +481,14 @@ process:
...
-			runtime.Gosched()
-			goto retry
+			shouldRetry := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
+			if err != nil {
+				return newErrResult(err)
+			}
```
rueian, line 491: Could this be removed?

cluster.go
```diff
@@ -720,16 +756,24 @@ process:
...
-				goto retry
+				shouldRetry := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+				if shouldRetry {
+					resultsp.Put(resps)
+					attempts++
+					goto retry
+				}
+				if err != nil {
+					return fillErrs(len(multi), err)
+				}
```
rueian, line 767: Could this be removed?

cluster.go
```diff
@@ -748,8 +792,14 @@ process:
...
-			runtime.Gosched()
-			goto retry
+			shouldRetry := c.retryHandler.WaitUntilNextRetry(ctx, attempts, resp.Error())
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
+			if err != nil {
+				return newErrResult(err)
+			}
```
rueian, line 802: Could this be removed?

cluster.go
```diff
@@ -605,6 +616,18 @@ func (c *clusterClient) doresultfn(ctx context.Context, results *redisresults, r
...
+					shouldRetry := c.retryHandler.WaitUntilNextRetry(
+						ctx, attempts, resp.Error(),
+					)
+
+					if !shouldRetry {
+						continue
+					}
+
+					isWaitingForRetry = true
+				}
```
rueian, line 630: Why do you put the waiting here? Instead of putting it at: https://github.com/redis/rueidis/blob/fc348f545e5ac0315ef9350fa7375bda92b8877e/cluster.go#L679-L681

cluster.go
proost: [1725eb5](https://github.com/redis/rueidis/pull/643/commits/1725eb5ca04b3c745838debe0cfa555fa2c2e227)

cluster.go
proost: [1725eb5](https://github.com/redis/rueidis/pull/643/commits/1725eb5ca04b3c745838debe0cfa555fa2c2e227)

cluster.go
proost: If we sleep not here, code is quite complicated. Because we have to distinguish retry or ask/moved in the L679. if retry, should wait. if not retry immediately.

proost: Even though there is randomness, i think waiting time for each goroutines, converges similarly. If you want exactly once to wait, i will change it.

rueian: Thanks for catching this detail I overlooked.

I think a better way to handle this is if any move/ask redirection happens, we don't back off for a retry.

Therefore, how about we add a new counter field to the `retries` struct? So that we can do this:

```go
	if len(retries.m) != 0 {
		if retries.redirects != 0 {
			retries.redirects = 0
			goto retry
		}
		if c.retryHandler.WaitUntilNextRetry(...) {
			goto retry
		}
	}
```

rueian: We can increment the counter after the lock:
image

proost: [2df785e](https://github.com/redis/rueidis/pull/643/commits/2df785eacfc7bd765452531bb96d763e21c0dcea)

We support configurable retry or not so should give error to `RetryDelay`. Because of this i use channel to gather errors.

We can't use just int. because it is not thread-safe. I prefer `atomic` in this case, because the type itself guarantees   thread-safe. Even though  `doretry` and `doretrycache` use mutex inside, but it is implicit to me to be thread-safe when load int value.

rueian: Hi @proost,

I think a `retryErrCh := make(chan error, len(multi))` is too over kill. Let me find a better way for you later.

> but it is implicit to me to be thread-safe when load int value.

It is actually thread-safe, because we read the field after a `wg.Wait()`.

rueian: Atomic is ok too, but please just use `uint32` which is smaller and has no alignment issue.

rueian: Hi @proost,

I think a better alternative for `retryErrCh` is calling `RetryDelay` directly in the `doresultfn` function and we store its maximum into the `retries` struct. Could be something like this:

image

syncp.go
```diff
@@ -215,8 +219,9 @@ func (r *conncount) ResetLen(n int) {
 }
 
 type connretry struct {
-	m map[conn]*retry
-	n int
+	m         map[conn]*retry
+	n         int
+	Redirects atomic.Uint64
```
rueian, line 224: Remember to reset its value to 0 when we put the struct back into the sync.Pool.

```go
func (r *connretry) ResetLen(n int) {
	clear(r.m)
	r.Redirects.Store(0) // <- remember to do this
}
```

cluster.go
```diff
@@ -666,18 +683,35 @@ func (c *clusterClient) DoMulti(ctx context.Context, multi ...Completed) []Redis
...
+			if shouldRetry {
+				attempts++
+				goto retry
+			}
+		}
+
+		if retries.Redirects.Load() > 0 {
+			retries.Redirects.Store(0) // reset
+			goto retry
+		}
```
rueian, line 714: I think `if retries.Redirects.Load() > 0` should have a higher priority than `for retryableErr := range retryErrCh`.

If there is a redirect response, we should not delay the retry.

syncp.go
proost: I missed it. Thank you!
[f04b433](https://github.com/redis/rueidis/pull/643/commits/f04b433a99cf7b644720617ea5e4c5457ad9882b)

cluster.go
proost: [f04b433](https://github.com/redis/rueidis/pull/643/commits/f04b433a99cf7b644720617ea5e4c5457ad9882b)

cluster.go
proost: > It is actually thread-safe, because we read the field after a wg.Wait().

Oh, i missed waiting using `sync.WaitGroup`. 

change to use `retries.RetryDelay`.
[f04b433](https://github.com/redis/rueidis/pull/643/commits/f04b433a99cf7b644720617ea5e4c5457ad9882b)

rueian (approved):
I think this PR looks very good now. We can merge it. Thank you @proost for this great work! Really impressive.

One more thing I want to optimize is the positions of the newly added `retryHandler` field:
image

Putting it after a bool can cause some space waste, could you rearrange the field in a followup PR? Also, can `retryHandler` be just a struct instead of a pointer to struct?

proost:
@rueian 
Sure. I will change in a followup PR.

Issue #644, opened 2024-10-07
Support Go 1.23 iterators for the ZADD command


rueian:
We support Go 1.23 iterators for HMSET, HSET, and XADD:

https://github.com/redis/rueidis/blob/fc348f545e5ac0315ef9350fa7375bda92b8877e/internal/cmds/iter.go#L7-L26

It would be better to add support for ZADD as well.

rueian:
Thanks @mingdaoy!

PR #652, opened 2024-10-25
Support Go 1.23 iterators for the ZADD command 

mingdaoy:
https://github.com/redis/rueidis/issues/644

### DIFF
--- a/internal/cmds/iter.go
+++ b/internal/cmds/iter.go
@@ -3,8 +3,8 @@
 package cmds
 
 import (
-	"fmt"
 	"iter"
+	"strconv"
 )
 
 func (c HmsetFieldValue) FieldValueIter(seq iter.Seq2[string, string]) HmsetFieldValue {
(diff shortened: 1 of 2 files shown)


internal/cmds/iter.go
```diff
@@ -24,3 +27,10 @@ func (c XaddFieldValue) FieldValueIter(seq iter.Seq2[string, string]) XaddFieldV
 	}
 	return c
 }
+
+func (c ZaddScoreMember) ScoreMemberIter(seq iter.Seq2[float64, string]) ZaddScoreMember {
```
rueian, line 31:
```suggestion
func (c ZaddScoreMember) ScoreMemberIter(seq iter.Seq2[string, float64]) ZaddScoreMember {
```
Using [string, float64] will be much more convenient for users because the first value usually refers to a unique key. 

And Redis sorted set allows elements with the same score, so we better put float64 at the second value.

internal/cmds/iter.go
```diff
@@ -24,3 +27,10 @@ func (c XaddFieldValue) FieldValueIter(seq iter.Seq2[string, string]) XaddFieldV
...
+	for score, member := range seq {
+		c.cs.s = append(c.cs.s, fmt.Sprintf("%f", score), member)
```
rueian, line 33:
```suggestion
		c.cs.s = append(c.cs.s, strconv.FormatFloat(score, 'f', -1, 64), member)
```

`strconv.FormatFloat` is much more performant than `fmt.Sprintf`

Issue #656, opened 2024-10-28
Enhancement: Graceful Handling of Redis LOADING State in Rueidis


nesty92:
When connecting to a Redis server that is in the process of loading its dataset into memory (e.g., during startup or after a restart), Redis responds to client commands with a LOADING Redis is loading the dataset in memory error. Currently, Rueidis does not handle this state internally, which requires applications to implement custom logic to manage this scenario.

**Proposal:**

Enhance Rueidis to automatically detect and handle the LOADING state returned by Redis. This could involve:

- **Automatic Retry Mechanism:** When a LOADING error is received, Rueidis would retry the command after a configurable delay.
- **Configurable Settings:** Allow developers to set retry intervals, maximum retry attempts, and total timeout durations.
- **Exponential Backoff (Optional):** Implement an exponential backoff strategy to prevent overwhelming Redis during its loading phase.
- **Error Propagation:** If Redis remains in the LOADING state beyond the maximum timeout, Rueidis would return an appropriate error to the application.

**Use Case:**

In production environments with large datasets, Redis may take significant time to load data from disk into memory. During this period, clients attempting to send commands will receive a LOADING error. Handling this scenario within Rueidis would:

- **Improve Application Resilience:** Applications can automatically wait for Redis to become ready without custom retry logic.
- **Reduce Boilerplate Code:** Developers won't need to implement repetitive error handling across different applications.
- **Enhance Reliability:** Services depending on Redis can recover more gracefully from restarts or failovers.

**Current Workaround:**

Here's an example of how this issue is currently handled in application code:
```go
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
for {
    select {
    case <-ctx.Done():
        return nil, fmt.Errorf("context deadline exceeded while pinging redis: %w", ctx.Err())
    default:
        pingResp := client.Do(ctx,  client.B().Ping().Build())

        if pingResp.Error() == nil {
            return client, nil
        }

        if !strings.Contains(pingResp.Error().Error(), "LOADING Redis is loading the dataset in memory") {
            return nil, pingResp.Error()
        }

        time.Sleep(100 * time.Millisecond)
    }
}
```

**Redis Logs During Loading:**
```bash
redis-1 | 1:M 28 Oct 2024 10:00:22.750 * DB loaded from incr file appendonly.aof.334.incr.aof: 3.745 seconds
[...]

PR #657, opened 2024-10-29
feat: improve handling of LOADING response

nesty92:
When a Redis node is loading data (e.g. after restart), it responds with LOADING errors. This makes the client more resilient during Redis node restarts and initial data loading phases.

Close: #656

### DIFF
--- a/client.go
+++ b/client.go
@@ -47,7 +47,7 @@ func (c *singleClient) Do(ctx context.Context, cmd Completed) (resp RedisResult)
 	attempts := 1
 retry:
 	resp = c.conn.Do(ctx, cmd)
-	if c.retry && cmd.IsReadOnly() && c.isRetryable(resp.NonRedisError(), ctx) {
+	if c.retry && cmd.IsReadOnly() && c.isRetryable(resp.Error(), ctx) {
 		shouldRetry := c.retryHandler.WaitOrSkipRetry(
 			ctx, attempts, cmd, resp.Error(),
 		)
@@ -88,7 +88,7 @@ retry:
 	resps = c.conn.DoMulti(ctx, multi...).s
 	if c.retry && allReadOnly(multi) {
 		for i, resp := range resps {
-			if c.isRetryable(resp.NonRedisError(), ctx) {
+			if c.isRetryable(resp.Error(), ctx) {
 				shouldRetry := c.retryHandler.WaitOrSkipRetry(
 					ctx, attempts, multi[i], resp.Error(),
 				)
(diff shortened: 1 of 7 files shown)


nesty92:
Hi @rueian, making some progress here and wanted to confirm that I'm heading in the right direction.

I noticed that Ping wasn't marked as read-only. Is there a specific reason for this, or might it just be an oversight? I’ll wait for your feedback on this before continuing with the cluster client.

rueian:
Hi @nesty92,

I think we are on the right track, but I am concerned about marking PING as readonly. It is also not marked as readonly in the source https://github.com/redis/redis/blob/unstable/src/commands/ping.json.

Can we at least do that in a separate PR?

nesty92:
I agree with your concern about marking PING as readonly. I will handle this in a separate PR to keep the changes isolated and easier to review.

nesty92:
@rueian, I raised [this question](https://github.com/redis/redis/issues/13629) to get feedback from the Redis team regarding why the PING command is not marked as readonly

nesty92:
Hi @rueian, I removed the changes for making `PING` read-only, but handling it in some way is important since `PING` should be the most reliable way to detect the [loading state](https://github.com/redis/docs/blob/37f2a2472cfba1b7ac2d6575de81f53909a68c81/content/commands/ping/index.md?plain=1#L41).

client.go
```diff
@@ -312,7 +312,13 @@ func (c *dedicatedSingleClient) release() {
...
 
 func (c *singleClient) isRetryable(err error, ctx context.Context) bool {
-	return err != nil && err != ErrDoCacheAborted && atomic.LoadUint32(&c.stop) == 0 && ctx.Err() == nil
+	if err == nil || err == ErrDoCacheAborted || atomic.LoadUint32(&c.stop) != 0 || ctx.Err() != nil {
+		return false
+	}
+	if redisErr, ok := err.(*RedisError); ok {
+		return redisErr.IsLoading()
+	}
+	return true
```
rueian, line 321: Hi @nesty92,

Everything looks good to me, except that the existing rueidiscompat tests are not passing, even after several retries. I am afraid there are something we missed caused by the changes here.

client.go
```diff
@@ -312,7 +312,13 @@ func (c *dedicatedSingleClient) release() {
...
 }
 
 func isRetryable(err error, w wire, ctx context.Context) bool {
```
rueian, line 324: It turns out this `isRetryable` used by `dedicatedSingleClient` should also be updated.

Issue #658, opened 2024-10-31
Client degradation during node failure


justinreddit:
Hello

Been using this client for some redis clusters and noticed that it doesn't gracefully handle node failure. There are two setups we'd like to support

1. Clusters with replicas - we've observed that if the master fails, the client is unable to serve any requests.
2. Clusters without replicas - we've observed that if a node fails, the client will continue sending requests to unresponsive nodes and timing out.

Is it possible to have the client partition off unresponsive nodes so that it can continue operating?

rueian:
Hi @justinreddit,

I think the current implementation can discover the topology changes and recover from them. Could you elaborate more about your setup and what client version you use?

> Clusters with replicas - we've observed that if the master fails, the client is unable to serve any requests.

Just to confirm, your cluster setup is [this](https://redis.io/docs/latest/operate/oss_and_stack/management/scaling/), instead of [this](https://redis.io/docs/latest/operate/oss_and_stack/management/replication/#read-only-replica), right?

In the current implementation, the client is unable to serve any requests, which should only happen if all your masters fail and no replicas are promoted.

> Clusters without replicas - we've observed that if a node fails, the client will continue sending requests to unresponsive nodes and timing out.

I think this is the correct behavior: the client should simply follow the topology it discovered from the `CLUSTER SHARDS` response to send commands. You can use a shorter context deadline to avoid a long timeout.

justinreddit:
> I think this is the correct behavior: the client should simply follow the topology it discovered from the CLUSTER SHARDS response to send commands. You can use a shorter context deadline to avoid a long timeout.

Is there a way to provide a node level circuit breaker? We've had a few incidents where a node becomes unresponsive and we'd want to ignore those slots without affecting client latency.

rueian:
> Is there a way to provide a node level circuit breaker or manage the node topology directly? We've had a few incidents where a node becomes unresponsive and we'd want to ignore those slots without affecting client latency.

I understand that sometimes penetrations are desirable, but I currently don't have a good idea of the interface that would allow you to have this functionality. The only way we currently have is wrapping connections by customizing the `DialFn`.

> Edit: By setting the cluster to not failover during a node failure things work as expected.

It's good news but, however, weird. How can no failover work as expected?

> One thing I did find confusing is that in cases of nodes being unresponsive, even though we had ConnWriteTimeout set to 200ms. could you clarify the expected behavior of configuring that value?

[...]

PR #659, opened 2024-11-01
Move sleep into goroutine in `LazyDo`

justinmir:
This moves the sleep of up to `threshold` out of the caller's request path. This allows callers to use `LazyDo` without sleeping for up to `threshold` when called in succession.

This allows requests to connections that fail to return immediately instead of waiting for a second when they are selected to lazily perform the topology refresh.

Discussed here: https://github.com/redis/rueidis/issues/658

### DIFF
--- a/singleflight.go
+++ b/singleflight.go
@@ -48,8 +48,10 @@ func (c *call) LazyDo(threshold time.Duration, fn func() error) {
 	c.cn++
 	ts := c.ts
 	c.mu.Unlock()
-	time.Sleep(time.Until(ts.Add(threshold)))
-	go c.do(ch, fn)
+	go func() {
+		time.Sleep(time.Until(ts.Add(threshold)))
+		c.do(ch, fn)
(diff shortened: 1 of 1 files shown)


singleflight.go
```diff
@@ -48,8 +48,10 @@ func (c *call) LazyDo(threshold time.Duration, fn func() error) {
 	c.cn++
 	ts := c.ts
 	c.mu.Unlock()
-	time.Sleep(time.Until(ts.Add(threshold)))
-	go c.do(ch, fn)
+	go func() {
+		time.Sleep(time.Until(ts.Add(threshold)))
+		c.do(ch, fn)
+	}()
```
rueian, line 54:
```suggestion
	}(ts.Add(threshold), ch, fn)
```
It will be better to move these into function parameters to avoid heap allocations.

rueian:
Thanks @justinmir!

PR #664, opened 2024-11-07
fix: refresh cluster topology with `CLUSTER SLOTS` instead of `CLUSTER SHARDS` for clusters which version < 8 due to the server bug.

rueian:
Ref:

https://github.com/redis/rueidis/issues/384#issuecomment-2463031313

https://github.com/redis/redis/pull/13422

https://github.com/valkey-io/valkey/issues/784

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -137,7 +137,7 @@ type clusterslots struct {
 }
 
 func (s clusterslots) parse(tls bool) map[string]group {
-	if s.ver < 7 {
+	if s.ver < 8 {
 		return parseSlots(s.reply.val, s.addr)
 	}
 	return parseShards(s.reply.val, s.addr, tls)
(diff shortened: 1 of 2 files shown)


DhruvilK7:
@rueian Is there a reason for keeping the condition as `< 8` instead of saving the full version tuple and checking `< 7.2.6`? With the current approach, systems running `7.2.6` still hit the https://github.com/valkey-io/valkey-go/issues/46 while upgrading the clusters.

DhruvilK7:
@rueian Could you please take a look at this? Lowering the threshold from 8 to 7.2.6 would significantly ease engine upgrade activities across all our clusters. We are already running on 7.2.6, which supports `CLUSTER SHARDS`, so this change aligns with our current deployment.

Issue #651, opened 2024-10-25
Feature: remove idle connections from pool


rueian:
Currently connections in the connection pools (`spool` and `dpool`) inside a `mux` will live forever.

It would be nice if we had an option, such as `ConnIdleDuration` or a better naming, to clean connections that are idle for a long time to reduce memory usage.

abuzaforfagun:
I am interested in working on this issue.

proost:
@rueian 
Does need option to min size of connection pool? making a new connection is expensive. If all connections is closed, can be harmful to latency sensitive apps. current pool only support cap size.

rueian:
> @rueian Does need option to min size of connection pool? making a new connection is expensive. If all connections is closed, can be harmful to latency sensitive apps. current pool only support cap size.

Yes, there should be an option for that.

PR #674, opened 2024-11-16
Added options for configuring removal of idle connections

veirfuna:
Approach for [#651 ](https://github.com/redis/rueidis/issues/651)

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -10,7 +10,7 @@ func newPool(cap int, dead wire, idleConnTTL time.Duration, minSize int, makeFn 
 		cap = DefaultPoolSize
 	}
 
-	p := &pool{
+	return &pool{
 		size:        0,
 		minSize:     minSize,
 		cap:         cap,
@@ -20,25 +20,20 @@ func newPool(cap int, dead wire, idleConnTTL time.Duration, minSize int, makeFn 
 		cond:        sync.NewCond(&sync.Mutex{}),
 		idleConnTTL: idleConnTTL,
 	}
-
-	if idleConnTTL != 0 {
-		p.timer = time.AfterFunc(idleConnTTL, p.removeIdleConns)
-	}
-
-	return p
 }
 
 type pool struct {
-	dead        wire
-	cond        *sync.Cond
-	make        func() wire
-	list        []wire
-	size        int
-	minSize     int
-	cap         int
-	down        bool
-	idleConnTTL time.Duration
-	timer       *time.Timer
+	dead          wire
+	cond          *sync.Cond
+	make          func() wire
+	list          []wire
+	size          int
+	minSize       int
+	cap           int
+	down          bool
+	idleConnTTL   time.Duration
+	timer         *time.Timer
+	timerIsActive bool
 }
 
 func (p *pool) Acquire() (v wire) {
@@ -64,7 +59,7 @@ func (p *pool) Store(v wire) {
 	p.cond.L.Lock()
 	if !p.down && v.Error() == nil {
 		p.list = append(p.list, v)
-		p.resetTimerIfNeeded()
+		p.startTimerIfNeeded()
 	} else {
 		p.size--
 		v.Close()
@@ -84,11 +79,21 @@ func (p *pool) Close() {
 	p.cond.Broadcast()
 }
 
-func (p *pool) removeIdleConns() {
-	if p.idleConnTTL == 0 {
+func (p *pool) startTimerIfNeeded() {
+	if p.idleConnTTL == 0 || p.timerIsActive || len(p.list) <= p.minSize {
 		return
 	}
 
+	p.timerIsActive = true
+	if p.timer == nil {
+		p.timer = time.AfterFunc(p.idleConnTTL, p.removeIdleConns)
+		return
+	}
+
+	p.timer.Reset(p.idleConnTTL)
+}
+
+func (p *pool) removeIdleConns() {
 	p.cond.L.Lock()
 	defer p.cond.L.Unlock()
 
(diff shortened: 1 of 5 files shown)


pool.go
```diff
@@ -50,6 +60,11 @@ func (p *pool) Store(v wire) {
 	p.cond.L.Lock()
 	if !p.down && v.Error() == nil {
 		p.list = append(p.list, v)
+
+		if p.idleConnTTL != nil {
+			p.removeIdleConns()
```
rueian, line 65: Hi @veirfuna,

I guess calling `.removeIdleConns()` alone in the `.Store()` can't remove idle connections. For example, if `.Store()` is not called, then idle connections are still there. You probably need to put `.removeIdleConns()` into a ticker.

and thanks for your contribution!

veirfuna: Hi @rueian,

My first attempts were to create a goroutine that is observe idle connections and remove them. Do you see any problems in such solution? The only idea to not use goroutine was to add `.removeIdleConns()` to `.Store()`

rueian: I think we have no choice but to use an additional goroutine. We can use time.AfterFunc and it is actually an additional goroutine. We can schedule the cleanup function only when p.size > minIdle.

pool.go
```diff
@@ -67,3 +82,28 @@ func (p *pool) Close() {
...
+		timer, ok := p.timers[w]
+		if !ok {
+			return
+		}
+
+		select {
+		case <-timer.C:
+			w.Close()
+			p.size--
+			p.list = p.list[1:]
```
rueian, line 102: Doing this will make the underlying storage of the slice unable to be reclaimed. We should still remove connections from the tail.

pool.go
```diff
@@ -1,30 +1,47 @@
...
-	down bool
+	dead         wire
+	cond         *sync.Cond
+	make         func() wire
+	list         []wire
+	size         int
+	minSize      int
+	cap          int
+	down         bool
+	idleConnTTL  *time.Duration
```
rueian, line 41:
```suggestion
	idleConnTTL  time.Duration
```
We can just treat 0 as not enabling this feature.

pool.go
```diff
@@ -1,30 +1,47 @@
...
+	stopChan     chan struct{}
+	stopChanOnce *sync.Once
```
rueian, line 43:
```suggestion
	timer  *time.Timer
```
Can we just use `time.AfterFunc` and store the timer here? I think a timer is convenient for us because we can use `.Stop()` when `len(p.list) <= minSize` and `.Reset()` when `len(p.list) > minSize`.

pool.go
```diff
@@ -1,30 +1,47 @@
...
+	lastUsage    map[wire]time.Time
```
rueian, line 44: I am wondering if we need to keep track of idle time for each connection. Can we just clear every connection in the `p.list` that exceeds `minSize` whenever the timer is fired?

veirfuna: Sound like a good idea. Made an update in PR

pool.go
```diff
@@ -1,30 +1,44 @@
...
+		dead:        dead,
+		make:        makeFn,
+		list:        make([]wire, 0, 4),
+		cond:        sync.NewCond(&sync.Mutex{}),
+		idleConnTTL: idleConnTTL,
 	}
+
+	if idleConnTTL != 0 {
+		p.timer = time.AfterFunc(idleConnTTL, p.removeIdleConns)
+	}
```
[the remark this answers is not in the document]
veirfuna: Your idea is to start the timer only after `len(p.list) > p.minSize`?

pool.go
```diff
@@ -61,9 +76,45 @@ func (p *pool) Store(v wire) {
...
+
+	newLen := len(p.list) - min(p.size-p.minSize, len(p.list))
+	for _, w := range p.list[newLen:] {
+		w.Close()
+	}
+
+	p.list = p.list[:newLen]
+	p.size = p.minSize
+
+	p.stopTimer()
```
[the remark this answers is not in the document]
veirfuna: Could you explain why you think the timer should not be stopped inside `removeIdleConns`?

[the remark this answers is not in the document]
rueian: Isn't the timer already stopped at this point?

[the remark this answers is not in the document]
veirfuna: Yep, you are right

pool.go
```diff
@@ -61,9 +76,45 @@ func (p *pool) Store(v wire) {
...
+}
+
+func (p *pool) resetTimerIfNeeded() {
+	if p.timer != nil && p.size > p.minSize {
+		p.timer.Reset(p.idleConnTTL)
```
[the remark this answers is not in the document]
veirfuna: Agreed, added a new flag

pool.go
```diff
@@ -61,9 +71,49 @@ func (p *pool) Store(v wire) {
...
+	}
+
+	p.timer.Reset(p.idleConnTTL)
+}
+
+func (p *pool) removeIdleConns() {
+	p.cond.L.Lock()
+	defer p.cond.L.Unlock()
+
+	if p.down || p.size <= p.minSize || len(p.list) == 0 {
```
rueian, line 100:
```suggestion
	if p.down || len(p.list) <= p.minSize {
```

pool.go
```diff
@@ -61,9 +71,49 @@ func (p *pool) Store(v wire) {
...
+		return
+	}
+
+	newLen := min(p.minSize, len(p.list))
+	for _, w := range p.list[newLen:] {
+		w.Close()
+	}
```
rueian, line 107:
```suggestion
	for i, w := range p.list[newLen:] {
		w.Close()
		p.list[i] = nil
	}
```

rueian: This allows those connections to be GC.

pool.go
```diff
@@ -61,9 +71,49 @@ func (p *pool) Store(v wire) {
...
+
+	p.list = p.list[:newLen]
+	p.size = p.minSize
```
rueian, line 110:
```suggestion
```

Setting it directly to the `p.minSize` seems incorrect. We probably should put it in the above loop and use `p.size--` instead.

rueian:
Merged. Thank you for your great work @veirfuna!

PR #676, opened 2024-11-20
Feat rueidiscompat redisearch

unknowntpo:
Changes and cautions:
- docker-compose
  - add `compat-redisearch` service for testing `RediSearch` with `RESP2` 
- FT.CREATE:
  - search_command.json
	- Add options: `INDEXEMPTY`, `INDEXMISSING`, `WITHCOUNT`, `ADDSCORES`
	- Add field_type: `GEOSHAPE`
- FT.AGGREGATE:
  - search_command.json:
    - Add option: `WITHCOUNT`
- FT.SEARCH
  - go-redis doesn't implement `SUMMARIZE` option
  - Add option: `WITHCOUNT`


rueidiscompat/pipeline.go
```diff
@@ -97,6 +97,8 @@ type Pipeline struct {
 	rets []Cmder
 }
 
+var pp = Pipeline{}
```
rueian, line 100: What is this for?

rueidiscompat/adapter.go
```diff
@@ -4835,6 +4835,615 @@ func (c *Compat) Watch(ctx context.Context, fn func(Tx) error, keys ...string) e
...
+				flatArgs := []interface{}{
+					"TYPE", sc.VectorArgs.FlatOptions.Type,
+					"DIM", sc.VectorArgs.FlatOptions.Dim,
+					"DISTANCE_METRIC", sc.VectorArgs.FlatOptions.DistanceMetric,
+				}
+				if sc.VectorArgs.FlatOptions.InitialCapacity > 0 {
+					flatArgs = append(flatArgs, "INITIAL_CAP", sc.VectorArgs.FlatOptions.InitialCapacity)
+				}
+				if sc.VectorArgs.FlatOptions.BlockSize > 0 {
+					flatArgs = append(flatArgs, "BLOCK_SIZE", sc.VectorArgs.FlatOptions.BlockSize)
```
rueian, line 5132: Lint Error: adapter.go:5132:6: this value of flatArgs is never used (SA4006)

rueidiscompat/pipeline.go
unknowntpo: removed.

rueidiscompat/adapter.go
unknowntpo: fixed.

message.go
```diff
@@ -471,6 +471,16 @@ func (r RedisResult) ToAny() (v any, err error) {
...
 
+// IsMap delegates to RedisMessage.IsMap
+func (r RedisResult) IsMap() bool {
+	return r.val.IsMap()
+}
+
+// IsArray delegates to RedisMessage.IsArray
+func (r RedisResult) IsArray() bool {
+	return r.val.IsArray()
+}
```
rueian, line 482:
```suggestion
func (r RedisResult) isMap() bool {
	return r.val.IsMap()
}

// IsArray delegates to RedisMessage.IsArray
func (r RedisResult) isArray() bool {
	return r.val.IsArray()
}
```

Can we make these 2 methods private first? I am concerned about making these direct delegations to `r.val` as public APIs. I think we should at least ask users to check the `r.err` first.

rueidiscompat/util.go
```diff
@@ -0,0 +1,160 @@
...
+	t := time.NewTimer(dur)
+	defer t.Stop()
+
+	select {
+	case <-t.C:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
rueian, line 47: Can we remove unused functions in this file?

message.go
```diff
@@ -456,7 +456,7 @@ func (r RedisResult) ToMap() (v map[string]RedisMessage, err error) {
 	if r.err != nil {
 		err = r.err
 	} else {
-		v, err = r.val.ToMap()
+		v, err = r.val.AsMap()
```
rueian, line 459: Why do we need this change? I think we already have an `AsMap` delegation.

unknowntpo: This is accidentally committed, I've reverted it.

message.go
unknowntpo: But if we make `IsMap` private, functions in `command.go` can not access it,
I think we could check if any error exists, then delegate to `r.val.IsMap`.

```
func (r RedisResult) IsMap() bool {
	if r.err != nil {
		panic(fmt.Errorf("RedisResult has error: %w, please check the error before calling RedisResult.IsMap", r.err))
	}
	return r.val.IsMap()
}
```

rueidiscompat/util.go
unknowntpo: Okay.

message.go
unknowntpo: UPDATE: I'll use `res.ToMessage()` to get underlying message first, then do the `IsMap`, `IsArray` check.

rueian (approved):

PR #678, opened 2024-11-25
fix incorrect conversion between int and int64

unknowntpo:
This PR fixed incorrect conversion between int and int64 at `FTSearchResult.Result`, simply made `Result` an `int` type to avoid conversion.


unknowntpo:
@rueian Would you like to review this PR ?

PR #681, opened 2024-11-27
FEATURE rueidislimiter: Introduce RateLimitOption to RateLimiterClient

altanozlu:
Currently it's not possible to use instance of RateLimiter and share it with different use cases.

### DIFF
--- a/rueidislimiter/limit.go
+++ b/rueidislimiter/limit.go
@@ -1,0 +1,15 @@
+package rueidislimiter
+
+import "time"
+
+type RateLimitOption struct {
+	limit  int64
+	window time.Duration
+}
+
+func WithCustomRateLimit(limit int, window time.Duration) RateLimitOption {
+	return RateLimitOption{
+		limit:  int64(limit),
+		window: window,
+	}
+}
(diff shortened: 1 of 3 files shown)


rueidislimiter/limit.go
```diff
@@ -0,0 +1,21 @@
...
+	return func(rl *RateLimit) {
+		rl.limit = limit
+	}
+}
+
+func WithWindow(window time.Duration) RateLimitOption {
+	return func(rl *RateLimit) {
+		rl.window = window
+	}
+}
```
rueian, line 21:
```suggestion
type RateLimit struct {
	limit  int
	window time.Duration
}
type RateLimitOption func(RateLimit) RateLimit

func WithLimit(limit int) RateLimitOption {
	return func(rl RateLimit) RateLimit {
		rl.limit = limit
		return rl
	}
}

func WithWindow(window time.Duration) RateLimitOption {
	return func(rl RateLimit) RateLimit {
		rl.window = window
		return rl
	}
}
```

This can have better performance and fewer allocations.

rueian: This one looks good to me. Thanks @altanozlu!

PR #687, opened 2024-12-07
feat: Hide InitAddress when connecting to a redis cluster using NewClient()

### LINKED ISSUE #635: Hide InitAddress when connecting to a valkey cluster using NewClient()
See the original issue: https://github.com/valkey-io/valkey-go/issues/9

unknowntpo:
Resolves #635

What's changed:

In `(c *clusterClient) _refresh()`, when values in `ClientOption.InitAddress` is not in the return value of `getClusterSlots`, it should be hidden.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -39,6 +39,7 @@ type clusterClient struct {
 type connrole struct {
 	conn    conn
 	replica bool
+	hidden  bool
 }
 
 func newClusterClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*clusterClient, error) {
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -289,8 +292,11 @@ func (c *clusterClient) single() (conn conn) {
 func (c *clusterClient) nodes() []string {
 	c.mu.RLock()
 	nodes := make([]string, 0, len(c.conns))
-	for addr := range c.conns {
-		nodes = append(nodes, addr)
+	for addr, role := range c.conns {
+		// only display non-hidden connrole
+		if !role.hidden {
```
rueian, line 297: We don't need to change this `nodes()` method. It is a private method for other testing purposes.

We should test this new `hidden` field explicitly in a new test case for the public `Node()` method.

PR #695, opened 2024-12-12
fix: panic on cluster redirection to a node with stale role

### LINKED ISSUE #694: Observed panics during cluster topology change
While switching a cluster in AWS from redis to valkey, also modifying the topology, I observed a couple of panics like:

```
PANIC: runtime error: index out of range [423] with length 0

runtime/debug.Stack()
	/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.5.linux-amd64/src/runtime/debug/stack.go:24 +0x5e
golang.org/x/sync/singleflight.newPanicError({0x21c0a20, 0xc1983cad50})
	/go/pkg/mod/golang.org/x/sync@v0.9.0/singleflight/singleflight.go:44 +0x25
golang.org/x/sync/singleflight.(*Group).doCall.func2.1()
	/go/pkg/mod/golang.org/x/sync@v0.9.0/singleflight/singleflight.go:193 +0x34
panic({0x21c0a20?, 0xc1983cad50?})
	/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.5.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/redis/rueidis.(*clusterClient).redirectOrNew(0xc002032000, {0xc20be37d4a, 0x51}, {0x2a415e0, 0xc0024a1e00}, 0x1a7, 0x1)
	/go/pkg/mod/github.com/redis/rueidis@v1.0.49/cluster.go:446 +0x4de
github.com/redis/rueidis.(*clusterClient).do(0xc002032000, {0x2a2d000, 0xc20a6deed0}, {0xc194081a40, 0x2000, 0x1a7})
	/go/pkg/mod/github.com/redis/rueidis@v1.0.49/cluster.go:478 +0x6b1
github.com/redis/rueidis.(*clusterClient).Do(0xc036798f78?, {0x2a2d000?, 0xc20a6deed0?}, {0xc194081a40?, 0x8f20?, 0x3679?})
	/go/pkg/mod/github.com/redis/rueidis@v1.0.49/cluster.go:461 +0x7b
```

Not sure why.  Perhaps the `SHARDS` command didn't return any groups for a short while?

Possibly there could be a more graceful behavior in the client in this case?

rueian:
Fixes #694

The situation happened when a cluster node got promoted to be the primary and the client followed a `MOVED` redirection but still think the node as a replica. Then the client set the connection to the `rslots` which was not yet initialized. This PR changes two things:

1. Remove the `connrole.replica` field because, in the future, a node can have mixed roles. https://github.com/valkey-io/valkey/issues/1372
2. Trust the `MOVED` redirection. ie. It always points to the primary of the slot so the client should set the connection to `pslots` directly.

Hi @proost, could you help review this?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -37,9 +37,9 @@ type clusterClient struct {
 
 // NOTE: connrole and conn must be initialized at the same time
 type connrole struct {
-	conn    conn
-	replica bool
-	hidden  bool
+	conn   conn
+	hidden bool
+	//replica bool <- this field is removed because a server may have mixed roles at the same time in the future. https://github.com/valkey-io/valkey/issues/1372
 }
 
 func newClusterClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*clusterClient, error) {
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -397,9 +395,6 @@ func (c *clusterClient) _pick(slot uint16, toReplica bool) (p conn) {
 	c.mu.RLock()
 	if slot == cmds.InitSlot {
 		for _, cc := range c.conns {
-			if cc.replica {
```
proost, line 400: Because of This, can we remove comments `cmds.go` L14?

proost (approved):
Thank you!
1. Mixed roles is quite surprising to me.
2. We should trust `MOVED`, and also set conn `pslots` is okay because of lazy refresh.

PR #698, opened 2024-12-13
refactor: clean up unused return values

proost:
I found that last slot of command and toReplica are useless. So remove it.


rueian:
Thanks @proost!

PR #701, opened 2024-12-16
fix: cache auto retry on MOVED / ASK / ...

### LINKED ISSUE #700: Retry  `DoXCache` during redis cluster scaling
Recently we got `failed to fetch the cache because EXEC was aborted by redis or connection closed` by `DoMultiCache` during redis cluster scaling  ( relate to #696 )

We believe that fail reason should be something relation to cluster scaling. 
Check source below, seem like in TRANSACTION do commands `MULTI, PTTL, GET/MGET, EXEC`, then redis server return a error `MOVED {slot} {ip:port}` for `PTTL` command, and `EXEC` get a `EXECABORT` error. But driver only check with the `EXEC` command result.

https://github.com/redis/rueidis/blob/main/pipe.go#L1446-L1482

Seem like we can fix this easily, just handle previous error in TRANSACTION and let driver handle retry automaticlly.

We will test it ASAP and maybe file a PR later.

wyxloading:
try fix #700 

`DoCache, DoMultiCache` end out with one or some redis transaction like:
`MULTI, PTTL {key}, GET {key}, EXEC`

When we scale a redis-cluster, `PTTL {key}` may return a `MOVED / ASK` error because slot have mark migrate / migrating / migrated to another node, then the transaction may fail. 

We hope driver can handle this case automaticlly by just retry the transaction on node which error said.

EDIT:
We tested in our real test environment and it works.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -828,20 +828,17 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults 
 }
 
 func askingMultiCache(cc conn, ctx context.Context, multi []CacheableTTL) *redisresults {
-	commands := make([]Completed, 0, len(multi)*7)
+	commands := make([]Completed, 0, len(multi)*6)
 	for _, cmd := range multi {
 		ck, _ := cmds.CacheKey(cmd.Cmd)
-		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), cmds.AskingCmd, Completed(cmd.Cmd), cmds.ExecCmd)
+		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), Completed(cmd.Cmd), cmds.ExecCmd)
 	}
 	results := resultsp.Get(0, len(multi))
 	resps := cc.DoMulti(ctx, commands...)
-	for i := 6; i < len(resps.s); i += 7 {
+	for i := 5; i < len(resps.s); i += 6 {
 		if arr, err := resps.s[i].ToArray(); err != nil {
-			for _, j := range []int{i - 3, i - 1} { // check if PTTL or {Cmd} failed
-				if preErr := resps.s[j].Error(); preErr != nil {
-					err = preErr
-					break
-				}
+			if preErr := resps.s[i-1].Error(); preErr != nil { // if {Cmd} get a RedisError
+				err = preErr
 			}
 			results.s = append(results.s, newErrResult(err))
 		} else {
--- a/cluster.go
+++ b/cluster.go
@@ -828,15 +828,21 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults 
 }
 
 func askingMultiCache(cc conn, ctx context.Context, multi []CacheableTTL) *redisresults {
-	commands := make([]Completed, 0, len(multi)*6)
+	commands := make([]Completed, 0, len(multi)*7)
 	for _, cmd := range multi {
 		ck, _ := cmds.CacheKey(cmd.Cmd)
-		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), Completed(cmd.Cmd), cmds.ExecCmd)
+		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), cmds.AskingCmd, Completed(cmd.Cmd), cmds.ExecCmd)
 	}
 	results := resultsp.Get(0, len(multi))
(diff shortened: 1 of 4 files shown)


rueian:
> BTW, happened to find out testcase `TestClusterClient_SendToOnlyPrimaryNodeWhenPrimaryNodeSelected` `Do with no slot` may fail sometimes. Seems like driver will pick any conn to do initSlot commands, but testcase assume it only use primary node conn

Thank you for catching that. We can just delete the test case because we don't guarantee using primaries for those commands. In the future, nodes may have mixed roles of primaries and replicas.

cluster.go
```diff
@@ -828,15 +828,21 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults
 }
 
 func askingMultiCache(cc conn, ctx context.Context, multi []CacheableTTL) *redisresults {
-	commands := make([]Completed, 0, len(multi)*6)
+	commands := make([]Completed, 0, len(multi)*7)
 	for _, cmd := range multi {
 		ck, _ := cmds.CacheKey(cmd.Cmd)
-		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), Completed(cmd.Cmd), cmds.ExecCmd)
+		commands = append(commands, cmds.OptInCmd, cmds.MultiCmd, cmds.AskingCmd, cmds.NewCompleted([]string{"PTTL", ck}), cmds.AskingCmd, Completed(cmd.Cmd), cmds.ExecCmd)
```
rueian, line 834: Is the second `cmds.AskingCmd` really needed? It looks like we only need one:
image
https://redis.io/docs/latest/commands/asking/

cluster.go
```diff
@@ -828,15 +828,21 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults
...
 	}
 	results := resultsp.Get(0, len(multi))
 	resps := cc.DoMulti(ctx, commands...)
-	for i := 5; i < len(resps.s); i += 6 {
+	for i := 6; i < len(resps.s); i += 7 {
 		if arr, err := resps.s[i].ToArray(); err != nil {
+			for _, j := range []int{i - 3, i - 1} { // check if PTTL or {Cmd} failed
```
rueian, line 840: Could we check only the response of cmd?

pipe.go
```diff
@@ -1315,6 +1315,11 @@ func (p *pipe) DoCache(ctx context.Context, cmd Cacheable, ttl time.Duration) Re
 	if err != nil {
 		if _, ok := err.(*RedisError); ok {
 			err = ErrDoCacheAborted
+			if preErr := resp.s[2].Error(); preErr != nil { // if PTTL command get a RedisError
```
rueian, line 1318: Shouldn't we better check the response of cmd instead? It is possible that the PTTL command queued successfully but the following cmd command failed to queue in the transaction with MOVED.

wyxloading: PTTL embeded in a TRANSACTION, i think checking the PTTL command result is enough though (but not so sure). Maybe we need to check both command results?

cluster.go
wyxloading: I would run tests to find out.

pipe.go
wyxloading: We check the redis response while we are doing cluster scaling (using tcpdump), and we find out the MOVED / ASK error comes twice every time. Which means that either checking PTTL  or {cmd} result is enough. 

Are there any chance that server will response with only one MOVED / ASK error ?

rueian: > PTTL embeded in a TRANSACTION

Actually, the transaction hasn't started at the moment the server receives the PTTL command from the client. The atomicity only happens when the server executes the `EXEC` command.

> Are there any chance that server will response with only one MOVED / ASK error ?

So, yes, I think there is a chance that PTTL gets queued successfully, but the command we want fails if slot migration happens between them.

cluster.go
wyxloading: tested and yes, only one ASKING is needed. Thank you for pointing out.

cluster.go
wyxloading: Yes. After rethinking this code last night, find out we only need to check the response of the {cmd}, cause it's the last command queued, and it cover all keys we needed.

pipe.go
wyxloading: Get it. Thank you for pointing out

wyxloading:
@rueian all review discussions resolved, please check this out.

cluster.go
```diff
@@ -837,6 +837,9 @@ func askingMultiCache(cc conn, ctx context.Context, multi []CacheableTTL) *redis
 	resps := cc.DoMulti(ctx, commands...)
 	for i := 5; i < len(resps.s); i += 6 {
 		if arr, err := resps.s[i].ToArray(); err != nil {
+			if preErr := resps.s[i-1].Error(); preErr != nil { // if {Cmd} get a RedisError
+				err = preErr
```
rueian, line 841: Can we have a test case for this?

wyxloading: Done

PR #709, opened 2024-12-22
fix: broken tx retries for cluster clients after #697

rueian:


cluster.go
```diff
@@ -643,9 +646,18 @@ func (c *clusterClient) pickMulti(ctx context.Context, multi []Completed) (*conn
...
+func isMulti(cmd Completed) bool {
+	return len(cmd.Commands()) == 1 && cmd.Commands()[0] == "MULTI"
+}
+func isExec(cmd Completed) bool {
+	return len(cmd.Commands()) == 1 && cmd.Commands()[0] == "EXEC"
+}
+
 func (c *clusterClient) doresultfn(
-	ctx context.Context, results *redisresults, retries *connretry, mu *sync.Mutex, cc conn, cIndexes []int, commands []Completed, resps []RedisResult, attempts int,
+	ctx context.Context, results *redisresults, retries *connretry, mu *sync.Mutex, cc conn, cIndexes []int, commands []Completed, resps []RedisResult, attempts int, single bool,
```
rueian, line 657: `single` is a hint that the original DoMulti is only involved with 1 connection. We use this hint to skip unnecessary tx checks.

cluster.go
```diff
@@ -589,25 +589,28 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry) {
...
 			panic(panicMixCxSlot)
 		}
-		p := c.pslots[cmd.Slot()]
-		if p == nil {
+		cc := c.pslots[cmd.Slot()]
+		if cc == nil {
 			return nil
 		}
-		count.m[p]++
+		count.m[cc]++
```
rueian, line 596: make naming consistent.

cluster.go
```diff
@@ -589,25 +589,28 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry) {
...
+				count.m[cc] = inits
 				break
 			}
 		}
 		if last == cmds.InitSlot {
 			return nil
 		}
+	} else if init {
+		cc := c.pslots[last]
+		count.m[cc] += inits
```
rueian, line 613: We don't count cmd.InitSlot commands to `c.pslots[last]` in the code above this, so add the count back here.

wyxloading:
Find one extreme case:
```
MULTI
MULTI
GET foo
EXEC
```

We will  get [4]RedisResult with all successful results. And resp[3] (which means `EXEC` result) will get a len == 3 array, that's not  accurate right. The second `MULTI` should get a error `ERR MULTI calls can not be nested` and the transaction can `EXEC` though.

wyxloading:
>  I think we shouldn't retry the transaction if the nearest MULTI command fails with a Redis error.

make sense though. 

BTW i got different response from redis:7.4 cluster

cmd:
```
MULTI
MULTI
PTTL foo
EXEC
```
response
```
OK
ERR MULTI calls can not be nested
QUEUED
1) (integer) -2
```

cluster.go
```diff
@@ -664,6 +675,37 @@ func (c *clusterClient) doresultfn(
 			} else {
 				nc = c.redirectOrNew(addr, cc, cm.Slot(), mode)
 			}
+			if hasInit && ei < i { // find out if there is a transaction block or not.
+				for mi = i; mi >= 0 && !isMulti(commands[mi]) && !isExec(commands[mi]); mi-- {
+				}
+				for ei = i; ei < len(commands) && !isMulti(commands[ei]) && !isExec(commands[ei]); ei++ {
+				}
```
rueian, line 682: Both `mi` and `ei` cursor will stop at either `MULTI` or `EXEC` to avoid crossing tx boundaries.

cluster.go
```diff
@@ -664,6 +675,37 @@ func (c *clusterClient) doresultfn(
...
+				if mi >= 0 && mi < ei && ei < len(commands) && isMulti(commands[mi]) && isExec(commands[ei]) && resps[mi].val.string == "QUEUED" { // a transaction is found.
```
rueian, line 683: If the nearest `MULTI` doesn't succeed, we don't retry the tx.

rueian:
> > I think we shouldn't retry the transaction if the nearest MULTI command fails with a Redis error.
> 
> make sense though.
> 
> BTW i got different response from redis:7.4 cluster
> 
> cmd:
> 
> ```
> MULTI
> MULTI
> PTTL foo
> EXEC
> ```
> 
> response
> 
> ```
> OK
> ERR MULTI calls can not be nested
> QUEUED
> 1) (integer) -2
> ```
> 
> EDIT: If we sent the second `MULTI` command after `PTTL foo`, `EXEC` with same result. If key `foo` belong to slot in state migrating, it will get `ASK` error, seems like it will fit in the code that find out a transaction and retry logic

I tested these cases on Valkey 8, and they all failed with EXECABORT. But anyway, I think we should just not retry on these wired cases.

wyxloading:
LGTM mostly, but i have to test this PR again tomorrow

cluster.go
```diff
@@ -664,6 +675,37 @@ func (c *clusterClient) doresultfn(
 			} else {
 				nc = c.redirectOrNew(addr, cc, cm.Slot(), mode)
 			}
+			if hasInit && ei < i { // find out if there is a transaction block or not.
+				for mi = i; mi >= 0 && !isMulti(commands[mi]) && !isExec(commands[mi]); mi-- {
+				}
+				for ei = i; ei < len(commands) && !isMulti(commands[ei]) && !isExec(commands[ei]); ei++ {
+				}
+				if mi >= 0 && mi < ei && ei < len(commands) && isMulti(commands[mi]) && isExec(commands[ei]) && resps[mi].val.string == ok { // a transaction is found.
```
rueian, line 683: If the nearest `MULTI` didn't succeed, we don't retry the tx.

cluster.go
```diff
@@ -664,6 +675,37 @@ func (c *clusterClient) doresultfn(
...
+						} else {
+							nr.cIndexes = append(nr.cIndexes, ii)
+							nr.commands = append(nr.commands, cm)
+						}
+					}
+					mu.Unlock()
+					continue // the transaction has been added to the retries, go to the next cmd.
+				}
+			}
+			if hasInit && (mi < i && i < ei && mi >= 0 && isMulti(commands[mi]) || i > ei && ei >= 0 && isMulti(commands[ei])) {
```
wyxloading, line 706: `isMulti(commands[ei])` maybe a typo ? should be `isExec(commands[ei])`

cluster.go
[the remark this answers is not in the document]
rueian: Yes, it should be `OK`.

cluster.go
rueian: It is not a typo. If `commands[ei]` is a `MULTI` then we should not retry the command on `i`. However, I found that it is actually a dead code, `i` will never be larger than `ei` at this point, so the condition is removed now.

wyxloading (approved):
LGTM :+1:

wyxloading:
IMO we should publish a release for it.

rueian:
> IMO we should publish a release for it.

Yeap, but we just forget handling ASKING correctly. I have made a new commit for it.

PR #715, opened 2025-01-06
feat: implement Sliding Window Bloom Filter

nesty92:
# Sliding Window Bloom Filter

### Core Concept
 - Contains two Bloom filters: current and next
 - Elements are added to both filters. 
 - At the half of the window the filters are rotated current is dropped, next becomes current
this is done in Redis to warranty an atomic operation.

### Redis implementation.
1. Key structure:
    - '{name}' is the current filter.
    -  '{name:c}' is the current counter.
    - '{name}:n' is the next filter
    - '{name}:nc' is the next counter.
    - '{name}:lr' is the last rotation lock to avoid multiples rotations in a given period

2. Scripts:
   - Add: Sets bits in the current and next filters.
   - Exists: Checks just bits in the current filter.
   - Rotate: swaps the next filter with the current one, creating a new next filter.

3. Rotation Coordination:
   - Clients track rotation timing locally and just trigger the rotation that will only be executed if the last rotation lock is not present

### DIFF
--- a/rueidisprob/README.md
+++ b/rueidisprob/README.md
@@ -134,3 +134,62 @@ func main() {
     fmt.Println(count) // 1
 }
 ```
+
+### Sliding Window Bloom Filter
+
+It is a variation of the standard Bloom filter that adds a sliding window mechanism. 
+Useful for use cases where you need to keep track of items for a certain amount of time.
+
+Example:
+
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/redis/rueidis"
+	"github.com/redis/rueidis/rueidisprob"
+)
+
+func main() {
+	client, err := rueidis.NewClient(rueidis.ClientOption{
+		InitAddress: []string{"localhost:6379"},
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	sbf, err := NewSlidingBloomFilter(client, "sliding_bloom_filter", 1000, 0.01, time.Minute)
+
+	err = sbf.Add(context.Background(), "hello")
+	if err != nil {
+		panic(err)
+	}
+
+	err = sbf.Add(context.Background(), "world")
+	if err != nil {
+		panic(err)
+	}
+
+	exists, err := sbf.Exists(context.Background(), "hello")
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(exists) // true
+
+	exists, err = sbf.Exists(context.Background(), "world")
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(exists) // true
+
+	count, err := sbf.Count(context.Background())
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(count) // 2
+}
+```
--- a/rueidisprob/slidingbloomfilter.go
+++ b/rueidisprob/slidingbloomfilter.go
@@ -1,0 +1,446 @@
+package rueidisprob
+
+import (
+	"context"
+	"errors"
+	"strconv"
+	"time"
+
+	"github.com/redis/rueidis"
+)
+
+const (
+	slidingBloomFilterInitializeScript = `
+local filterKey = KEYS[1]
+local nextFilterKey = KEYS[2]
+local counterKey = KEYS[3]
+local nextCounterKey = KEYS[4]
+local lastRotationKey = KEYS[5]
+local windowHalf = tonumber(ARGV[1])
+
+if redis.call('EXISTS', filterKey, nextFilterKey, counterKey, nextCounterKey, lastRotationKey) == 0 then
+	local time = redis.call('TIME')
+	local current_time = tonumber(time[1]) * 1000 + math.floor(tonumber(time[2]) / 1000)
+
+	redis.call('MSET', filterKey, "", counterKey, 0, nextFilterKey, "", nextCounterKey, 0)
+	redis.call('SET', lastRotationKey, tostring(current_time), 'PX', windowHalf, 'NX')
+end
+
+return 1
+`
+
+	slidingBloomFilterAddMultiScript = `
+local hashIterations = tonumber(ARGV[1])
+local windowHalf = tonumber(ARGV[2])
+local numElements = tonumber(#ARGV) - 2
+
+local filterKey = KEYS[1]
+local nextFilterKey = KEYS[2]
+local counterKey = KEYS[3]
+local nextCounterKey = KEYS[4]
+local lastRotationKey = KEYS[5]
+
+local time = redis.call('TIME')
+local current_time = tonumber(time[1]) * 1000 + math.floor(tonumber(time[2])/1000)
+local acquiredLock = redis.call('SET', lastRotationKey, tostring(current_time), 'PX', windowHalf, 'NX')
+
+if acquiredLock then
+	redis.call('RENAME', nextFilterKey, filterKey)
+	redis.call('RENAME', nextCounterKey, counterKey)
+	redis.call('SET', nextFilterKey, "")
+	redis.call('SET', nextCounterKey, 0)
+end
+
+local counter = 0
+local oneBits = 0
+for i=1, numElements do
+	local bitset = redis.call('BITFIELD', filterKey, 'SET', 'u1', ARGV[i+2], '1')
+	redis.call('BITFIELD', nextFilterKey, 'SET', 'u1', ARGV[i+2], '1')
+
+	oneBits = oneBits + bitset[1]
+	if i % hashIterations == 0 then
+		if oneBits ~= hashIterations then
+			counter = counter + 1
+		end
+
+		oneBits = 0
+	end
+end
+
+redis.call('INCRBY', nextCounterKey, counter)
+return redis.call('INCRBY', counterKey, counter)
+`
+	slidingBloomFilterExistsMultiScript = `
+local hashIterations = tonumber(ARGV[1])
+local windowHalf = tonumber(ARGV[2])
+local numElements = tonumber(#ARGV) - 2
+
+local filterKey = KEYS[1]
+local nextFilterKey = KEYS[2]
+local counterKey = KEYS[3]
+local nextCounterKey = KEYS[4]
+local lastRotationKey = KEYS[5]
+
+local time = redis.call('TIME')
+local current_time = tonumber(time[1]) * 1000 + math.floor(tonumber(time[2])/1000)
+local acquiredLock = redis.call('SET', lastRotationKey, tostring(current_time), 'PX', windowHalf, 'NX')
+
+if acquiredLock then
+	redis.call('RENAME', nextFilterKey, filterKey)
+	redis.call('RENAME', nextCounterKey, counterKey)
+	redis.call('SET', nextFilterKey, "")
+	redis.call('SET', nextCounterKey, 0)
+end
+
+local result = {}
+local oneBits = 0
+for i=1, numElements do
+	local index = tonumber(ARGV[i+2])
+	local bitset = redis.call('BITFIELD', filterKey, 'GET', 'u1', index)
+
+	oneBits = oneBits + bitset[1]
+	if i % hashIterations == 0 then
+		table.insert(result, oneBits == hashIterations)
+
+		oneBits = 0
+	end
+end
+
+return result
+`
+
+	slidingBloomFilterExistsReadOnlyMultiScript = `
+local hashIterations = tonumber(ARGV[1])
+local windowHalf = tonumber(ARGV[2])
+local numElements = tonumber(#ARGV) - 2
+
+local filterKey = KEYS[1]
+local nextFilterKey = KEYS[2]
+local counterKey = KEYS[3]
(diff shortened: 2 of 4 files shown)


nesty92:
Hi @rueian, I was working on this and maybe is a good plus to the lib. Any feedback is more than welcome

I had to borrow some of the internal utilities for this module, but I'm not sure if that's the best way to do it.

I made some improvements for the indexes that could also be applied to existing filters, it won't improve that much the execution time because the slowest part is waiting for Redis, but it will reduce the GC pressure when the filter is on a "hot path".

rueian:
Hi @nesty92, thank you. I will look into it later. @proost would you like to take a look as well?

rueidisprob/README.md
```diff
@@ -134,3 +134,62 @@ func main() {
...
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/redis/rueidis"
+    "github.com/redis/rueidis/rueidisprob"
```
rueian, line 154: The indentions here seem inconsistent.

rueidisprob/slidingbloomfilter_retry.go
```diff
@@ -0,0 +1,37 @@
...
+	return
+}
+
+// Default backoff implementation
+func defaultBackoff(attempts int, maxRetries int) (time.Duration, bool) {
+	if attempts >= maxRetries {
+		return 0, false
+	}
+	base := 1 << min(maxRetries, attempts)
+	jitter := fastRand(base)
```
rueian, line 35: I guess you can just use the `util.FastRand` from the `internal/util` package and no need to have a separated slidingbloomfilter_retry.1.22.go

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,443 @@
...
+		s.rotationMu.Unlock()
+		return nil
+	}
+	s.lastRotationAttempt = now
+	s.rotationMu.Unlock()
+
+	s.attempts = 0
+	for {
+		resp := s.rotateScript.Exec(ctx, s.client, s.rotateKeys, s.rotateArgs)
+		if resp.Error() == nil {
```
proost, line 431: doesn't we need to check rotation is success?

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,443 @@
...
+			return nil
+		}
+
+		sleep, retry := s.backoff(s.attempts, s.maxRetries)
+		if !retry || sleep >= s.rotation {
+			return fmt.Errorf("failed to rotate sliding bloom filter: %w", resp.Error())
+		}
+
+		s.attempts++
```
proost, line 440: `attempts` is not thread safe. If multiple goroutines can add 1 concurrently, before sleeping.

proost (commented):
Clever Idea!

rueidisprob/slidingbloomfilter.go
nesty92: To avoid having to block until the rotation is done, we set the timer even before the rotation is done so that other calls will return, leaving only one that will retry the rotation until it finishes or reaches max, skipping it to the window and that could be a problem... maybe an additional goroutine that maintains the rotation would be better?, WDYT @proost @rueian ?

proost: Extreme edge case:
If `rotation` is short(e.g. 500ms), there are multiple goroutines and timeout in context bit long and executing script might long,

1. race condition happen in `attempts`.
2. no error but "0" returned.

I think:
1. appliying single flight
2. make L428 to L441 to critical section
Or something other choice are ok.

proost: As you said, leaving only one is critical.

proost: Let me clarify this:

rotation is important. but i think trying rotation less in each application server can be applied to here even extreme case.

I think lazy rotation is enough.

nesty92: It's not a problem if multiples goroutines try to rotate in Redis (but it should not happen) the lock will avoid rotating several times in those cases. We could track the last rotation and set it back in case of failures, meaning that the next call will try to rotate again.

The main idea is to don't block on the rotation path several calls, they are safe to go to Redis

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,438 @@
...
+		resp := s.rotateScript.Exec(ctx, s.client, s.rotateKeys, s.rotateArgs)
+		if resp.Error() == nil {
+			return nil
+		}
+
+		sleep, retry := s.backoff(s.attempts, s.maxRetries)
+		if !retry || sleep >= s.rotation {
+			s.rotationMu.Lock()
+			s.lastRotationAttempt = oldRotation
+			s.rotationMu.Unlock()
```
nesty92, line 431: I refer to this @proost, this will avoid the case of spiking the rotation until the next rotation, it will be tried again on the next call to Exists or Add

rueidisprob/slidingbloomfilter.go
proost: OK, I understand your intention. Then it skip it.

proost (approved):
LGTM.

I just wandering: 

there are partitioned and discrete bloom filter idea for unbounded input. your choice overlapping and sliding window bloom filter, do you think partitioned and discrete bloom filter is too expensive?

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		return 0, err
+	}
+
+	return count, nil
+}
+
+func (s *slidingBloomFilter) checkAndRotate(ctx context.Context) error {
+	// Fast path - check if enough time has passed without locking
```
rueian, line 404: Accessing without locking is incorrect. You still need to use lock here, otherwise it is still data race.

nesty92: That don't sound bad,  I’ll explore that idea

nesty92:
Hi @rueian @proost, Following the idea of moving the rotation to redis, i found that some times we get a weird behavior with the expirations.

In the test case the PX is 500 but when running it on multiples test  Redis says that the pttl is ~6s, see the video


https://github.com/user-attachments/assets/49c1a64e-13fe-4534-8c18-6e1e69f3f801

 ```
 1736337047.130384 [0 lua] "SET" "{test}:lr" "1736337047130" "PX" "500" "NX"
1736337047.633866 [0 lua] "SET" "{test}:lr" "1736337047633" "PX" "500" "NX"
1736337048.134485 [0 lua] "SET" "{test}:lr" "1736337048134" "PX" "500" "NX"
1736337048.145867 [0 lua] "SET" "{test}:lr" "1736337048145" "PX" "500" "NX"
1736337054.405496 [0 lua] "SET" "{test}:lr" "1736337054405" "PX" "500" "NX"
1736337059.410190 [0 lua] "SET" "{test}:lr" "1736337059410" "PX" "500" "NX"
1736337057.153281 [0 lua] "SET" "{test}:lr" "1736337057153" "PX" "500" "NX"
1736337057.657655 [0 lua] "SET" "{test}:lr" "1736337057657" "PX" "500" "NX"
1736337058.161746 [0 lua] "SET" "{test}:lr" "1736337058161" "PX" "500" "NX"
1736337058.171059 [0 lua] "SET" "{test}:lr" "1736337058170" "PX" "500" "NX"
1736337064.421463 [0 lua] "SET" "{test}:lr" "1736337064421" "PX" "500" "NX"
1736337069.431727 [0 lua] "SET" "{test}:lr" "1736337069431" "PX" "500" "NX"
1736337069.448439 [0 lua] "SET" "{test}:lr" "1736337069448" "PX" "500" "NX"
1736337074.444494 [0 lua] "SET" "{test}:lr" "1736337074444" "PX" "500" "NX"
1736337072.370417 [0 lua] "SET" "{test}:lr" "1736337072370" "PX" "500" "NX"
1736337072.870308 [0 lua] "SET" "{test}:lr" "1736337072870" "PX" "500" "NX"
1736337072.872326 [0 lua] "SET" "{test}:lr" "1736337072872" "PX" "500" "NX"
1736337079.452635 [0 lua] "SET" "{test}:lr" "1736337079452" "PX" "500" "NX"
1736337079.467039 [0 lua] "SET" "{test}:lr" "1736337079466" "PX" "500" "NX"
1736337084.460596 [0 lua] "SET" "{test}:lr" "1736337084460" "PX" "500" "NX"
1736337082.378464 [0 lua] "SET" "{test}:lr" "1736337082378" "PX" "500" "NX"
1736337082.878354 [0 lua] "SET" "{test}:lr" "1736337082878" "PX" "500" "NX"
1736337082.881565 [0 lua] "SET" "{test}:lr" "1736337082881" "PX" "500" "NX"
1736337089.463478 [0 lua] "SET" "{test}:lr" "1736337089463" "PX" "500" "NX"
1736337089.477189 [0 lua] "SET" "{test}:lr" "1736337089477" "PX" "500" "NX"
1736337094.470253 [0 lua] "SET" "{test}:lr" "1736337094470" "PX" "500" "NX"
1736337092.386336 [0 lua] "SET" "{test}:lr" "1736337092386" "PX" "500" "NX"
1736337092.886976 [0 lua] "SET" "{test}:lr" "1736337092886" "PX" "500" "NX"
1736337092.887048 [0 lua] "SET" "{test}:lr" "1736337092887" "PX" "500" "NX"
1736337099.481321 [0 lua] "SET" "{test}:lr" "1736337099481" "PX" "500" "NX"
```

Do you have any idea why this happens?

**Review by proost — COMMENTED** [review]
Hmm, that is very weird operation. I read code again, but i don't know how can be happen.

File: rueidisprob/slidingbloomfilter.go [hunk 7]
```diff
@@ -0,0 +1,460 @@
...
+local lastRotationKey = KEYS[5]
+
+local time = redis.call('TIME')
+local current_time = tonumber(time[1]) * 1000 + math.floor(tonumber(time[2])/1000)
+local last_rotation = redis.call('GET', lastRotationKey)
+
+if not last_rotation or (current_time - tonumber(last_rotation)) >= windowHalf then
+	if redis.call('SET', lastRotationKey, tostring(current_time), 'PX', windowHalf, 'NX') then
+		redis.call('DEL', filterKey)
+		redis.call('RENAME', nextFilterKey, filterKey)
```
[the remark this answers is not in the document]
nesty92: nice one

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,448 @@
...
+return result
+`
+
+	slidingBloomFilterResetScript = `
+local filterKey = KEYS[1]
+local nextFilterKey = KEYS[2]
+local counterKey = KEYS[3]
+local nextCounterKey = KEYS[4]
+
+redis.call('DEL', filterKey)
```
proost, line 157: we can remove "DEL" command here too.

proost (commented):
I found that code is changed. is weird pttl phenomenon gone? i guess same situation happen.

nesty92:
> I found that code is changed. is weird pttl phenomenon gone? i guess same situation happen.

@proost, It continue from time to time, even try with redis-benchmark to discard any possible error from our side and still, I'll prepare a issue on Redis to get more info about it

nesty92:
> issue on Redis to get more info about it

https://github.com/redis/redis/issues/13733

rueian:
> Do you have any idea why this happens?

Hi @nesty92, I can't reproduce the issue with valkey 8. These are the commands I used:

```sh
watch -n0 ./valkey-cli PTTL a
watch -n0 ./valkey-cli SET a 1 PX 500 NX
```

Which version of Redis are you using? But, anyway, I think we can merge this already.

nesty92:
Hi @rueian , I was using 7.4 that is the same in the cluster used for testing. Not sure yet why this behavior but there's nothing to do from our side

rueidisprob/slidingbloomfilter.go
```diff
@@ -0,0 +1,446 @@
...
+local lastRotationKey = KEYS[5]
+
+local time = redis.call('TIME')
+local current_time = tonumber(time[1]) * 1000 + math.floor(tonumber(time[2])/1000)
+local acquiredLock = redis.call('SET', lastRotationKey, tostring(current_time), 'PX', windowHalf, 'NX')
+
+if acquiredLock then
+	redis.call('RENAME', nextFilterKey, filterKey)
+	redis.call('RENAME', nextCounterKey, counterKey)
+	redis.call('SET', nextFilterKey, "")
```
rueian, line 50: Here the script always rotate the filters once if it acquires the lock. But shouldn’t it rotate the filters twice if we access it after idling for a while (> 2*windowHalf)?

nesty92: You're absolutely correct. If more than half of the window has passed since the last rotation, it means that both the current and next filters are no longer valid since they no longer contain relevant data. In this case, saving the last rotation and reseting the filter is to much time as passed will solve it

PR #721, opened 2025-01-12
feat: add AZ to replica info

proost:
previous discussion: https://github.com/redis/rueidis/discussions/710

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -207,15 +207,26 @@ func (c *clusterClient) _refresh() (err error) {
 
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
-	for master, g := range groups {
-		conns[master] = connrole{conn: c.connFn(master, c.opt)}
+	for primary, g := range groups {
+		primaryConn := c.connFn(primary, c.opt)
+		conns[primary] = connrole{conn: primaryConn}
+		if len(g.nodes) > 0 {
+			g.nodes[0].AZ = primaryConn.AZ()
+		}
+
 		if c.rOpt != nil {
-			for _, nodeInfo := range g.nodes[1:] {
-				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.rOpt)}
+			for i, nodeInfo := range g.nodes[1:] {
+				replicaConn := c.connFn(nodeInfo.Addr, c.rOpt)
+				conns[nodeInfo.Addr] = connrole{conn: replicaConn}
+
+				g.nodes[i+1].AZ = replicaConn.AZ()
 			}
 		} else {
-			for _, nodeInfo := range g.nodes[1:] {
-				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.opt)}
+			for i, nodeInfo := range g.nodes[1:] {
+				replicaConn := c.connFn(nodeInfo.Addr, c.opt)
+				conns[nodeInfo.Addr] = connrole{conn: replicaConn}
+
+				g.nodes[i+1].AZ = replicaConn.AZ()
 			}
 		}
 	}
--- a/cluster.go
+++ b/cluster.go
@@ -265,7 +265,7 @@ func (c *clusterClient) _refresh() (err error) {
 						mu sync.Mutex
 						wg sync.WaitGroup
 					)
-					for i := 0; i < n; i += 4 {
+					for i := 0; i < n; i += 4 { // batch 4 nodes for each goroutine
 						for j := i; j < i+4 && j < n; j++ {
 							replica := g.nodes[j+1]
 							rConn := conns[replica.Addr].conn
--- a/cluster.go
+++ b/cluster.go
@@ -207,26 +207,15 @@ func (c *clusterClient) _refresh() (err error) {
 
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
-	for primary, g := range groups {
-		primaryConn := c.connFn(primary, c.opt)
-		conns[primary] = connrole{conn: primaryConn}
-		if len(g.nodes) > 0 {
-			g.nodes[0].AZ = primaryConn.AZ()
-		}
-
+	for master, g := range groups {
+		conns[master] = connrole{conn: c.connFn(master, c.opt)}
 		if c.rOpt != nil {
-			for i, nodeInfo := range g.nodes[1:] {
-				replicaConn := c.connFn(nodeInfo.Addr, c.rOpt)
-				conns[nodeInfo.Addr] = connrole{conn: replicaConn}
-
-				g.nodes[i+1].AZ = replicaConn.AZ()
+			for _, nodeInfo := range g.nodes[1:] {
+				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.rOpt)}
 			}
 		} else {
-			for i, nodeInfo := range g.nodes[1:] {
-				replicaConn := c.connFn(nodeInfo.Addr, c.opt)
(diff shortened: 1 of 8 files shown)


pipe.go
```diff
@@ -73,6 +74,7 @@ type pipe struct {
 	nsubs           *subs                       // pubsub  message subscriptions
 	psubs           *subs                       // pubsub pmessage subscriptions
 	info            map[string]RedisMessage
+	az              string
```
rueian, line 77: Can we remove this? Just relying on the `info` is enough.

pipe.go
```diff
@@ -248,13 +252,19 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps, nobg
...
 		p.onInvalidations = option.OnInvalidations
 	} else {
 		if !option.DisableCache {
 			p.Close()
 			return nil, ErrNoCache
 		}
 		init = init[:0]
+		if !isServerVersionLowerThan6 {
+			init = append(init, []string{"HELLO"})
+		}
```
rueian, line 267: Can we remove `isServerVersionLowerThan6` and just always send `HELLO 2` anyway?

cluster.go
```diff
@@ -207,15 +207,26 @@ func (c *clusterClient) _refresh() (err error) {
 
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
-	for master, g := range groups {
-		conns[master] = connrole{conn: c.connFn(master, c.opt)}
+	for primary, g := range groups {
+		primaryConn := c.connFn(primary, c.opt)
+		conns[primary] = connrole{conn: primaryConn}
+		if len(g.nodes) > 0 {
+			g.nodes[0].AZ = primaryConn.AZ()
```
rueian, line 214: I think the AZ() is not yet available right after `c.connFn`. `c.connFn` only constructs the `*mux` without making the actual TCP connection. We may need another way to do this.

rueian: Oh, I just saw you implemented AZ() as `m.pipe(0).AZ()` then it would make an actual connection to the server.
But here is still not a good place to call AZ(). It will be better we call AZ() only when the AZ_AFFINITY is enable and call it here https://github.com/redis/rueidis/blob/09ad427372a36519b3ef64c6678d755e749b51ef/cluster.go#L265 with batching.

pipe.go
proost: [908bc94](https://github.com/redis/rueidis/pull/721/commits/908bc940f1ea3372e4e2236d8426184d2f032163)

pipe.go
proost: [908bc94](https://github.com/redis/rueidis/pull/721/commits/908bc940f1ea3372e4e2236d8426184d2f032163)

cluster.go
proost: So, you mean not always set AZ to `ReplicaInfo` but if user want to know, then set AZ to `ReplicaInfo`?

rueian: I mean the `AZ()` should be invoked only when necessary, which is when the AZ_AFFINITY is configured.

pipe.go
```diff
@@ -255,6 +256,7 @@ func _newPipe(connFn func() (net.Conn, error), option *ClientOption, r2ps, nobg
 			return nil, ErrNoCache
 		}
 		init = init[:0]
+		init = append(init, []string{"HELLO"})
```
rueian, line 259:
```suggestion
		init = append(init, []string{"HELLO", "2"})
```

cluster.go
rueian: And, the current calling point, which is right after `result.parse`, is not a good place to call `AZ()`. It will always create new connections. We should instead call it after:
https://github.com/redis/rueidis/blob/09ad427372a36519b3ef64c6678d755e749b51ef/cluster.go#L232-L243

So that we can reuse existing connections as much as possible.

pipe.go
proost: [908bc94](https://github.com/redis/rueidis/pull/721/commits/908bc940f1ea3372e4e2236d8426184d2f032163)

cluster.go
```diff
@@ -259,6 +259,10 @@ func (c *clusterClient) _refresh() (err error) {
 			}
 			if len(g.nodes) > 1 {
 				n := len(g.nodes) - 1
+				for i, replica := range g.nodes[1:] {
+					rConn := conns[replica.Addr].conn
+					g.nodes[i+1].AZ = rConn.AZ()
```
rueian, line 264: All the others look good to me. The only thing left is that `AZ()` itself is too costly. Calling `AZ()` sequentially could block the `refresh` for a long time if all the `rConn` haven't actually connected.

I think we should continue to complete the `AZ_AFFINITY` feature in this PR so that we can skip `AZ()` calls when it is not configured on the client. Parallelizing `AZ()` calls can be another PR.

proost: [462f2a9](https://github.com/redis/rueidis/pull/721/commits/462f2a920ed06718df23cf0e9ece1fe91201e45c)

I'm perfectly wrong. How about this?

cluster.go
```diff
@@ -259,6 +259,33 @@ func (c *clusterClient) _refresh() (err error) {
 			}
 			if len(g.nodes) > 1 {
 				n := len(g.nodes) - 1
+
+				if c.opt.EnableReplicaAZLookUp {
+					var (
+						mu sync.Mutex
+						wg sync.WaitGroup
+					)
+					for i := 0; i < n; i += 4 { // batch 4 nodes for each goroutine
```
rueian, line 268:
```suggestion
					for i := 0; i < n; i += 4 { // batch AZ() for every 4 connections
```

rueian: `for each goroutine` is not quite accurate.

cluster.go
```diff
@@ -259,6 +259,33 @@ func (c *clusterClient) _refresh() (err error) {
...
+						for j := i; j < i+4 && j < n; j++ {
+							replica := g.nodes[j+1]
+							rConn := conns[replica.Addr].conn
+
+							wg.Add(1)
+							go func(j int, rConn conn) {
```
rueian, line 274:
```suggestion
							go func(j int, rConn conn, replica *ReplicaInfo) {
```
The `mu` is not necessary if we pass `*ReplicaInfo` into the goroutine.

rueidis.go
```diff
@@ -220,6 +220,10 @@ type ClientOption struct {
 	// NOTE: This function can't be used with ReplicaOnly option.
 	// NOTE: This function must be used with SendToReplicas function.
 	ReplicaSelector func(slot uint16, replicas []ReplicaInfo) int
+
+	// EnableReplicaAZLookUp enables the client to look up the AZ of the replica node.
+	// If true, the client will set `AZ` field in `ReplicaInfo`.
+	EnableReplicaAZLookUp bool
```
rueian, line 226:
```suggestion
	// EnableReplicaInfoAZ enables the client to load the replica node's availability zone.
	// If true, the client will set the `AZ` field in `ReplicaInfo`.
	EnableReplicaInfoAZ bool
```

cluster.go
proost: [afae5ee](https://github.com/redis/rueidis/pull/721/commits/afae5ee597ade7dbb2bc51191acee3253f371754)

cluster.go
proost: [afae5ee](https://github.com/redis/rueidis/pull/721/commits/afae5ee597ade7dbb2bc51191acee3253f371754)

Oh, I found that event `ReplicaInfo` doesn't need.

rueidis.go
proost: [929287f](https://github.com/redis/rueidis/pull/721/commits/929287feea3ffb9ffa92b4ff0bf65f7baba72316)

rueian (approved):
LGTM, thanks a lot @proost!

PR #727, opened 2025-01-28
feat: support connection lifetime for single client

terut:
## Background

Recently I noticed that request is unbalanced when its replica failover on memorystore for redis of GCP if the connection keeps. So I consider about connection lifetime to reconnect to redis endpoint because existing connection are not rerouted when a node reintroduced.

Here is the document about archtecture and connection balance manegement.

https://cloud.google.com/memorystore/docs/redis/about-read-replicas#architecture
https://cloud.google.com/memorystore/docs/redis/about-read-replicas#connection_balance_management

Ref: https://github.com/redis/rueidis/discussions/725

 ## Solution

Support connection lifetime for single client to reconnect fixed read endpoint.


terut:
@rueian Here is draft. Cloud you check your additional points on the discussion. There is no additional tests yet.

- Use p.lifeTm = time.AfterFunc(...) instead, because we need more fine-grained control over the timer.
- Stop p.lifeTm early if there is a network error or the connection is closed manually.
- For dedicated and blocking usages, we should stop p.lifeTm after acquiring the connection from the dpool or spool and reset it when putting the connection back.

pipe.go
```diff
@@ -1576,6 +1587,7 @@ func (p *pipe) Close() {
 	}
 	atomic.AddInt32(&p.waits, -1)
 	atomic.AddInt32(&p.blcksig, -1)
+	p.StopTimer()
```
terut, line 1590: Oops. I will fix it...

pool.go
```diff
@@ -59,6 +59,7 @@ retry:
 		}
 	}
 	p.cond.L.Unlock()
+	v.StopTimer()
```
rueian, line 62: If the timer is not stopped successfully, we need to acquire another connection.

pipe.go
```diff
@@ -89,6 +91,10 @@ type pipe struct {
 	recvs           int32
 	r2ps            bool // identify this pipe is used for resp2 pubsub or not
 	noNoDelay       bool
+	lftm            time.Duration // lifetime
+	lftmMu          sync.Mutex    // guards lifetime timer
```
rueian, line 95: Do we really need the mutex and the bool flag?

pool.go
terut: Ah, that's right. Thanks!

pipe.go
terut: I reviewed again, we don't need bool flag.
I thought that time.Reset and time.Stop need mutex when using `<= go 1.22 `. Maybe I've got it wrong.

rueian: The source looks like it is thread-safe https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/runtime/time.go;l=314.

terut: Thanks, As you said, it looks like thread-safe. I will remove it.

I misread `This cannot be done concurrent to other receives from the Timer's channel or other calls to the Timer's Stop method.` of https://pkg.go.dev/time@go1.21.13#Timer.Stop . Sorry.

rueian: And we are using the `AfterFunc` timer which has no channel associated.

terut: That's right.

pool.go
terut: Fixed [390e19b](https://github.com/redis/rueidis/pull/727/commits/390e19b48220265d50a59d22624e8280b55b009b)

pipe.go
terut: Fixed [f950c1e](https://github.com/redis/rueidis/pull/727/commits/f950c1e7294962a119774ffb33f4b6271278d932)

terut:
The rest is the implementation about retrying on singleclient. 
I feel like that maybe it's enough to use ConnLifetime option with enabling retry handler. What do you think about retrying by force for errConnExpired ? @rueian

pipe.go
rueian: Yes, this should be removed.

pipe.go
```diff
@@ -1576,6 +1585,7 @@ func (p *pipe) Close() {
 	}
 	atomic.AddInt32(&p.waits, -1)
 	atomic.AddInt32(&p.blcksig, -1)
+	p.StopTimer()
```
rueian, line 1588:
```suggestion
```

rueian:
> The rest is the implementation about retrying on singleclient. I feel like that maybe it's enough to use ConnLifetime option with enabling retry handler. What do you think about retrying by force for errConnExpired ? @rueian

I think we should use your original proposal and nothing to do with the retry handler.

```go
retry:
	resp = c.conn.Do(ctx, cmd)
	if resp.Error() == errConnExpired {
		goto retry
	}
	if c.retry && cmd.IsReadOnly() && c.isRetryable(resp.Error(), ctx) {
		...
```

Because whenever an errConnExpired occurs, we know the connection is closed by ourselves, so it should be safe to retry immediately.

terut:
@rueian Thanks. Surely we know the error and it's not good to show errConnExpired to outside when disabling retry too. Retry logic is almost done, just need to add that tests.

client.go
```diff
@@ -86,6 +90,13 @@ func (c *singleClient) DoMulti(ctx context.Context, multi ...Completed) (resps [
 	attempts := 1
 retry:
 	resps = c.conn.DoMulti(ctx, multi...).s
+	if c.hasConnLftm {
+		for _, resp := range resps {
+			if resp.Error() == errConnExpired {
```
rueian, line 95: Is it possible that `errConnExpired` happens in the middle of `DoMulti`? I am not sure, but If it is possible then we should not retry preceding requests that don't receive the error.

terut: Ah,  I think it's unlikely. Surely all responses have same error when changing `p.state`.

I will change that like the following.

```
if resps[0].Error() == errConnExpired {
  goto retry
}
```

terut: Fixed [c0c3657](https://github.com/redis/rueidis/pull/727/commits/c0c365740db9eeeae1d0833a475b1737078d8423)

rueian: ok, could you leave a comment in the code to explain why it won't happen?

terut: When I was checking the behavior of connection lifetime on concurrent process and then I found the error `read tcp [::1]:35190->[::1]:6379: use of closed network connection`  through `singleClient.DoMulti`. I think probably the error occurred because of `pipe.Close()` , but the investigation isn't going well. Could you advice me for that? @rueian

rueian: Hi @terut, I'm sorry that I just saw your message. I think the problem could be in the `_backgroundRead`. It returns the network error directly without checking the p.Error().

rueian: https://github.com/redis/rueidis/blob/b799437d0fd865cf2c1342a4edbf82abecc66d77/pipe.go#L490
https://github.com/redis/rueidis/blob/b799437d0fd865cf2c1342a4edbf82abecc66d77/pipe.go#L616

terut: @rueian Thanks! As you said, it looks like that `_backgroundRead` returns that error.

I will change that lines to return `errConnExpired` when expired and then `errConnExpired` happens in the middle of `DoMulti`, so we may should check the error of all response.

rueian: Hi @terut, any update?

terut: Sorry for late. I had no time to spare... I think I can work on this problem from this week. @rueian

rueian: > Sorry for late. I had no time to spare... I think I can work on this problem from this week.

No worries. 

> Sometime incremented value is 10001, 10002 and so on when loop is 10000.

What do you mean by this? I think this can be a general feature for those who want a limited lifetime on each connection for any reason.

terut: When I just counted up for 10000 times using connection lifetime option, the value of keys is over 10000. But my implementation is not correct for now at the point of view of the error handling of `_backgroundRead`, I will try to count up again after implementing correctly.

rueian:
Hi @terut, would you mind adding the retry logic to the cluster client and sentinel client in a follow-up PR?

terut:
@rueian Sorry for waiting long time. Okay, I'll take care of it.

rueian:
Thanks @terut!

terut:
@rueian Thank you for your great help for a long time 😄

rueian:
You’re welcome! I know some users really need this feature, so it’s great that we have it. However, I hope we can have your follow-up PR for adding retries to cluster and sentienl soon because next week is the next release cycle. If we don't have the PR merged, we probably can't include this new feature in the next release.

Issue #738, opened 2025-02-07
Add ACLLogReset to rueidiscompat


rueian:
Take the below `ACLDryRun` as an example to implement the new function in `rueidiscompat` that provides the same interface as go-redis:

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/adapter.go#L3199-L3203

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/pipeline.go#L2086-L2090

See the parent issue for more details https://github.com/redis/rueidis/issues/736.

hcc429:
Hi! May I take this issue ? Thanks!

PR #745, opened 2025-02-09
add ACLLogReset to rueidiscompat

hcc429:
As #738  mentioned, add an ACLLogReset method to provide same interface with go-redis.


rueian:
Thank you @hcc429.

Issue #739, opened 2025-02-07
Add ACLSetUser to rueidiscompat


rueian:
Take the below `ACLDryRun` as an example to implement the new function in `rueidiscompat` that provides the same interface as go-redis:

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/adapter.go#L3199-L3203

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/pipeline.go#L2086-L2090

See the parent issue for more details https://github.com/redis/rueidis/issues/736.

astorig:
hi, I would like to take this issue. could you assign it to me?

PR #744, opened 2025-02-09
feat: Add ACLLog to rueidiscompat

blueBlue0102:
#737 

I have run the unit tests multiple times locally, and each time, a different test seems to fail unpredictably.
I'm not sure if this is an issue specific to my local environment.

Unit test results:
image
image
image

PR #746, opened 2025-02-09
feat: add ACLSetUser to rueidiscompat

astorig:
#739


astorig:
i have encountered the same issue with unit tests failing unpredictably with the same error as described in [PR#744](https://github.com/redis/rueidis/pull/744). it seems to be a recurring issue.

rueian:
Thanks @astorig. Don't worry about the test. It is not relevant to this PR.

Issue #749, opened 2025-02-11
Add LCS to rueidiscompat


rueian:
The `rueidiscompat` sub-module aims to provide the same interface as go-redis as closely as possible while utilizing rueidis under the hood.

go-redis has the following interface:

```go
// https://github.com/redis/go-redis/blob/40e049e67a4cce574f7972b53b1a8172bbe45d5e/string_commands.go#L20
type StringCmdable interface {
	...
	LCS(ctx context.Context, q *LCSQuery) *LCSCmd
	...
}
```

while in `rueidiscompat`, it is currently not implemented:

https://github.com/redis/rueidis/blob/20fc2cd51af4bbb3bc92dbead900c151f25fba1e/rueidiscompat/adapter.go#L183

Therefore, we should catch up go-redis by implementing the function to the `rueidiscompat.Compat` and the `rueidiscompat.Pipeline` struct.

Take the below `ACLDryRun` as an example to implement the new function in `rueidiscompat` that provides the same interface as go-redis:

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/adapter.go#L3199-L3203

https://github.com/redis/rueidis/blob/93e9af3c7c1a6b3a94ff4da64256c6df6c47d81e/rueidiscompat/pipeline.go#L2086-L2090

arbha1erao:
Hi, I would like to work on this issue. Thanks!

rueian:
Hi @arbha1erao, thank you! Let me know if you have any problems during the work.

rueian:
Hi @arbha1erao, how are you doing? Is everything going well?

arbha1erao:
I’ve created a PR—let me know if you have any feedback.

arbha1erao:
Also, I haven’t added any tests yet, as I wasn’t sure where to place them. Could you please guide me on that?"

rueian:
Hi @arbha1erao, It is ok not to have a test in code for this right now, as long as if your implementation looks good.

PR #767, opened 2025-02-16
Add LCS

arbha1erao:
#749

### DIFF
--- a/rueidiscompat/command.go
+++ b/rueidiscompat/command.go
@@ -5182,15 +5182,20 @@ func (cmd *LCSCmd) from(res rueidis.RedisResult) {
 		// match string
 		if lcs.MatchString, err = res.ToString(); err != nil {
 			cmd.SetErr(err)
+			return
 		}
 	case 2:
 		// match len
 		if lcs.Len, err = res.AsInt64(); err != nil {
 			cmd.SetErr(err)
+			return
 		}
 	case 3:
 		// read LCSMatch
-		if msgMap, err := res.AsMap(); err == nil {
+		if msgMap, err := res.AsMap(); err != nil {
+			cmd.SetErr(err)
+			return
+		} else {
 			// Validate length (should have exactly 2 keys: "matches" and "len")
 			if len(msgMap) != 2 {
 				cmd.SetErr(fmt.Errorf("redis: got %d elements in the map, wanted %d", len(msgMap), 2))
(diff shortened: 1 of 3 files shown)


rueidiscompat/command.go
```diff
@@ -5008,3 +5008,51 @@ func (cmd *ClusterLinksCmd) Val() []ClusterLink {
...
+	baseCmd[any]
+}
+
+func newLCSCmd(res rueidis.RedisResult) *LCSCmd {
+	cmd := &LCSCmd{}
+	cmd.from(res)
+	return cmd
+}
+
+func (cmd *LCSCmd) from(res rueidis.RedisResult) {
```
rueian, line 5031: This parsing function seems incorrect. Is it compatible with https://github.com/redis/go-redis/blob/196fc9b21ac460f0d5251d349e4249e2ffedb9ff/command.go#L4510?

arbha1erao: My bad. I misunderstood what was required. I’ve added a new commit with the necessary changes.

rueidiscompat/command.go
```diff
@@ -5107,3 +5107,189 @@ func newSlowLogCmd(res rueidis.RedisResult) *SlowLogCmd {
...
+	lcs := &LCSMatch{}
+	var err error
+
+	switch cmd.readType {
+	case 1:
+		// match string
+		if lcs.MatchString, err = res.ToString(); err == nil {
```
rueian, line 5183:
```suggestion
		if lcs.MatchString, err = res.ToString(); err != nil {
```
Shouldn't this be `err != nil`?

rueidiscompat/command.go
```diff
@@ -5107,3 +5107,189 @@ func newSlowLogCmd(res rueidis.RedisResult) *SlowLogCmd {
...
+			cmd.SetErr(err)
+		}
+	case 2:
+		// match len
+		if lcs.Len, err = res.AsInt64(); err == nil {
```
rueian, line 5188:
```suggestion
		if lcs.Len, err = res.AsInt64(); err != nil {
```
Shouldn't this be `err != nil`?

rueidiscompat/command.go
```diff
@@ -5107,3 +5107,189 @@ func newSlowLogCmd(res rueidis.RedisResult) *SlowLogCmd {
...
+			cmd.SetErr(err)
+		}
+	case 3:
+		// read LCSMatch
+		if msgMap, err := res.AsMap(); err == nil {
```
rueian, line 5193: What if `err != nil`?

arbha1erao: Addressed

rueian (approved):

rueian:
Thank you @arbha1erao

Issue #771, opened 2025-02-20
How can I make use of both autopipelining and pool


liuzhaohui:
It's proved that autopipelining can lower average cost and improve throughput in my system. But we encounter some 'badcases' that really annoying.  
Most of the requests(command) are small commands and are fast(we know the commands' pattern), while some commands may be large and slow (we donot know which commands they are). We are using the pool mode (disable autopipelining) and get the same performance compared to goredis we used before.
We have considered the dedicate interface, but it is inconvenient for us because it's difficult for us to figure out which commands to dedicate. And it seems that the spool instead of dpool(dedicated pool) is for our senario.  Currently, we can only choose either blocking(use pool) or autopipelining, not both.
Is there any solution?

rueian:
> we could probably have a new interface allowing you to ...

We could reuse this path to opt in auto pipelining explicitly:
https://github.com/redis/rueidis/blob/050c4d92b3a122899648162921e82768a7051fef/mux.go#L225-L227

liuzhaohui:
@rueian Thank you for reply and good suggestions. Reusing opt in mux.go sounds a good idea. That means we could add a new tag value if necessary to facilliate?

rueian:
Yes, for example, the new interface could look like this:

```go
cmd := client.B().Get().Key("key").Build().ToPipe()
client.Do(ctx, cmd)
```

The `ToPipe()` could modify the tag value for the mux to pipeline the command.

liuzhaohui:
> Yes, for example, the new interface could look like this:
> 
> cmd := client.B().Get().Key("key").Build().ToPipe()
> client.Do(ctx, cmd)
> The `ToPipe()` could modify the tag value for the mux to pipeline the command.

Thank you, That's greate, maybe I could try it in my local branch.

PR #773, opened 2025-02-21
add feature to cmd, give a choice to change execution from spool to auto pipeline

liuzhaohui:
A solution to issue #771 

```go
client, err := NewClient(ClientOption{
    InitAddress:       []string{"127.0.0.1:6379"},
    // other options
    DisableAutoPipelining:  true,  // change default action to use pool 
})

cmd := client.B().Get().Key(key).Build().ToPipe()   // assign this command to auto pipeline manually
client.Do(ctx, cmd)
```

### DIFF
--- a/internal/cmds/cmds.go
+++ b/internal/cmds/cmds.go
@@ -10,7 +10,7 @@ const (
 	mtGetTag = uint16(1<<11) | readonly // make mtGetTag can also be retried
 	scrRoTag = uint16(1<<10) | readonly // make scrRoTag can also be retried
 	unsubTag = uint16(1<<9) | noRetTag
-	pipeTag  = uint16(1 << 8) // make blocking mode request can use auto pipelining
+	pipeTag  = uint16(1<<8) // make blocking mode request can use auto pipelining
 	// InitSlot indicates that the command be sent to any redis node in cluster
 	InitSlot = uint16(1 << 14)
 	// NoSlot indicates that the command has no key slot specified
@@ -109,7 +109,7 @@ func (c Completed) Pin() Completed {
 	return c
 }
 
-// ToPipe marks the command with pipeTag
+// ToPipe returns a new command with pipeTag
 func (c Completed) ToPipe() Completed {
 	c.cf |= pipeTag
 	return c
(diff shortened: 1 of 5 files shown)


mux.go
```diff
@@ -210,7 +210,7 @@ func (m *mux) DoMultiStream(ctx context.Context, multi ...Completed) MultiRedisR
 }
 
 func (m *mux) Do(ctx context.Context, cmd Completed) (resp RedisResult) {
-	if m.usePool && !cmd.NoReply() {
+	if m.usePool && !cmd.NoReply() && !cmd.IsPipe() {
```
rueian, line 213:
```suggestion
	if m.usePool && !cmd.IsPipe() {
```
Can we just keep IsPipe() here? That is, add the new pipeTag into the existing noRetTag.

internal/cmds/cmds.go
```diff
@@ -10,6 +10,7 @@ const (
 	mtGetTag = uint16(1<<11) | readonly // make mtGetTag can also be retried
 	scrRoTag = uint16(1<<10) | readonly // make scrRoTag can also be retried
 	unsubTag = uint16(1<<9) | noRetTag
+	pipeTag  = uint16(1 << 8) // make blocking mode request can use auto pipelining
```
rueian, line 13:
```suggestion
	pipeTag  = uint16(1<<8) // make blocking mode request can use auto pipelining
```
Just aligning the style above.

rueian:
It mostly looks good! Could you also add an example to the Readme's [`Disable Auto Pipelining`](https://github.com/redis/rueidis?tab=readme-ov-file#disable-auto-pipelining) section?

internal/cmds/cmds.go
```diff
@@ -108,6 +109,12 @@ func (c Completed) Pin() Completed {
 	return c
 }
 
+// ToPipe marks the command with pipeTag
```
rueian, line 112:
```suggestion
// ToPipe returns a new command with pipeTag
```

internal/cmds/cmds.go
```diff
@@ -143,6 +150,11 @@ func (c *Completed) IsWrite() bool {
 	return !c.IsReadOnly()
 }
 
+// IsPipe checks if it is set pipeTag which prefers auto pipelining
+func (c *Completed) IsPipe() bool {
+	return c.NoReply() || c.cf&pipeTag == pipeTag
```
rueian, line 155: Can't we just do `noRetTag = uint16(1<<12) | readonly | pipeTag`?

rueian (approved):

PR #777, opened 2025-02-23
Replace error atomic.Value with error atomic.Pointer[errs]

### LINKED ISSUE #776: Replace `error atomic.Value` with `error atomic.Pointer[errs]`
Currently, we have this:
https://github.com/redis/rueidis/blob/b799437d0fd865cf2c1342a4edbf82abecc66d77/pipe.go#L64

Since our minimum go version is 1.22 now, the declaration can be replaced by `error atomic.Pointer[errs]` literally to slightly improve performance and avoid some mistakes in the future.

tarunprabhu11:
Closes #776


rueian:
Hi @tarunprabhu11, you also need to resolve build errors.

PR #793, opened 2025-03-04
feat: add time to live for every idle connection

liuzhaohui:
### Intuition
When ClientOption.BlockingPoolCleanup is set, all connections beyond BlockingPoolMinSize in idle list of the pool will be recycled. Which will cause connection be closed and created frequently even if my application has high concurrency. As my obervation, at the very moment the timer function is triggered, there may be more than BlockingPoolMinSize of connection in the idle list which are store recently and supposed to be acquired very soon.

### Approach
My immediate thougth is adding an access time to every connection. The access time is updated when connection is put back into the pool.  Access time will be checked when recycling. If time.Since(accessTime) < BlockingPoolConnIdleTime, that connection will not be recycled. BlockingPoolConnIdleTime is a new ClientOption to define the idle time for every connection.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -101,8 +101,8 @@ func newMux(dst string, option *ClientOption, init, dead wire, wireFn wireFn, wi
 		m.wire[i].Store(init)
 	}
 
-	m.dpool = newPool(option.BlockingPoolSize, dead, option.BlockingPoolCleanup, option.BlockingPoolMinSize, wireFn)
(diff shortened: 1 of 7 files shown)


liuzhaohui:
> Hi @liuzhaohui, thank you for the contribution, but I don’t think we really need this, and storing access time looks expensive. You can avoid frequent reconnections by a larger BlockingPoolMinSize and a longer BlockingPoolCleanup.

Thanks for replying. Currently, I'm using a  large BlockingPoolMinSize and setting BlockingPoolCleanup to 0 to disable clean up. As you mentioned, this is not a necessary feature.

PR #797, opened 2025-03-11
fix: restore rueidisaside go.mod and add module validation

### LINKED ISSUE #796: Accidental deletion of modules go.mod files causing ambiguous imports
There were two incidents where module `go.mod` files were accidentally deleted:
1. `rueidislimiter` `go.mod` was deleted in commit 34c5717
2. `rueidisaside`  `go.mod` was deleted in commit 87327f9

RueidisAside:
https://go.dev/play/p/iC7P-EDfbff

RueidisLimiter:
https://go.dev/play/p/KTVW5CW_2XJ

Current Status
- `rueidislimiter` has been fixed after #795, but a new release is still needed
- `rueidisaside` still needs its `go.mod` restored
- `rueidislock` was never a module in self so it's fine but maybe it should be

nesty92:
Adding CI validation to prevent future accidental module deletions

Closes #796

### DIFF
--- a/rueidisaside/go.mod
+++ b/rueidisaside/go.mod
@@ -1,0 +1,11 @@
+module github.com/redis/rueidis/rueidisaside
+
+go 1.22.0
+
+toolchain go1.24.1
+
+replace github.com/redis/rueidis => ../
+
+require github.com/redis/rueidis v1.0.55
+
+require golang.org/x/sys v0.30.0 // indirect
--- a/rueidisaside/go.mod
+++ b/rueidisaside/go.mod
@@ -2,8 +2,6 @@ module github.com/redis/rueidis/rueidisaside
 
 go 1.22.0
 
-toolchain go1.24.1
-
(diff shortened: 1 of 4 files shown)


nesty92:
Hi @rueian, a new tag will be needed after this is merged to fully fix the problem when importing the modules

rueidisaside/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/redis/rueidis/rueidisaside
+
+go 1.22.0
+
+toolchain go1.24.1
```
rueian, line 5:
```suggestion
```
We don’t use the toolchain directive in other mod files. Could you also help delete https://github.com/redis/rueidis/blob/696b5648b60bd6971e0f0d39f6c6ed964059bccd/rueidislimiter/go.mod#L5?

nesty92: Sure not problem

rueian (approved):

Issue #757, opened 2025-02-13
Use timer for background PINGs instead of a long-live goroutine


rueian:
Currently, we use a long-live goroutine for each connection to do background PINGs:
https://github.com/redis/rueidis/blob/31719474cb562b2eb93672a4ae4bed0eee8f8d3b/pipe.go#L325-L327

In the current approach, we create the same amount of background goroutines as the number of connections that wait periodically to send a PING to the connection. I’m curious to know if using `time.AfterFunc` instead would enhance memory usage and improve the scheduling for other more important goroutines.

SoulPancake:
This can be closed I guess @rueian

PR #805, opened 2025-03-16
Use timer for background PINGs instead of a long-live goroutine

liuzhaohui:
Try to solve #757

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -75,6 +75,7 @@ type pipe struct {
 	ssubs           *subs                       // pubsub smessage subscriptions
 	nsubs           *subs                       // pubsub  message subscriptions
 	psubs           *subs                       // pubsub pmessage subscriptions
+	pingTimer       *time.Timer					// timer for background ping
 	info            map[string]RedisMessage
 	timeout         time.Duration
 	pinggap         time.Duration
@@ -90,7 +91,6 @@ type pipe struct {
 	r2ps            bool // identify this pipe is used for resp2 pubsub or not
 	noNoDelay       bool
 	optin           bool
-	pingTimer       *time.Timer
 }
 
 type pipeFn func(ctx context.Context, connFn func(ctx context.Context) (net.Conn, error), option *ClientOption) (p *pipe, err error)
@@ -636,17 +636,13 @@ func (p *pipe) backgroundPing() {
 	p.pingTimer = time.AfterFunc(p.pinggap, func() {
 		var err error
 		recv = atomic.LoadInt32(&p.recvs)
-		reset := false
 		defer func(){
 			prev = atomic.LoadInt32(&p.recvs)
-			if reset {
+			if err == nil {
 				p.pingTimer.Reset(p.pinggap)
-			} else {
-				p.pingTimer.Stop()
 			}
 		}()
 		if recv != prev || atomic.LoadInt32(&p.blcksig) != 0 || (atomic.LoadInt32(&p.state) == 0 && atomic.LoadInt32(&p.waits) != 0) {
-			reset = true
 			return
 		}
 		ch := make(chan error, 1)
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -629,37 +629,48 @@ func (p *pipe) _backgroundRead() (err error) {
...
-				err = os.ErrDeadlineExceeded
-			case err = <-ch:
-				tm.Stop()
-			}
-			if err != nil && atomic.LoadInt32(&p.blcksig) != 0 {
-				err = nil
+	var timer *time.Timer
+	
+	prev = atomic.LoadInt32(&p.recvs)
+	timer = time.AfterFunc(p.pinggap, func() {
```
rueian, line 636: The `timer` should be stored in `p` so we can stop it when closing it.

pipe.go
```diff
@@ -629,37 +629,48 @@ func (p *pipe) _backgroundRead() (err error) {
...
+			return
+		}
+		ch := make(chan error, 1)
+		tm := time.NewTimer(p.timeout)
+		go func() { ch <- p.Do(context.Background(), cmds.PingCmd).NonRedisError() }()
+		select {
+		case <-tm.C:
+			err = os.ErrDeadlineExceeded
+		case err = <-ch:
+			// noop
```
rueian, line 657:
```suggestion
			tm.Stop()
```

pipe.go
```diff
@@ -629,37 +629,48 @@ func (p *pipe) _backgroundRead() (err error) {
...
+		if err != nil && atomic.LoadInt32(&p.blcksig) != 0 {
+			err = nil
+		}
+		if err != nil && err != ErrClosing {
+			p._exit(err)
+		}
+		select {
 		case <-p.close:
 			return
+		default:
```
rueian, line 668:
```suggestion
```

rueian: Close the timer when `p` is closed instead.

liuzhaohui: Fixed

pipe.go
```diff
@@ -90,6 +90,7 @@ type pipe struct {
 	r2ps            bool // identify this pipe is used for resp2 pubsub or not
 	noNoDelay       bool
 	optin           bool
+	pingTimer       *time.Timer
```
rueian, line 93: This should be moved to close other pointer fields.

pipe.go
```diff
@@ -629,37 +630,49 @@ func (p *pipe) _backgroundRead() (err error) {
...
+	p.pingTimer = time.AfterFunc(p.pinggap, func() {
+		var err error
+		recv = atomic.LoadInt32(&p.recvs)
+		reset := false
+		defer func(){
+			prev = atomic.LoadInt32(&p.recvs)
+			if reset {
+				p.pingTimer.Reset(p.pinggap)
+			} else {
+				p.pingTimer.Stop()
```
rueian, line 645:
```suggestion
```

No need.

rueian (approved):

rueian:
Although the changes look good, I still like to know if this really enhances memory usage and improves the scheduling for other more important goroutines.

liuzhaohui:
> Although the changes look good, I still like to know if this really enhances memory usage and improves the scheduling for other more important goroutines.

Ok, I'll try to get some data. May be controlled environments only background pings and measure their performance differences.

rueian:
I am not sure if you can use `for range b.N {` multiple times in multiple goroutines, but I guess we can benchmark the difference between long-live goroutines and timers without rueidis involved. See @SoulPancake's ongoing work on this https://github.com/redis/rueidis/pull/760#discussion_r1987868546

liuzhaohui:
@rueian 
I wrote a [testcase](https://github.com/liuzhaohui/test/blob/master/go/timer_vs_goroutine/main.go) and got some results [here](https://github.com/liuzhaohui/test/blob/master/go/timer_vs_goroutine/README.md)

rueian:
Thanks @liuzhaohui!

Issue #806, opened 2025-03-19
Replace all `sync.Once` with int32


rueian:
We don’t actually require `sync.Once`; a straightforward `int32` with an atomic swap suffices.

PR #808, opened 2025-03-20
Replace sync.Once with int32 in pipe

liuzhaohui:
Try to solve #806

### DIFF
--- a/mock/result.go
+++ b/mock/result.go
@@ -174,7 +174,6 @@ type pipe struct {
 	timeout         time.Duration
 	pinggap         time.Duration
 	maxFlushDelay   time.Duration
-	once            sync.Once
 	r2mu            sync.Mutex
 	version         int32
 	_               [10]int32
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -80,14 +80,14 @@ type pipe struct {
...
 	maxFlushDelay   time.Duration
-	once            sync.Once
 	r2mu            sync.Mutex
 	version         int32
 	_               [10]int32
 	blcksig         int32
 	state           int32
 	waits           int32
 	recvs           int32
+	bgState			int32
```
rueian, line 90: The indention seems incorrect.

mock/result.go
```diff
@@ -174,7 +174,6 @@ type pipe struct {
 	timeout         time.Duration
 	pinggap         time.Duration
 	maxFlushDelay   time.Duration
-	once            sync.Once
```
rueian, line 177: A `bgState` should also be added to the same position.

pipe.go
liuzhaohui: Fixed

rueian (approved):

Issue #807, opened 2025-03-19
Merge `recvs` and `waits` into one `uint64`.


rueian:
We have two `int32` counters in the `pipe` struct, `recvs` and `waits`, for tracking concurrent requests on the `pipe`.

`waits` will be increased when receiving a request and decreased when the request is finished.
`recvs` will be increased when a request is finished and never decreased.

For `recvs`, we don't care if it overflows because we don't care about its actual value but only check if it changes.
For `waits`, we know it will not overflow because it should stay close to 0.

Can we merge these counters into one `uint64` to save atomic operations? For example:

`recvs` is the upper 32bits and `waits` is the lower 32bits.
when receiving a request, we increment the `uint64` by 1.
when a request is finished, we decrement the `uint64` by 1 + 1<<32.

SoulPancake:
Picking this up @rueian 
Can you please assign

rueian:
Sure, remember to use the `atomic.Uint64{}` struct.

PR #809, opened 2025-03-20
feat: merge waits and recvs to wrCounter

SoulPancake:
@rueian Please take a look at this
solves #807

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -371,10 +371,10 @@ func (p *pipe) _background() {
 		select {
 		case <-p.close:
 		default:
-			p.IncrWaits()
+			p.incrWaits()
 			go func() {
 				<-p.queue.PutOne(cmds.PingCmd) // avoid _backgroundWrite hanging at p.queue.WaitForWrite()
-				p.DecrWaits()
+				p.decrWaits()
 			}()
 		}
 	}
@@ -405,7 +405,7 @@ func (p *pipe) _background() {
 	}
 
 	resp := newErrResult(err)
-	for p.LoadWaits() != 0 {
+	for p.loadWaits() != 0 {
 		select {
 		case <-p.close: // p.queue.NextWriteCmd() can only be called after _backgroundWrite
 			_, _, _ = p.queue.NextWriteCmd()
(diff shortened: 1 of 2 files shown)


SoulPancake:
taking a look at the test failure, race

SoulPancake:
I did that now,  still miss what went wrong, can you help me out? @rueian

rueian:
> I did that now, still miss what went wrong, can you help me out? @rueian

`^uint64(1)` for decrement doesn't seem correct. Also, we need to merge `p.DecrWaits()` and `p.IncrRecvs()` into one `pDecrWaitsAndRecvs`. That's the whole point of this improvement. Another thing is, my bad, we should make these helper functions private.

rueian:
So we have two cases for decrements
1. `p.decrWaits` only decrements the lower 32bit. The upper 32bits should not be touched.
2. `p.decrWaitsAndRecvs` decrements both upper 32bit and lower 32bit in one operation.

SoulPancake:
please take a look at it once @rueian

pipe.go
```diff
@@ -907,10 +906,10 @@ func (p *pipe) Do(ctx context.Context, cmd Completed) (resp RedisResult) {
 	} else {
 		resp = newErrResult(p.Error())
 	}
-	if left := atomic.AddInt32(&p.waits, -1); state == 0 && left != 0 {
+	if left := p.decrWaits(); state == 0 && left != 0 {
 		p.background()
 	}
-	atomic.AddInt32(&p.recvs, 1)
+	p.incrRecvs()
```
rueian, line 912: These should be merged to `decrWaitsAndIncrRecvs`.

pipe.go
```diff
@@ -1557,6 +1552,44 @@ func (p *pipe) DoMultiCache(ctx context.Context, multi ...CacheableTTL) *redisre
...
+	return uint32(p.wrCounter.Add(^uint64(0)) & 0xFFFFFFFF)
+}
+
+// incrRecvs increments the upper 32 bits (recvs) of wrCounter by 1.
+func (p *pipe) incrRecvs() uint32 {
+	// Increment the upper 32 bits (recvs)
+	p.wrCounter.Add(1 << 32)
+	// Return the updated value of recvs (upper 32 bits)
+	return uint32(p.wrCounter.Load() >> 32)
+}
```
rueian, line 1573:
```suggestion
```
We should not have this.

pipe.go
```diff
@@ -1557,6 +1552,44 @@ func (p *pipe) DoMultiCache(ctx context.Context, multi ...CacheableTTL) *redisre
 	return results
 }
 
+// incrWaits increments the lower 32 bits (waits).
+func (p *pipe) incrWaits() uint32 {
+	// Increment the lower 32 bits (waits)
+	return uint32(p.wrCounter.Add(1) & 0xFFFFFFFF)
```
rueian, line 1558:
```suggestion
	return uint32(p.wrCounter.Add(1))
```
Do we really need `& 0xFFFFFFFF`?

SoulPancake:
Please re-review @rueian Sorry for the delay

Issue #789, opened 2025-03-02
perf: reduce the size of RedisMessage struct by merging fields


rueian:
Currently, we use this struct for holding a redis response:

```go
type RedisMessage struct {
	attrs   *RedisMessage
	string  string
	values  []RedisMessage
	integer int64
	typ     byte
	ttl     [7]byte
}
```

Where only one of the three fields `string`, `values`, and `integer` will be used at a time. They occupy 16+24+8 bytes which seems a bit wasteful. Could we merge them by using `unsafe.SliceData` and `unsafe.StringData`, and store the length of either `string` or `values` in the `integer` field:

```go
type RedisMessage struct {
	attrs   *RedisMessage
	string  *byte
	values  *int
	integer int64
	typ     byte
	ttl     [7]byte
}
```

By doing this, we can reduce the struct size by 24 bytes (48 -> 24 = 50%).

fscnick:
Hi @rueian, I'd like to help with this.

liuzhaohui:
@rueian Can we use the following RedisMessage which retain the type info making it a specific type of pointer?

```
type RedisMessage struct {
	attrs   *RedisMessage
	string  *string // instead of *byte
	values  *[]RedisMessage // instead of *int
	integer int64
	typ     byte
	ttl     [7]byte
}
```

or general type any (but more complex to manipulate)

```
type RedisMessage struct {
	attrs   *RedisMessage
	data   any // either string/[]RedisMessage/integer
	typ     byte
	ttl     [7]byte
}
```

rueian:
Hi @liuzhaohui, though they reduce the size of the structure, they create an indirection for dereference and put pressure on GC. I don’t think they are worthy. Why do you think they can be better?

liuzhaohui:
> Hi [@liuzhaohui](https://github.com/liuzhaohui), though they reduce the size of the structure, they create an indirection for dereference and put pressure on GC. I don’t think they are worthy. Why do you think they can be better?

I'm not sure which one is better. At the moment, there'are 3 candidate approches in my mind and I'm really struggling to figure out which one is better.

| No. | approach | pros. | cons. | 
| ----- | ----- | ----- | ----- | 
| 1 | using \*byte/\*int.  | more controll on our side | manually convert between raw pointer and string/[]RedisMessage type. What about the data \*byte or \*int pointer to, will it under the control of gc or our own pool manaement? | 
[...]

PR #810, opened 2025-03-21
feat: reduce size of redis message

fscnick:
Replace the fields in `RedisMessage` with `*byte` and `*RedisMessage` in conjunction with integer to save the size of struct.
Use `unsafe.String()` or `unsafe.Slice()` to recover back to the original `string` or `[]RedisMessage` accordingly.

related issue: https://github.com/redis/rueidis/issues/789

### DIFF
--- a/cache_test.go
+++ b/cache_test.go
@@ -23,14 +23,14 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("flights before Update should return empty RedisMessage and non-nil CacheEntry")
 		}
 
-		store.Delete([]RedisMessage{redisMessageContainString('+', "key")}) // Delete should not affect pending CacheEntry
+		store.Delete([]RedisMessage{strmsg('+', "key")}) // Delete should not affect pending CacheEntry
 
 		v2, e2 := store.Flight("key", "cmd", time.Millisecond*100, now)
 		if v2.typ != 0 || e != e2 {
 			t.Fatal("flights before Update should return empty RedisMessage and the same CacheEntry, not be affected by Delete")
 		}
 
-		v = redisMessageContainString('+', "val")
+		v = strmsg('+', "val")
 		v.setExpireAt(now.Add(time.Second).UnixMilli())
 		if pttl := store.Update("key", "cmd", v); pttl < now.Add(90*time.Millisecond).UnixMilli() || pttl > now.Add(100*time.Millisecond).UnixMilli() {
 			t.Fatal("Update should return a desired pttl")
@@ -52,7 +52,7 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("CachePXAT should return a desired pttl")
 		}
 
-		store.Delete([]RedisMessage{redisMessageContainString('+', "key")})
+		store.Delete([]RedisMessage{strmsg('+', "key")})
 		v, e = store.Flight("key", "cmd", time.Millisecond*100, now)
 		if v.typ != 0 || e != nil {
 			t.Fatal("flights after Delete should return empty RedisMessage and nil CacheEntry")
@@ -74,7 +74,7 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("flights before Update should return empty RedisMessage and non-nil CacheEntry")
 		}
 
-		store.Delete([]RedisMessage{redisMessageContainString('+', "key")}) // Delete should not affect pending CacheEntry
+		store.Delete([]RedisMessage{strmsg('+', "key")}) // Delete should not affect pending CacheEntry
 
 		v2, e2 := store.Flight("key", "cmd", time.Millisecond*100, now)
 		if v2.typ != 0 || e != e2 {
@@ -99,13 +99,13 @@ func test(t *testing.T, storeFn func() CacheStore) {
 		var store = storeFn()
 
 		for _, deletions := range [][]RedisMessage{
-			{redisMessageContainString('+', "key")},
+			{strmsg('+', "key")},
 			nil,
 		} {
 			store.Flight("key", "cmd1", time.Millisecond*100, now)
 			store.Flight("key", "cmd2", time.Millisecond*100, now)
-			store.Update("key", "cmd1", redisMessageContainString('+', "val"))
-			store.Update("key", "cmd2", redisMessageContainString('+', "val"))
+			store.Update("key", "cmd1", strmsg('+', "val"))
+			store.Update("key", "cmd2", strmsg('+', "val"))
 
 			store.Delete(deletions)
 
@@ -128,7 +128,7 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("first flight should return empty RedisMessage and nil CacheEntry")
 		}
 
-		v = redisMessageContainString('+', "val")
+		v = strmsg('+', "val")
 		v.setExpireAt(now.Add(time.Millisecond).UnixMilli())
 		store.Update("key", "cmd", v)
 
--- a/cache_test.go
+++ b/cache_test.go
@@ -23,21 +23,21 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("flights before Update should return empty RedisMessage and non-nil CacheEntry")
 		}
 
-		store.Delete([]RedisMessage{{typ: '+', string: "key"}}) // Delete should not affect pending CacheEntry
+		store.Delete([]RedisMessage{redisMessageContainString('+', "key")}) // Delete should not affect pending CacheEntry
 
 		v2, e2 := store.Flight("key", "cmd", time.Millisecond*100, now)
 		if v2.typ != 0 || e != e2 {
 			t.Fatal("flights before Update should return empty RedisMessage and the same CacheEntry, not be affected by Delete")
 		}
 
-		v = RedisMessage{typ: '+', string: "val"}
+		v = redisMessageContainString('+', "val")
 		v.setExpireAt(now.Add(time.Second).UnixMilli())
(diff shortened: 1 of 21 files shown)


rueian:
Build failed. The mock package should also be updated.

cache_test.go
```diff
@@ -23,36 +23,36 @@ func test(t *testing.T, storeFn func() CacheStore) {
 			t.Fatal("flights before Update should return empty RedisMessage and non-nil CacheEntry")
 		}
 
-		store.Delete([]RedisMessage{{typ: '+', string: "key"}}) // Delete should not affect pending CacheEntry
+		store.Delete([]RedisMessage{redisMessageContainString('+', "key")}) // Delete should not affect pending CacheEntry
```
rueian, line 26:
```suggestion
		store.Delete([]RedisMessage{strmsg('+', "key")}) // Delete should not affect pending CacheEntry
```
I prefer a shorter function name.

resp.go
```diff
@@ -77,12 +77,16 @@ func init() {
 }
 
 func readSimpleString(i *bufio.Reader) (m RedisMessage, err error) {
-	m.string, err = readS(i)
+	var s string
+	s, err = readS(i)
+	m.setString(s)
```
rueian, line 82: Using `readS` first and then `setString` seems a little redundant. Can we make `readS` return the length and the unsafe.SliceData directly? Same to the `readB`, `readA`, and `readE`.

message.go
```diff
@@ -52,11 +52,19 @@ func IsRedisErr(err error) (ret *RedisError, ok bool) {
 // RedisError is an error response or a nil message from redis instance
 type RedisError RedisMessage
 
+// string retrives the contained string of the RedisMessage
+func (m *RedisError) string() string {
```
rueian, line 56:
```suggestion
func (m RedisError) string() string {
```
Is there any performance downside to using a value receiver on this function? Can this function be inlined?

message.go
```diff
@@ -523,24 +531,51 @@ func (r *prettyRedisResult) MarshalJSON() ([]byte, error) {
...
+	array *RedisMessage
+
+	// integer is used for a simple number or
+	// in conjunction with array or bytes to store the length of array or string
 	integer int64
 	typ     byte
 	ttl     [7]byte
 }
 
+func (m *RedisMessage) string() string {
```
rueian, line 545:
```suggestion
func (m RedisMessage) string() string {
```
Is there any performance downside to using a value receiver on this function? Can this function be inlined?

message.go
```diff
@@ -523,24 +531,51 @@ func (r *prettyRedisResult) MarshalJSON() ([]byte, error) {
...
-	attrs   *RedisMessage
-	string  string
-	values  []RedisMessage
+	attrs *RedisMessage
+	bytes *byte
+	array *RedisMessage
+
+	// integer is used for a simple number or
+	// in conjunction with array or bytes to store the length of array or string
 	integer int64
```
rueian, line 540:
```suggestion
	intlen int64
```
Can we change the name of the integer field as well? I am afraid there are some accesses to the integer field that should also be changed.

message.go
```diff
@@ -523,24 +531,51 @@ func (r *prettyRedisResult) MarshalJSON() ([]byte, error) {
...
 }
 
+func (m *RedisMessage) string() string {
+	if m.bytes == nil {
+		return ""
+	}
+	return unsafe.String(m.bytes, m.integer)
+}
+
+func (m *RedisMessage) values() []RedisMessage {
```
rueian, line 552:
```suggestion
func (m RedisMessage) values() []RedisMessage {
```
Is there any performance downside to using a value receiver on this function? Can this function be inlined?

message.go
fscnick: Both of value receiver and pointer receiver could be inlined.
Pointer receiver:
```
./message.go:56:6: can inline (*RedisError).string with cost 13 as: method(*RedisError) func() string { if m.bytes == nil { return "" }; return <node unsafe.String> }
```

Value receiver:
```
./message.go:56:6: can inline RedisError.string with cost 13 as: method(RedisError) func() string { if m.bytes == nil { return "" }; return <node unsafe.String> }
```

Additionally, it seems have around 25% improve on benchmarking for a pointer receiver. Attached the benchmark code for reference.
```
BenchmarkValueReceiver
BenchmarkValueReceiver-4     	493528951	         2.412 ns/op	       0 B/op	       0 allocs/op
BenchmarkPointerReceiver
BenchmarkPointerReceiver-4   	690465366	         1.777 ns/op	       0 B/op	       0 allocs/op
```
[message_bench_test.go.txt](https://github.com/user-attachments/files/19407155/message_bench_test.go.txt)

message.go
```diff
@@ -681,13 +715,13 @@ func (m *RedisMessage) Error() error {
 // ToString check if message is a redis string response, and return it
 func (m *RedisMessage) ToString() (val string, err error) {
 	if m.IsString() {
-		return m.string, nil
+		return m.string(), nil
 	}
-	if m.IsInt64() || m.values != nil {
+	if m.IsInt64() || m.values() != nil {
```
rueian, line 720:
```suggestion
	if m.IsInt64() || m.array != nil {
```

message.go
```diff
@@ -1281,30 +1315,30 @@ func (m *RedisMessage) AsGeosearch() ([]GeoLocation, error) {
...
 			//hash
 			if i < len(info) && info[i].IsInt64() {
-				loc.GeoHash = info[i].integer
+				loc.GeoHash = info[i].intlen
 				i++
 			}
 			//coordinates
-			if i < len(info) && info[i].values != nil {
-				cord := info[i].values
+			if i < len(info) && info[i].values() != nil {
```
rueian, line 1340:
```suggestion
			if i < len(info) && info[i].array != nil {
```

resp.go
```diff
@@ -136,43 +139,43 @@ func readArray(i *bufio.Reader) (m RedisMessage, err error) {
...
 		if peek, _ = i.Peek(4); string(peek) == okrn {
 			_, _ = i.Discard(4)
-			return ok, nil
+			return unsafe.StringData(ok), int64(len(ok)), nil
 		}
 	}
 	bs, err := i.ReadBytes('\n')
 	if err != nil {
-		return "", err
+		return unsafe.StringData(""), 0, err
```
rueian, line 171:
```suggestion
		return nil, 0, err
```

resp.go
```diff
@@ -136,43 +139,43 @@ func readArray(i *bufio.Reader) (m RedisMessage, err error) {
...
 	}
 	if trim := len(bs) - 2; trim < 0 {
-		return "", errors.New(unexpectedNoCRLF)
+		return unsafe.StringData(""), 0, errors.New(unexpectedNoCRLF)
```
rueian, line 174:
```suggestion
		return nil, 0, errors.New(unexpectedNoCRLF)
```

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
 	return v * s, nil
 }
 
-func readB(i *bufio.Reader) (string, error) {
+func readB(i *bufio.Reader) (*byte, int64, error) {
 	length, err := readI(i)
 	if err != nil {
-		return "", err
+		return unsafe.StringData(""), 0, err
```
rueian, line 210:
```suggestion
		return nil, 0, err
```

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
...
 	}
 	if length == -1 {
-		return "", errOldNull
+		return unsafe.StringData(""), 0, errOldNull
```
rueian, line 213:
```suggestion
		return nil, 0, errOldNull
```

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
...
 	}
 	bs := make([]byte, length)
 	if _, err = io.ReadFull(i, bs); err != nil {
-		return "", err
+		return unsafe.StringData(""), 0, err
```
rueian, line 217:
```suggestion
		return nil, 0, err
```

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
...
 	}
 	if _, err = i.Discard(2); err != nil {
-		return "", err
+		return unsafe.StringData(""), 0, err
```
rueian, line 220:
```suggestion
		return nil, 0, err
```

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
...
 	}
-	return BinaryString(bs), nil
+	return unsafe.SliceData(bs), int64(len(bs)), nil
 }
 
-func readE(i *bufio.Reader) ([]RedisMessage, error) {
-	v := make([]RedisMessage, 0)
+func readE(i *bufio.Reader) (*RedisMessage, int64, error) {
+	v := unsafe.SliceData(make([]RedisMessage, 0))
```
rueian, line 226: Can we keep `v := make([]RedisMessage, 0)` and only extract the SliceData when returning?

resp.go
```diff
@@ -201,46 +204,48 @@ func readI(i *bufio.Reader) (v int64, err error) {
...
-		v = append(v, n)
+		v = unsafe.SliceData(append(unsafe.Slice(v, length), n))
+		length++
 	}
 }
 
-func readA(i *bufio.Reader, length int64) (v []RedisMessage, err error) {
-	v = make([]RedisMessage, length)
+func readA(i *bufio.Reader, length int64) (v *RedisMessage, l int64, err error) {
+	v = unsafe.SliceData(make([]RedisMessage, length))
```
rueian, line 242: Can we keep v := make([]RedisMessage, length) and only extract the SliceData when returning?

rueian:
@fscnick Thanks so much for your huge work!

PR #813, opened 2025-03-26
feat: Create and aliax index

SoulPancake:


om/json.go
```diff
@@ -145,6 +145,31 @@ func (r *JSONRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+		return err
+	}
+
+	// Drop the old index
+	if err := r.DropIndex(ctx); err != nil {
+		return err
+	}
+
+	// Update the repository index name
+	r.idx = newIndex
```
rueian, line 168: It seems like a user is required to always invoke `CreateAndAliasIndex` at startup for `Search` to work correctly. Furthermore, a user can't invoke `CreateAndAliasIndex` more than once during the lifetime of a repository; otherwise, the index and alias names will be appended the `_new` and `_alias` suffix multiple times.

Can we allow the following scenario?
1. `CreateAndAliasIndex` is not required at every startup.
2. Allow multiple `CreateAndAliasIndex` invocations during the same lifetime of a repository.

SoulPancake: Correct, take a look at it once, I am checking it using FT LIST

SoulPancake: @rueian

om/hash.go
```diff
@@ -146,6 +148,61 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+
+	aliasExists := true
+	if err := r.client.Do(ctx, r.client.B().FtInfo().Index(alias).Build()).Error(); err != nil {
+		aliasExists = false
+	}
+
+	var currentIndex string
+	if aliasExists {
+		infoCmd := r.client.B().FtInfo().Index(alias).Build()
+		infoResp, err := r.client.Do(ctx, infoCmd).ToMap()
```
rueian, line 163: Can we do FT.INFO only once?

om/hash.go
```diff
@@ -146,6 +148,61 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+		if err != nil {
+			return err
+		}
+		message := infoResp["index_name"]
+		currentIndex = message.String()
```
rueian, line 168:
```suggestion
		currentIndex, err = message.ToString()
```

om/hash.go
```diff
@@ -146,6 +148,61 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+	}
+
+	newIndex := r.idx + "_v1"
+	if aliasExists {
+		parts := strings.Split(currentIndex, "_v")
```
rueian, line 173: what if there are multiple `_v`?

om/hash.go
```diff
@@ -146,6 +148,61 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+		if len(parts) == 2 {
+			if version, err := strconv.Atoi(parts[1]); err == nil {
```
rueian, line 175: what if err != nil?

om/hash.go
```diff
@@ -146,6 +148,61 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+		}
+	}
+
+	if aliasExists && currentIndex != "" {
+		if err := r.client.Do(ctx, r.client.B().FtDropindex().Index(currentIndex).Build()).Error(); err != nil {
+			return err
+		}
+	}
+
+	r.idx = newIndex
```
rueian, line 201: We should not change `r.idx` here. Actually, `r.idx` should be the alias. That is the point of an alias, isn't it?

Same as the above:
```go
alias := r.idx + "_alias"
```
should be just:
```go
alias := r.idx
```

rueian:
Hi @SoulPancake, I just reviewed it again and left some comments. Could you also write some tests to demonstrate how a user can use this feature?

om/hash.go
```diff
@@ -146,6 +148,66 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+
+		currentIndex, err = message.ToString()
+		if err != nil {
+			return fmt.Errorf("failed to convert index_name to string: %w", err)
+		}
+	}
+
+	newIndex := alias + "_v1"
+	if aliasExists && currentIndex != "" {
+		// Find the last occurrence of "_v" followed by digits
```
SoulPancake, line 174: What do you think about this? @rueian

om/json.go
```diff
@@ -145,6 +146,55 @@ func (r *JSONRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
 	return r.client.Do(ctx, cmdFn(r.client.B().FtCreate().Index(r.idx).OnJson().Prefix(1).Prefix(r.prefix+":").Schema())).Error()
 }
 
+// CreateAndAliasIndex creates a new index, aliases it, and drops the old index if needed.
+func (r *JSONRepository[T]) CreateAndAliasIndex(ctx context.Context, cmdFn func(schema FtCreateSchema) rueidis.Completed) error {
+	alias := r.idx + "_alias"
```
rueian, line 151:
```suggestion
	alias := r.idx
```

om/hash.go
```diff
@@ -146,6 +148,66 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
 }
 
+// CreateAndAliasIndex creates a new index, aliases it, and drops the old index if needed.
+func (r *HashRepository[T]) CreateAndAliasIndex(ctx context.Context, cmdFn func(schema FtCreateSchema) rueidis.Completed) error {
+	alias := r.idx
+
+	var currentIndex string
+	infoCmd := r.client.B().FtInfo().Index(alias).Build()
+	infoResp, err := r.client.Do(ctx, infoCmd).ToMap()
+	aliasExists := err == nil
```
rueian, line 158: What will be the error returned when the index does not exist? Will it be a redis Nil? I think we should return early here if the err is caused by other reasons, such as network failures.

om/hash.go
```diff
@@ -146,6 +148,77 @@ func (r *HashRepository[T]) CreateIndex(ctx context.Context, cmdFn func(schema F
...
+// CreateAndAliasIndex creates a new index, aliases it, and drops the old index if needed.
+func (r *HashRepository[T]) CreateAndAliasIndex(ctx context.Context, cmdFn func(schema FtCreateSchema) rueidis.Completed) error {
+	alias := r.idx
+
+	var currentIndex string
+	aliasExists := false
+	infoCmd := r.client.B().FtInfo().Index(alias).Build()
+	infoResp, err := r.client.Do(ctx, infoCmd).ToMap()
+	if err != nil {
+		if strings.Contains(err.Error(), "Unknown index name") {
```
SoulPancake, line 160: I think this should work @rueian

rueian: ok. please write tests for that.

SoulPancake: Hey @rueian Sorry been busy with work, Will do this tomorrow

SoulPancake:
Hi @rueian Apologies for the huge delay
Can you take a look at this?

SoulPancake:
Hi @rueian Any chance to review this?

SoulPancake:
Thanks @rueian , No worries

Issue #770, opened 2025-02-20
rueidiscompat: Should sync commit `fix(aggregate, search): ft.aggregate bugfixes` from go-redis


unknowntpo:
This commit from go-redis contains some changes in `FTAggregate` and `FTSearch`, we should sync them to rueidiscompat.  

Reference: 

- https://github.com/redis/go-redis/pull/3263
- https://github.com/redis/go-redis/commit/7e517ec4a15eb22c60591a96e8a908c0dae40a3b

unknowntpo:
I wanna take this issue.

unknowntpo:
@rueian Sorry, I'm quite busy until the end of March, maybe someone can take this issue.

arbha1erao:
Hi @rueian , I'd like to work on this

PR #815, opened 2025-03-29
sync FT.AGGREGATE

arbha1erao:
fixes: https://github.com/redis/rueidis/issues/770


internal/cmds/gen_search.go
```diff
@@ -17,6 +17,58 @@ func (c FtAggregate) Index(index string) FtAggregateIndex {
 	return (FtAggregateIndex)(c)
 }
 
+type FtAggregateScorer Incomplete
```
rueian, line 20: The `internal/cmds/gen_search.go` file is auto-generated from `hack/cmds/commands_search.json` by running `go generate`. Did you make changes to `hack/cmds/commands_search.json`?

arbha1erao: I've now updated hack/cmds/commands_search.json and regenerated the code.

rueidiscompat/command.go
```diff
@@ -3466,6 +3466,8 @@ type FTAggregateOptions struct {
 	Verbatim          bool
 	LoadAll           bool
 	WithCursor        bool
+	Scorer            string
```
rueian, line 3469: Almost looks good to me, but this string field should be moved up to the next position after `Filter`.

arbha1erao: Addressed

PR #821, opened 2025-04-08
fix: syntax of the MAXAGE option in the CLIENT KILL command is aligned with the redis docs

aakosarev:
According to the redis documentation, the CLIENT KILL command has the following syntax:

`
CLIENT KILL <ip:port | <[ID client-id] | [TYPE <NORMAL | MASTER |
  SLAVE | REPLICA | PUBSUB>] | [USER username] | [ADDR ip:port] |
  [LADDR ip:port] | [SKIPME <YES | NO>] | [MAXAGE maxage]
  [[ID client-id] | [TYPE <NORMAL | MASTER | SLAVE | REPLICA |
  PUBSUB>] | [USER username] | [ADDR ip:port] | [LADDR ip:port] |
  [SKIPME <YES | NO>] | [MAXAGE maxage] ...]>>
`

Currently in the builder in the function `.Maxage(maxage int64)` the keyword "MAXAGE" is not formed.

Because of this, redis returns the error "ERR syntax error" in the new format of the KILL command and the error "ERR No such client" in the old format.

### DIFF
--- a/hack/cmds/commands.json
+++ b/hack/cmds/commands.json
@@ -725,9 +725,9 @@
         "optional": true
       },
       {
-        "name": "MAXAGE",
+        "command": "MAXAGE",
+        "name": "maxage",
         "type": "integer",
-        "token": "MAXAGE",
         "optional": true
       }
     ],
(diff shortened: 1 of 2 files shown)


rueian:
Thanks @aakosarev!

Issue #822, opened 2025-04-08
Add Vector Sets commands json


rueian:
Add the following commands to a new json file in https://github.com/redis/rueidis/tree/main/hack/cmds and re-generate the command builder.

- [ ] VADD
- [ ] VCARD
- [ ] VDIM
- [ ] VEMB
- [ ] VGETATTR
- [ ] VINFO
- [ ] VLINKS
- [ ] VRANDMEMBER
- [ ] VREM
- [ ] VSETATTR
- [ ] VSIM

Ref: https://redis.io/docs/latest/commands/?group=vector_set

dalaoqi:
Hello, @rueian 
Can I take this?

rueian:
Welcome. Let me know if you have any questions.

PR #823, opened 2025-04-09
feat: vectorsets commands json

SoulPancake:
Solves #822

### DIFF
--- a/hack/cmds/commands_vectorsets.json
+++ b/hack/cmds/commands_vectorsets.json
@@ -1,236 +1,254 @@
 {
-  "commands": [
-    {
-      "name": "VADD",
-      "summary": "Adds an element to a vector set, creating a new set if it doesn't already exist.",
-      "since": "Redis CE 8.0.0",
-      "group": "vector_set",
-      "complexity": "O(log(N)) for each element added, where N is the number of elements in the vector set.",
-      "arguments": [
-        { "name": "key", "type": "string" },
-        { "name": "REDUCE dim", "type": "integer", "optional": true },
-        { "name": "FP32 | VALUES num", "type": "string" },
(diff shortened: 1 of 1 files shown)


SoulPancake:
When running gen.go I was getting a panic in checkAllUsed
irrespective of my changes so I commented it out and ran it and it ran
LMK if there's anything else needed to be done, I don't think anything got generated  @rueian

rueian:
Hi @SoulPancake, I guess that was because your json is not in the form like others. Where did you get the json?

SoulPancake:
I think it is the correct form now, however it still panics 
and it panics even without my changes
Can u give it a try? @rueian 
The usage doc string is a little confusing

```
panic: unsued tag subscribe in noRetCMDs

goroutine 1 [running]:
main.checkAllUsed(...)
        /Users/anuragbandyopadhyay/rueidis/hack/cmds/gen.go:654
main.main()
        /Users/anuragbandyopadhyay/rueidis/hack/cmds/gen.go:463 +0x9dc
exit status 2
```

rueian:
> and it panics even without my changes

Really? It works on my side. no panic.

hack/cmds/commands_vectorsets.json
```diff
@@ -0,0 +1,254 @@
...
+      },
+      {
+        "name": "num",
+        "type": "integer",
+        "optional": true
+      },
+      {
+        "name": "vector",
+        "type": "string"
+      },
```
rueian, line 32: These don't seem correct. You should not allow "FP32" to be followed by "num".

Check this out:
image

rueian:
Hi @SoulPancake, this is already completed by https://github.com/redis/rueidis/pull/827

PR #830, opened 2025-04-16
feat: add a span around dialing connections

steved:
If there's some hesitation around creating more disconnected spans, I do think it'd still be useful to only create the span when there's a (rueidis) parent.

I took the span name from [go-redis](https://github.com/redis/go-redis/blob/v9.7.3/extra/redisotel/tracing.go#L94), but perhaps `CONNECT` might be more appropriate?

I also had some trouble testing that the dial span would be parented to the command appropriately since the pool isn't accessible.

### DIFF
--- a/rueidisotel/metrics.go
+++ b/rueidisotel/metrics.go
@@ -10,6 +10,7 @@ import (
 	"github.com/redis/rueidis"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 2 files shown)


rueian (approved):
LGTM

rueian:
> I also had some trouble testing that the dial span would be parented to the command appropriately since the pool isn't accessible.

The span has no chance of being associated with the following command unless we modify the DialCtxFn signature to return the new `ctx` as well. However, I believe the association isn’t particularly useful overall because even after changing the signature, the span will still not be associated with the following command most of the time since the connection will be reused.

PR #833, opened 2025-04-23
Add retry logic of connection lifetime to cluster client and sentinel client

terut:
This is follow-up PR regarding #727 .

See https://github.com/redis/rueidis/pull/727#issuecomment-2823047261.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -523,9 +523,10 @@ process:
 	switch addr, mode := c.shouldRefreshRetry(resp.Error(), ctx); mode {
 	case RedirectMove:
 		ncc := c.redirectOrNew(addr, cc, cmd.Slot(), mode)
+	recover:
 		resp = ncc.Do(ctx, cmd)
 		if resp.Error() == errConnExpired {
-			goto retry
+			goto recover
 		}
 		goto process
 	case RedirectAsk:
@@ -861,9 +862,11 @@ retry:
 process:
 	switch addr, mode := c.shouldRefreshRetry(resp.Error(), ctx); mode {
 	case RedirectMove:
-		resp = c.redirectOrNew(addr, cc, cmd.Slot(), mode).DoCache(ctx, cmd, ttl)
+		ncc := c.redirectOrNew(addr, cc, cmd.Slot(), mode)
+	recover:
+		resp = ncc.DoCache(ctx, cmd, ttl)
 		if resp.Error() == errConnExpired {
-			goto retry
+			goto recover
 		}
 		goto process
 	case RedirectAsk:
--- a/cluster.go
+++ b/cluster.go
@@ -774,10 +774,21 @@ func (c *clusterClient) doretry(
 			var ml []Completed
 		recover:
 			ml = ml[:0]
+			var txIdx int // check transaction block, if zero then not in transaction
 			for i, resp := range resps.s {
 				if resp.Error() == errConnExpired {
-					ml = re.commands[i:]
+					if txIdx > 0 {
+						ml = re.commands[txIdx:]
+					} else {
+						ml = re.commands[i:]
+					}
 					break
+				}
+				// if no error then check if transaction block
+				if isMulti(re.commands[i]) {
+					txIdx = i
+				} else if isExec(re.commands[i]) {
+					txIdx = 0
 				}
 			}
 			if len(ml) > 0 {
@@ -932,13 +943,13 @@ func (c *clusterClient) askingMulti(cc conn, ctx context.Context, multi []Comple
 		var ml []Completed
 	recover:
 		ml = ml[:0]
+		var askingIdx int
 		for i, resp := range resps.s {
+			if commands[i] == cmds.AskingCmd {
+				askingIdx = i
+			}
 			if resp.Error() == errConnExpired {
-				if i > 0 && commands[i-1] == cmds.AskingCmd {
-					ml = commands[i-1:]
-				} else {
-					ml = commands[i:]
-				}
+				ml = commands[askingIdx:]
 				break
 			}
 		}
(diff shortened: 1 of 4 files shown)


terut:
@rueian Sure, I'll add the testing. I’m not familiar to cluster and sentinel client. Is my understanding correct?

rueian:
Yes, I think you are on the right track.

cluster.go
```diff
@@ -514,13 +516,21 @@ retry:
...
+		goto retry
+	}
 process:
 	switch addr, mode := c.shouldRefreshRetry(resp.Error(), ctx); mode {
 	case RedirectMove:
-		resp = c.redirectOrNew(addr, cc, cmd.Slot(), mode).Do(ctx, cmd)
+		ncc := c.redirectOrNew(addr, cc, cmd.Slot(), mode)
+		resp = ncc.Do(ctx, cmd)
+		if resp.Error() == errConnExpired {
+			goto retry
```
rueian, line 528: Shouldn't this go back to L526?

cluster.go
```diff
@@ -514,13 +516,21 @@ retry:
...
+		}
 		goto process
 	case RedirectAsk:
-		results := c.redirectOrNew(addr, cc, cmd.Slot(), mode).DoMulti(ctx, cmds.AskingCmd, cmd)
+		ncc := c.redirectOrNew(addr, cc, cmd.Slot(), mode)
+		results := c.doMulti(ctx, ncc, cmds.AskingCmd, cmd)
```
rueian, line 533: `ASKING` can't be handled this way. It must be sent in every retry.

cluster.go
terut: Fix https://github.com/redis/rueidis/pull/833/commits/3623adcf14c62b50c88053c1f64bc5bcef126ca8 . Is it correct?

cluster.go
terut: @rueian Does it mean that `ASKING` needs at head of cmds when retrying due to `errConnExpired`

rueian: Yes, `ASKING` is only effective in the same connection and applied only to the next command; therefore, we need to re-send `ASKING` before the target `cmd` if `errConnExpired` happens.

terut: I see. I will also fix `askingMulti` and `askingMultiCache` as same.

terut: @rueian Fix [ce19729](https://github.com/redis/rueidis/pull/833/commits/ce197291cd94c3c6c61358d19a4ca131acf25592) [be811cc](https://github.com/redis/rueidis/pull/833/commits/be811cc271cf6556292e0ab81606c978f7f3fb7d)

cluster.go
```diff
@@ -889,6 +928,26 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults
...
+	if c.hasLftm {
+		var ml []Completed
+	recover:
+		ml = ml[:0]
+		for i, resp := range resps.s {
+			if resp.Error() == errConnExpired {
+				if i > 0 && commands[i-1] == cmds.AskingCmd {
+					ml = commands[i-1:]
+				} else {
+					ml = commands[i:]
```
rueian, line 940: This reminds me that we need to handle the `MULTI EXEC` pattern correctly: If `errConnExpired` occurs during the `MULTI EXEC`, we should retry the entire `MULTI EXEC` block. This applies to the single client as well, which we overlooked in the previous PR.

cluster.go
```diff
@@ -898,14 +957,30 @@ func askingMulti(cc conn, ctx context.Context, multi []Completed) *redisresults
...
 		ck, _ := cmds.CacheKey(cmd.Cmd)
 		commands = append(commands, cc.OptInCmd(), cmds.AskingCmd, cmds.MultiCmd, cmds.NewCompleted([]string{"PTTL", ck}), Completed(cmd.Cmd), cmds.ExecCmd)
 	}
 	results := resultsp.Get(0, len(multi))
 	resps := cc.DoMulti(ctx, commands...)
+	if c.hasLftm {
+		var ml []Completed
+	recover:
+		ml = ml[:0]
+		for i, resp := range resps.s {
```
rueian, line 972: Here, we can advance the `i` by 6 in every iteration and only check the Error of the ExecCmd.

cluster.go
terut: Would you check this pull-request https://github.com/redis/rueidis/pull/837 . This is draft so I will fix other methods and add tests after checking the implement.

terut: @rueian Would you check this pull-request https://github.com/redis/rueidis/pull/837 ? This is draft so I will fix other methods and add tests after checking the implement.

rueian: Yes, the draft looks good to me.

terut: Thanks!

terut: @rueian Please review #837 .

cluster.go
terut: Fix https://github.com/redis/rueidis/pull/833/commits/6362824f8c826e8c96313ea41fb92eb51eddf7c6

cluster.go
terut: I'm writing tests for `askingMulti()` but I'm not sure why multi has the following commands when calling that method. Any idea?

```
// Call for unit testing
multi := []Completed{
	client.B().Get().Key("1{t}").Build(),
	client.B().Multi().Build(),
	client.B().Incr().Key("2{t}").Build(),
	client.B().Incr().Key("3{t}").Build(),
	client.B().Exec().Build(),
	client.B().Get().Key("4{t}").Build(),
}
resps := client.DoMulti(context.Background(), multi...)

// multi in askingMulti(). I don't know why there is no [GET 2{t}] [Get 3{t}].
[ASKING]
[GET 1{t}]
[ASKING]
[MULTI]
[EXEC]
[ASKING]
[GET 4{t}]
```

terut: Fix https://github.com/redis/rueidis/pull/833/commits/6362824f8c826e8c96313ea41fb92eb51eddf7c6

rueian (approved):

rueian:
Thanks @terut!

terut:
@rueian Thank you for great help for couple of months :smile:

PR #849, opened 2025-06-06
Update package name in README example

hono0130:
Changed `valkey.Commands` to `rueidis.Commands` in the pinned commands example.


rueian:
Thanks @hono0130 !

PR #851, opened 2025-06-08
refactor: changed atomic.Value to atomic.Pointer in pipe

### LINKED ISSUE #847: Replace atomic.Value with atomic.Pointer
We have this
https://github.com/redis/rueidis/blob/0244b0fbdca15e8587140e0f04089079c3ea3a75/pipe.go#L67

It will be better to replace it with `atomic.Pointer[pshks]`.

nithinputhenveettil:
fixes #847

Tests executed successfully in the local env.

image

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -63,8 +63,8 @@ var _ wire = (*pipe)(nil)
 
 type pipe struct {
 	conn            net.Conn
-	clhks           atomic.Value // closed hook, invoked after the conn is closed
-	pshks           atomic.Value // pubsub hook, registered by the SetPubSubHooks
+	clhks           atomic.Value          // closed hook, invoked after the conn is closed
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -63,8 +63,8 @@ var _ wire = (*pipe)(nil)
 
 type pipe struct {
 	conn            net.Conn
-	clhks           atomic.Value // closed hook, invoked after the conn is closed
-	pshks           atomic.Value // pubsub hook, registered by the SetPubSubHooks
+	clhks           atomic.Value          // closed hook, invoked after the conn is closed
+	pshks           atomic.Pointer[pshks] // pubsub hook, registered by the SetPubSubHooks
```
rueian, line 67: Hi @nithinputhenveettil,

Could you also sync the change to https://github.com/redis/rueidis/blob/51c2ebfd3f20c292150905b91644ea96295bff23/mock/result.go#L190 to fix the CI failures?

PR #864, opened 2025-07-02
perf: optimize `RedisMessage.DecodeJSON()`

ash2k:
`json.Decoder()` copies data from the reader into a temporary buffer. This is an extra copy that is not necessary since we have the data as a slice already.


rueian (approved):

rueian:
Thanks @ash2k!

PR #894, opened 2025-08-16
feat: added support for route randomly flag

ayush-sarda:


rueidis.go
```diff
@@ -242,6 +242,12 @@ type ClientOption struct {
 	// EnableReplicaAZInfo enables the client to load the replica node's availability zone.
 	// If true, the client will set the `AZ` field in `ReplicaInfo`.
 	EnableReplicaAZInfo bool
+
+	// RouteRandomly routes read only command's traffic randomly between master and replica nodes.
+	// Enabling it alongside client side caching can result in higher memory usage due to 
+	// individual cache store against each connection.
+	// Only works when SendToReplicas return true
+	RouteRandomly bool
```
rueian, line 250: I think we don't need the new `RouteRandomly` option. We can just always invoke ReplicaSelector to pick a replica when necessary.

cluster.go
```diff
@@ -279,19 +282,34 @@ func (c *clusterClient) _refresh() (err error) {
 				for _, slot := range g.slots {
 					for i := slot[0]; i <= slot[1] && i >= 0 && i < 16384; i++ {
 						pslots[i] = conns[master].conn
-						rIndex := c.opt.ReplicaSelector(uint16(i), g.nodes[1:])
-						if rIndex >= 0 && rIndex < n {
-							rslots[i] = conns[g.nodes[1+rIndex].Addr].conn
+						var replicas nodes
+						if c.opt.RouteRandomly {
+							for _, node := range g.nodes {
```
rueian, line 287:
```suggestion
							for _, node := range g.nodes[1:] {  // exclude master node
```

rueidis.go
ayush-sarda: The idea I had in my mind was, user should be in control if they want to route traffic across all the nodes (including master). If this flag is turned off then distribution will happen as it was happening earlier.

ayush-sarda: If you want to make it default, then we can go with that approach. But we'd still need some flag to specify if master node should be included in distribution or not.

cluster.go
rueian: I am ok with that.

cluster.go
```diff
@@ -67,13 +73,21 @@ func newClusterClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*
 	if opt.ReplicaOnly && opt.ReplicaSelector != nil {
 		return nil, ErrReplicaOnlyConflictWithReplicaSelector
 	}
+	if opt.ReplicaOnly && opt.RouteRandomly {
```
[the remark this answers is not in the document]
ayush-sarda: got it. will change this.

rueian:
Thank you @ayush-sarda!

PR #898, opened 2025-09-14
fix: correctly handle early exit in `Scanner` iterators

ash2k:
Both inner and outer `for` must be exited when `yield` returns `false`.


rueian (approved):

rueian:
Thank you @ash2k

ash2k:
@rueian No problem, thanks for merging. Could you make a release please? I'm working on a refactoring (in my project) and that's how I found the bug. Now I need the fix to continue working on it :)

rueian:
Yes, I will release a new version this week!

PR #905, opened 2025-09-27
test: polish TestForceSingleClientInitialDialError

justinmayhew:
Removes the need to wait 100ms for the first dial to time out by
returning an error in DialCtxFn, adds more precise checks on the return
value from NewClient, and ensures there are exactly 2 Dial calls.


rueian (approved):

rueian:
Thank you @justinmayhew!

Issue #844, opened 2025-05-20
Question about MGet helper with cluster client


fabiomargarido:
Hey there.

I'm a bit confused by the wording in the comment and README regarding the MGet helper functions:

```
// MGet is a helper that consults the redis directly with multiple keys by grouping keys within the same slot into MGET or multiple GETs
func MGet(client Client, ctx context.Context, keys []string) (ret map[string]RedisMessage, err error) {
```

Specifically, the part where it says keys are grouped taking the slot into consideration, since looking at the code it seems it simply checks if the client is not a cluster client, in which case an `MGET` is sent, and falls back to a pipelined GET for each separate key when it is in fact a cluster.

I was wondering if it would be possible to pre-calculate the target slot for each key and send a separate MGET command to each node responsible the a particular set of slots. If this fails for some reason (e.g. cluster topology changed and refresh was not yet executed), then fallback to the current implementation.

Thanks.

rueian:
Hi @fabiomargarido,

You are right. Currently, the helper will simplify pipeline GETs to the cluster client. I think it is a good idea to group them into multiple MGETs.

Would you like to give it a try? `intl.Slot()` can do the pre calculation.

fabiomargarido:
Thanks for the quick reply. I'm actually a bit more puzzled now, since I found this commit from almost 2 years ago:

https://github.com/redis/rueidis/commit/086a2448e29f50da27dccccd3893be99e0f3b1ce

It seems this is how it worked before, but was changed for performance reasons. Is there anything you think could be done differently now to improve it?

rueian:
Hi @fabiomargarido,

Thanks for bringing up the old commit. Previous to the commit, we did `parallelMGet` which sent `MGET` for each slot parallelly using multiple `DoMulti`, but that was too fine-grained. Therefore, after the commit, we simplify put all `GET`s to one `DoMulti` and let it do pipelining by connections.

I wonder if we can put `MGET`s to one `DoMulti`? For example, if the input of the `MGet` helper is `[]string{"{1}a", "{2}a", "{1}b", "{2}b"}`, then we put `MGET {1}a {1}b` and `MGET {2}a {2}b` to one `DoMulti`.

fabiomargarido:
That could make sense... Furthermore, instead of sending one MGET per slot, wouldn't it make sense to check the slot against the range for each node and send one per node?

rueian:
I want to say it makes sense too but I think multiple slots in one MGET is forbidden by redis server.

[...]

PR #908, opened 2025-10-04
feat : Implement slot-based MGET batching for cluster clients

SoulPancake:
For #844

### DIFF
--- a/helper.go
+++ b/helper.go
@@ -50,12 +50,7 @@ func MGet(client Client, ctx context.Context, keys []string) (ret map[string]Red
 		return clientMGet(client, ctx, client.B().Mget().Key(keys...).Build(), keys)
 	}
 
-	cmds := mgetcmdsp.Get(len(keys), len(keys))
-	defer mgetcmdsp.Put(cmds)
-	for i := range cmds.s {
-		cmds.s[i] = client.B().Get().Key(keys[i]).Build()
-	}
-	return doMultiGet(client, ctx, cmds.s, keys)
+	return clusterMGet(client, ctx, keys)
 }
 
 // MSet is a helper that consults the redis directly with multiple keys by grouping keys within the same slot into MSETs or multiple SETs
@@ -139,12 +134,7 @@ func JsonMGet(client Client, ctx context.Context, keys []string, path string) (r
 		return clientMGet(client, ctx, client.B().JsonMget().Key(keys...).Path(path).Build(), keys)
 	}
 
-	cmds := mgetcmdsp.Get(len(keys), len(keys))
-	defer mgetcmdsp.Put(cmds)
-	for i := range cmds.s {
-		cmds.s[i] = client.B().JsonGet().Key(keys[i]).Path(path).Build()
-	}
-	return doMultiGet(client, ctx, cmds.s, keys)
+	return clusterJsonMGet(client, ctx, keys, path)
 }
 
 // JsonMSet is a helper that consults redis directly with multiple keys by grouping keys within the same slot into JSON.MSETs or multiple JSON.SETs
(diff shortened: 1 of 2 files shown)


SoulPancake:
I optimised the slice mem allocation, is there any other potential improvement I can work on? @rueian

rueian:
> I optimised the slice mem allocation, is there any other potential improvement I can work on? @rueian

I wonder if it is possible to avoid using `MGets` since it allocates a `make(map[uint16]Completed, 8)` though I am not sure if it is allocated on the heap or the stack after compiler optimization. Would you mind checking that? If it is allocated on the stack, then it should be fine.

SoulPancake:
> I wonder if it is possible to avoid using MGets since it allocates a make(map[uint16]Completed, 8) though I am not sure if it is allocated on the heap or the stack after compiler optimization. Would you mind checking that? If it is allocated on the stack, then it should be fine.


@rueian 
make(map[uint16]Completed, 8) in MGets/JsonMGets does escape to heap. 
What do you suggest we do here

rueian:
> What do you suggest we do here

1. Try inline MGets/JsonMGets manually.
2. Then replacing make(map[uint16]Completed, 8) with make(map[uint16]int, 8) where the values are the indexes of cmds.s, eliminating the use of slotCmds.

SoulPancake:
I have some trouble will my personal laptop, I will address these in a week approx, hope that is okay

SoulPancake:
@rueian 
 I inlined the grouping logic:
1. Added CRC16 hashing and slot calculation functions (copied from internal/cmds/slot.go) to helper.go for key-to-slot mapping. Let me know if I should move this to a common util?
2. Replaced the intl.MGets/intl.JsonMGets calls with manual key grouping into a map[uint16][]string.
3. Built the command slice (cmds.s) directly using the client builder (client.B().Mget().Key(group...).Build().Pin() for MGet, and similarly for JsonMGet with path).
4. Maintained a parallel cmdKeys slice to track keys per command for response parsing.
5. Eliminated the intermediate slotCmds map of Completed structs, reducing heap pressure while preserving functionality.

helper.go
```diff
@@ -277,6 +271,69 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
 	return m
 }
 
+func clusterMGet(client Client, ctx context.Context, keys []string) (ret map[string]RedisMessage, err error) {
+	ret = make(map[string]RedisMessage, len(keys))
+	slotGroups := make(map[uint16][]string)
+	for _, key := range keys {
+		ks := slot(key)
+		slotGroups[ks] = append(slotGroups[ks], key)
```
rueian: Is it possible to avoid the string slice allocation here? Maybe use a `map[uint16]int` instead?

SoulPancake:
Hi @rueian Any chance to review this?

helper.go
```diff
@@ -273,18 +273,19 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
...
+	cmds := mgetcmdsp.Get(0, len(slotCount))
 	defer mgetcmdsp.Put(cmds)
-	groups := make([][]string, 0, len(slots))
-	for _, group := range slots {
-		gkeys := make([]string, 0, len(group))
-		for _, i := range group {
-			gkeys = append(gkeys, keys[i])
+	groups := make([][]string, 0, len(slotCount))
+	for s, count := range slotCount {
+		gkeys := make([]string, 0, count)
```
rueian: hmm, ideally, we should avoid `gkeys := make([]string, 0, count)` too. We should only do allocations on `client.B().Mget().Key(...)`

helper.go
```diff
@@ -277,6 +271,99 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
...
+			if slot(key) == s {
+				if first {
+					builder = builder.(intl.Mget).Key(key)
+					first = false
+				} else {
+					builder = builder.(intl.MgetKey).Key(key)
+				}
+			}
+		}
+		cmds.s = append(cmds.s, builder.(intl.MgetKey).Build().Pin())
```
rueian: Hi @SoulPancake,

Right now, you count the number of keys for each slot, but then discard them, and then iterate over the whole keys slice for each slot, which I think is super inefficient.

How about storing the resulting indexes of cmd.s to the map[uint16]int map first? As I said [previously](https://github.com/redis/rueidis/pull/908#issuecomment-3422802256). So that you can use `slot(key)` to locate the index of cmds.s and only iterate the keys slice twice in total.

helper.go
```diff
@@ -277,6 +271,114 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
...
 }
 
+func clusterMGet(client Client, ctx context.Context, keys []string) (ret map[string]RedisMessage, err error) {
+	ret = make(map[string]RedisMessage, len(keys))
+	if len(keys) == 0 {
+		return ret, nil
+	}
+
+	slotIdx := make(map[uint16]int, len(keys)/2)
+	var builders []any
```
rueian:
```suggestion
	cmds := mgetcmdsp.Get(0, len(keys))
	defer mgetcmdsp.Put(cmds)
```

Again, we should avoid additional allocations. This `builders []any` can be replaced by the `cmds.s`

helper.go
```diff
@@ -277,6 +271,134 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
...
+		jp := jm.Path(path)
+
+		intl.PutCompletedForce(c)
+		cmds.s[idx] = jp.Build().Pin()
+	}
+
+	resps := client.DoMulti(ctx, cmds.s...)
+	defer resultsp.Put(&redisresults{s: resps})
+
+	values := make([][]RedisMessage, len(cmds.s))
```
SoulPancake: Hi @rueian 
Sorry I've been caught up with other things and missed focused work on this
Do you think this allocation is still alright?
I cannot think of a better way for this

helper.go
```diff
@@ -277,6 +271,134 @@ func arrayToKV(m map[string]RedisMessage, arr []RedisMessage, keys []string) map
...
+
+		intl.PutCompletedForce(c)
+		cmds.s[idx] = mk.Build().Pin()
+	}
+
+	resps := client.DoMulti(ctx, cmds.s...)
+	defer resultsp.Put(&redisresults{s: resps})
+
+	// Convert each response to an array once
+	values := make([][]RedisMessage, len(cmds.s))
```
SoulPancake: only using:
1. slotIdx map[uint16]int for slot → command index, and
2. the pooled cmds.s slice for the actual Completed commands.

rueian (approved):
Hi @SoulPancake, I have rewritten the implementation. You may take a look.

Issue #923, opened 2025-12-04
rueidislock: long-running locks are lost under high CPU load


arisudesu:
Hello!

I am observing sudden `context canceled` errors in our production systems, using a context which originates from rueidislock, when the lock is held for long periods of time (> 10m).

Code is simple, no fancy things:
```go
ctx, releaseJobLock, err := locker.TryWithContext(context.Background(), lockName)
if err != nil { return }
defer releaseJobLock()

if err := longRunningTask(ctx); err != nil {
    log.Println(err)
}
```

After some investigation and playing with rueidislock code, I've discovered that timers which extends keys' expiration time, can singificantly lag behind schedule when the system has moderate to high CPU load:

https://github.com/redis/rueidis/blob/3238f88c4b317af81e875bc0476507cbc60f78ce/rueidislock/lock.go#L272-L276

This code basically assumes that a `case <-timer.C` runs exactly after `m.interval`, but in fact it will run after `m.interval` plus some margin before `case <-timer.C` gets scheduled, which is non-zero on loaded system. Therefore, it causes key extension to lag and eventually keys may expire before next extension runs, which in turn causes context cancellation.

With some modifications for logging, I observed that even little single delays (1 to 10ms) can sum up and exceed m.interval in long runs.

Timer reset should probably account for scheduling delay, by subtracting it from interval, to match deadline increments, e.g.:
```go
case timerat <-timer.C:
	delay := time.Now().Sub(timerat) // how much it spent in go scheduler before actually running
	deadline = deadline.Add(m.interval)
	if err = m.script(ctx, extend, key, val, deadline); err == nil {
		timer.Reset(m.interval - delay) // subtract delay in attempt to synchronize with deadline increment
	}
```

I tested it on go1.25.1 rueidis 1.0.38. However I see that the problematic code is unchanged between 1.0.38 and 1.0.68, and may have the same issue.

rueian:
Hi @arisudesu, it seems like you already have a fix. Would you like to create a PR?

arisudesu:
@rueian I am testing the fix to make sure that it is indeed solving the issue. But I'm thinking that time.NewTimer() too should account for time difference, and both time.NewTimer() and timer.Reset() should instead try not to measure and subtract delay, but even to run at `deadline - m.interval` time. What do you think?

rueian:
[...]

PR #924, opened 2025-12-08
Ensure that monitored lock keys expiration is always ahead of current time by KeyValidity duration

arisudesu:
Ref issue #923


rueian:
Thank you @arisudesu!

Issue #925, opened 2025-12-10
Feature Request: LuaScript and rueidislock support Retryable


wyxloading:
Related to #919 

- Some lua script (not readonly) may also retryable ( for example the whole script represent a idempotent operation.
- The lua scripts in rueidislock mod can also considered idempotent operations.

For `Lua` add `LuaOption` for setting this lua script retryable.  
For all lua scripts in rueidislock mod, set retryable by default.

rueian:
Thanks @wyxloading!

PR #926, opened 2025-12-10
Feat: lua script support retryable; rueidislock set all lua scripts retryable by default

wyxloading:
For #925 

- For Lua add LuaOption for setting this lua script retryable.
  - add testcases to verify that `WithRetryable(true)` option make lua script commands Retryable
- For all lua scripts in rueidislock mod, set retryable by default.

### DIFF
--- a/lua.go
+++ b/lua.go
@@ -105,7 +105,7 @@ func (s *Lua) Exec(ctx context.Context, c Client, keys, args []string) (resp Red
 	var scriptSha1 string
 
 	// Determine which SHA-1 to use.
-	if !s.noSha1 && s.loadSha1 {
+	if s.loadSha1 {
 		// Check if SHA-1 is already loaded.
 		s.sha1Mu.RLock()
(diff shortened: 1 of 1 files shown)


lua.go
```diff
@@ -24,6 +24,14 @@ func WithLoadSHA1(enabled bool) LuaOption {
 	}
 }
 
+// WithRetryable allows making lua script (which not readOnly) execution retryable. It would retry
+// automatically under network issues or redis retryable issues. It works with Completed.ToRetryable.
+func WithRetryable(enabled bool) LuaOption {
```
rueian: While `WithRetryable` might look good by itself here, `rueidis.WithRetryable()` looks vague in terms of how it relates to Lua. I prefer not to provide `rueidis.WithRetryable`, but instead to provide a new `NewLuaScriptRetryable` constructor function.

wyxloading: indeed. feel weird when writing testcases

PR #930, opened 2025-12-26
feat: add availability zone awareness for standalone client

nesty92:


nesty92:
@rueian, would love to get your thoughts on this PR.

Quick question: should we also be updating the nodes when the redirect is active? Not 100% sure on that part.

rueian:
Hi @nesty92, thanks for the PR. Would you mind splitting the helpers to another PR? And the index 0 of the node slice may not be the primary if the user uses ReplicaSelector instead of ReadNodeSelector. I am also wondering if the helpers can be optimized with bit operations to avoid the second round of zone string comparison.

nesty92:
Hi @rueian, here is the PR for the helpers https://github.com/redis/rueidis/pull/932

PR #932, opened 2026-01-05
add Az awareness helpers

nesty92:
- Add `AZAffinityNodeSelector` to prioritize replicas in the same Availability
  Zone.
- Add `AZAffinityReplicasAndPrimaryNodeSelector` for granular control,
  prioritizing local replicas, then local primaries, before falling back to
  remote nodes.
- Add `PreferReplicaNodeSelector` for explicit routing requirements.
- Update `ClientOption` to use strict function types (`ReadNodeSelectorFunc`)
  for improved type safety.

Depends on https://github.com/redis/rueidis/pull/930

### DIFF
--- a/helper.go
+++ b/helper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"iter"
+	"sync/atomic"
 	"time"
 
 	intl "github.com/redis/rueidis/internal/cmds"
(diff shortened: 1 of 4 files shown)


helper.go
```diff
@@ -391,3 +392,105 @@ func (s *Scanner) Iter2() iter.Seq2[string, string] {
...
+// pickAZ selects a node index from nodes[startIdx:] that matches the clientAZ.
+func pickAZ(nodes []NodeInfo, clientAZ string, startIdx int, counter *atomic.Uint64) int {
+	n := len(nodes)
+	if n <= startIdx {
+		return -1
+	}
+
+	// We cap the search at 255 nodes
+	limit := min(n, 255)
+	var matches [64]uint8
```
rueian: 64 is quite big. Can we do 8 or 16 instead? I can't imagine that someone would have 64 replicas for a slot in the same zone.

PR #944, opened 2026-01-29
fix: validate connection health and handle missing slots in cluster refresh

billshen99:
## Summary
Fixes critical bug where broken connections were blindly reused during cluster 
topology refresh, causing persistent "wrong shard/slot" errors requiring pod restarts.

## Root Cause
During `_refresh()`, connections were reused without health validation. Network 
issues could break connections, which would persist in error state indefinitely 
across refresh cycles.

## Changes
1. **Critical, Connection health validation and reuse:**
 Validate connection health with `Error()` before reusing
* The cluster client now checks the health of existing connections before reusing them during a topology refresh. Connections reporting errors are replaced with new ones, preventing broken connections from being reused.

2. **Slot mapping resilience:**
* When the cluster topology information is incomplete (e.g., missing slots), the client preserves old slot-to-connection mappings for slots not covered by the new topology, ensuring no slot is left without a connection.

3. **Replica slot fallback improvements:**
Add replica-to-master fallback for missing replica slots
* In scenarios where replica slot information is missing, the client now falls back to using the master connection for those slots, ensuring continued availability. [[1]](diffhunk://#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R493-R494) [[2]](diffhunk://#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R654-R655) [[3]](diffhunk://#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R1145-R1146)

## Impact
- Broken connections now auto-detected and replaced within seconds
- No more manual pod restarts needed for this issue
- Graceful degradation during incomplete topology updates
This pull request improves the reliability of the Redis cluster client by ensuring that unhealthy (broken) connections are not reused after a topology refresh, and by making slot mapping more robust when the cluster topology is incomplete. It also adds comprehensive tests to validate these behaviors.

**Testing enhancements:**

* Added tests to verify that broken connections are not reused after a refresh, and that all slots are properly updated to use healthy connections, ensuring robust failover and recovery.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -241,10 +241,18 @@ func (c *clusterClient) _refresh() (err error) {
 	var removes []conn
 
 	c.mu.RLock()
+	// Preserve old slot mappings to use as fallback for incomplete topology
+	oldWslots := c.wslots
+	oldRslots := c.rslots
 	for addr, cc := range c.conns {
 		if fresh, ok := conns[addr]; ok {
-			fresh.conn = cc.conn
-			conns[addr] = fresh
+			// Validate connection health before reusing
+			// If connection has error, it will be replaced with a new one
+			if cc.conn.Error() == nil {
+				fresh.conn = cc.conn
+				conns[addr] = fresh
+			}
+			// else: let the new connection be created (conns[addr] already has fresh conn from connFn)
 		} else {
 			removes = append(removes, cc.conn)
 		}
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -241,10 +241,18 @@ func (c *clusterClient) _refresh() (err error) {
...
+	// Preserve old slot mappings to use as fallback for incomplete topology
+	oldWslots := c.wslots
+	oldRslots := c.rslots
 	for addr, cc := range c.conns {
 		if fresh, ok := conns[addr]; ok {
-			fresh.conn = cc.conn
-			conns[addr] = fresh
+			// Validate connection health before reusing
+			// If connection has error, it will be replaced with a new one
+			if cc.conn.Error() == nil {
```
rueian: `mux` handles reconnections internally, so no Error() check at the cluster client is okay.

PR #974, opened 2026-03-24
fix: add expiration key to rate limit script execution

junsred:
Error:
 
```
{
  "value": "Error running script (call to ...): @user_script:8: script tried accessing undeclared key, key: arl:{...}:ex",
  "key": "exception.message"
}
```

Solution:
Adding generated ex key in lua script to arguments.

---

> [!NOTE]
> **Low Risk**
> Low risk, localized change to Redis Lua script invocation to satisfy key declaration requirements. Main risk is behavior differences if callers/keys are constructed inconsistently, but logic is otherwise unchanged.
> 
> **Overview**
> Fixes Redis Lua script execution by explicitly passing the `:ex` (expires-at) key as `KEYS[2]` instead of deriving it inside the script.
> 
> `AllowN` now constructs `expiresAtKey` alongside the main rate-limit key and executes `rateLimitScript` with both keys, preventing "undeclared key" errors under strict key access modes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c513b1dbb34abe39a6df0da82fe475a1035d777. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rueidislimiter/limiter.go
+++ b/rueidislimiter/limiter.go
@@ -115,6 +115,11 @@ func (l *rateLimiter) AllowN(ctx context.Context, identifier string, n int64, op
 	key := rueidis.BinaryString(bufs.keyBuf[offset:])
 
 	offset = len(bufs.keyBuf)
+	bufs.keyBuf = append(bufs.keyBuf, key...)
+	bufs.keyBuf = append(bufs.keyBuf, ":ex"...)
+	expiresAtKey := rueidis.BinaryString(bufs.keyBuf[offset:])
+
+	offset = len(bufs.keyBuf)
 	bufs.keyBuf = strconv.AppendInt(bufs.keyBuf, n, 10)
 	arg1 := rueidis.BinaryString(bufs.keyBuf[offset:])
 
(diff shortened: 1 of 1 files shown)


rueidislimiter/limiter.go
junsred:
```suggestion
local expires_at_key = KEYS[2]
```

rueian:
Could you add a test?

junsred:
> Could you add a test?

not sure what exactly needs to be tested
BinaryString construction?

rueian:
A test that will result in the error you posted if we don't have your change.

junsred:
This is my fault for not including the reason of the error. The error appearing is due to [Dragonfly as default not allowing undeclared keys in lua scripts due to performance issues on multi-threaded instances](https://github.com/dragonflydb/dragonfly/issues/272#issuecomment-2019767858). 
This is also recommended way to access keys in redis clusters according to redis documents:
`Important: to ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments. The script should only access keys whose names are given as input arguments. Scripts should never access keys with programmatically-generated names or based on the contents of data structures stored in the database.` ([1](https://redis.io/docs/latest/commands/eval/))

junsred:
added the [allow test with dragonfly](https://github.com/redis/rueidis/pull/975) client which verifies the error.

rueian (approved):

PR #1004, opened 2026-06-08
fix: guard parseSlots against malformed CLUSTER SLOTS replies during failover

### LINKED ISSUE #929: Errors and Panic in Rueidis during Redis Cluster Shard Failover
### Description

We observed a set of issues when using rueidis v1.0.69 with a Redis Cluster (v7.4).

During a failover in one shard of the cluster, the application experienced an increase in Redis request latency, followed by multiple client-side errors. No application code changes were deployed during this time.

The following issues were observed:

- Intermittent transaction errors:
```
EXEC was aborted by redis or connection closed
```

- Redis protocol parse errors:
```
rueidis: parse error: redis message type simple string is not a array
rueidis: parse error: redis message type array is not a string
```

- Connection-level errors:
```
EOF
dial tcp <ip>:<port>: operation was canceled
```

Once the parse errors started, the client did not recover and required restarting application pods.

### Panic observed during cluster refresh
In the same time window, we observed a panic originating from the cluster topology refresh path.
```
panic: runtime error: index out of range [2] with length 0

goroutine 258842186 [running]:
github.com/redis/rueidis.parseSlots(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/cluster.go:378
github.com/redis/rueidis.clusterslots.parse(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/cluster.go:378
github.com/redis/rueidis.(*clusterClient)._refresh(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/cluster.go:167
github.com/redis/rueidis.(*call).do(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/cluster.go:217
github.com/redis/rueidis.(*call).LazyDo.func1(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/singleflight.go:58
github.com/redis/rueidis.(*call).LazyDo(...)
    /go/pkg/mod/github.com/redis/rueidis@v1.0.69/singleflight.go:53
```

### Client Code

- Connection dial timeout: default
- Connection write timeout: default

#### Client Initialization

```
client, err := rueidisotel.NewClient(
    rueidis.ClientOption{
        InitAddress:       conf.NodeAddresses,
        Username:          conf.UserName,
        Password:          conf.Password,
        DisableCache:      conf.DisableCache,
        CacheSizeEachConn: cacheSizeEachConnection,
    },
)
if err != nil {
    return err
}

if err := client.Do(
    context.Background(),
    client.B().Ping().Build(),
).Error(); err != nil {
    return err
}
```

#### Client Usage

```
func (c *Cache) MGetWithClientSideCache(
    ctx context.Context,
    keys []string,
) (map[string]string, error) {

    results, err := rueidis.MGetCache(
        c.client,
        ctx,
        30*time.Second,
        keys,
    )
    if err != nil {
        return nil, err
    }

    finalResult := make(map[string]string, len(results))
    for key, result := range results {
        value, err := result.ToString()
        if err != nil && !rueidis.IsRedisNil(err) {
            return nil, err
        }
        finalResult[key] = value
    }

    return finalResult, nil
}
```

### Environment

- Rueidis v1.0.69
- Go 1.23.0
- Redis Cluster v7.4
- Auto-Pipelining enabled

### Questions

1. When does rueidis refresh the cluster topology?
If no refresh interval is configured, what situations cause a topology refresh to happen automatically?

2. Is it normal to see EXEC abort errors during a shard failover?
Should applications expect these errors during cluster changes?

3. What situations can lead to Redis protocol parse errors in rueidis?
For example, can they happen due to connection interruptions or cluster changes?

4. After a protocol parse error, should the client recover on its own or be restarted?

mvanhorn:
## Summary
`parseSlots` and `parseShards` now tolerate malformed `CLUSTER SLOTS` / `CLUSTER SHARDS` replies during shard failover instead of panicking on out-of-range slice access.

## Why this matters
@ashmeet-zepto reported panics in #929 during Redis Cluster shard failover. While the topology is in flux, a `CLUSTER SLOTS` / `CLUSTER SHARDS` reply can arrive with fewer fields than the happy path assumes. The old code indexed `v.values()[2].values()[0]` and `dict["endpoint"]` unconditionally, so a truncated reply triggered an index-out-of-range panic that took down the client during the failover window, exactly when it needs to stay up.

## Changes
- `parseSlots`: bounds-check the slot entry, the master address triple, and each per-node value before indexing; skip entries that are too short rather than panicking.
- `parseShards`: require an `endpoint` field and a usable `port`/`tls-port` before adding a node; skip shards missing required fields.

## Testing
Added `TestClusterSlotsParsing` and `TestClusterShardsParsing` covering truncated/partial replies. `gofmt -l` and `go vet` are clean; the new unit tests pass (`go test -run 'Slots|Shards'` -> ok). The cluster integration suite needs a live multi-node cluster and is unchanged.

Fixes #929

AI was used for assistance.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster slot-map parsing used on every topology refresh; behavior on malformed replies is safer but could briefly omit nodes until the next good reply.
> 
> **Overview**
> **`parseSlots`** no longer assumes every `CLUSTER SLOTS` array entry has a full master triple and per-node host/port pairs. It now **bounds-checks** each slot row, the master sub-array, and each replica row; **truncated or partial entries are skipped** instead of indexing into empty slices and panicking.
> 
> This targets failover windows where topology replies can be incomplete while shards move (#929). **`TestClusterSlotsParsing`** was added to lock in existing fixture behavior and to assert that malformed replies are ignored or yield an empty group map without crashing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cd2fde0e7882dc2718f40254b217e0ae24b7850. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -375,21 +375,33 @@ func parseEndpoint(fallback, endpoint string, port int64) string {
 func parseSlots(slots RedisMessage, defaultAddr string) map[string]group {
 	groups := make(map[string]group, len(slots.values()))
 	for _, v := range slots.values() {
-		master := parseEndpoint(defaultAddr, v.values()[2].values()[0].string(), v.values()[2].values()[1].intlen)
+		values := v.values()
+		if len(values) < 3 {
+			continue
+		}
+		masterValues := values[2].values()
+		if len(masterValues) < 2 {
+			continue
+		}
+		master := parseEndpoint(defaultAddr, masterValues[0].string(), masterValues[1].intlen)
 		if master == "" {
 			continue
 		}
 		g, ok := groups[master]
 		if !ok {
 			g.slots = make([][2]int64, 0)
-			g.nodes = make(nodes, 0, len(v.values())-2)
-			for i := 2; i < len(v.values()); i++ {
-				if dst := parseEndpoint(defaultAddr, v.values()[i].values()[0].string(), v.values()[i].values()[1].intlen); dst != "" {
+			g.nodes = make(nodes, 0, len(values)-2)
+			for i := 2; i < len(values); i++ {
+				nodeValues := values[i].values()
+				if len(nodeValues) < 2 {
+					continue
+				}
+				if dst := parseEndpoint(defaultAddr, nodeValues[0].string(), nodeValues[1].intlen); dst != "" {
 					g.nodes = append(g.nodes, NodeInfo{Addr: dst})
 				}
 			}
 		}
-		g.slots = append(g.slots, [2]int64{v.values()[0].intlen, v.values()[1].intlen})
+		g.slots = append(g.slots, [2]int64{values[0].intlen, values[1].intlen})
 		groups[master] = g
 	}
 	return groups
@@ -419,11 +431,21 @@ func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]g
 			if dictHealth := dict["health"]; dictHealth.string() != "online" {
 				continue
 			}
-			port := dict["port"].intlen
-			if tls && dict["tls-port"].intlen > 0 {
-				port = dict["tls-port"].intlen
+			dictEndpoint, ok := dict["endpoint"]
+			if !ok {
+				continue
 			}
-			dictEndpoint := dict["endpoint"]
+			portMsg, portOK := dict["port"]
+			port := portMsg.intlen
+			_, tlsPortOK := dict["tls-port"]
+			if !portOK && !tlsPortOK {
+				continue
+			}
+			if tls {
+				if tlsPort, tlsPortOK := dict["tls-port"]; tlsPortOK && tlsPort.intlen > 0 {
+					port = tlsPort.intlen
+				}
+			}
 			if dst := parseEndpoint(defaultAddr, dictEndpoint.string(), port); dst != "" {
 				if dictRole := dict["role"]; dictRole.string() == "master" {
 					m = len(g.nodes)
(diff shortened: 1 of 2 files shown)


rueian:
Hi @mvanhorn, thanks for the defensive fix. Did you also run into this issue? I haven't personally seen any malformed responses, so I'm wondering whether this might be caused by another bug.

cluster.go
```diff
@@ -419,19 +431,29 @@ func parseShards(shards RedisMessage, defaultAddr string, tls bool) map[string]g
...
+			portMsg, portOK := dict["port"]
+			port := portMsg.intlen
+			_, tlsPortOK := dict["tls-port"]
+			if !portOK && !tlsPortOK {
+				continue
+			}
+			if tls {
+				if tlsPort, tlsPortOK := dict["tls-port"]; tlsPortOK && tlsPort.intlen > 0 {
+					port = tlsPort.intlen
+				}
```
rueian: I think these changes are not necessary. The old code won't panic anyway.

mvanhorn: Good call - removed the parseShards guards (and their test) in `7cd2fde`. You're right that nothing there panics: the `dict[...]` accessors return zero values, and `m >= 0` already implies a node was appended.

I kept the parseSlots bounds checks, though - those are the actual #929 fix. The panic in that report is at `cluster.go:378` in `parseSlots` (`index out of range [2] with length 0` when a CLUSTER SLOTS entry arrives truncated during failover), not in parseShards. So the PR is now just the parseSlots guard.

rueian (approved):
Thanks!

mvanhorn:
Thanks for reviewing and merging, @rueian. Guarding parseSlots against malformed CLUSTER SLOTS replies during failover should stop those panics.

PR #1008, opened 2026-07-08
feat: refresh dynamic auth credentials

### LINKED ISSUE #1007: Support `AUTH/HELLO` refresh
Related to https://github.com/redis/rueidis/issues/895.

Both Microsoft Entra ID authentication for Azure Managed Redis [ref1](https://learn.microsoft.com/en-us/azure/redis/entra-for-authentication#prerequisites-and-limitations) and IAM authentication for AWS ElastiCache [ref2](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/auth-iam.html#auth-iam-limits) require refreshing the access token in a new `AUTH` or `HELLO` command before the previous one expires; otherwise, the connection will be dropped. Therefore, we need to provide a way to do that.

What the way should be will need some discussion, but proposals and PRs are welcome.

My current preference is simply adding a new `RefreshBefore` field to the existing `AuthCredentials` struct, and each pipe will schedule a timer to call `AuthCredentialsFn` again before that timestamp.

mturac:
Fixes #1007

## Summary

- Add `AuthCredentials.RefreshAfter` so dynamic credential providers can schedule per-pipe credential refreshes after a caller-selected duration.
- Refresh credentials with `AUTH` for RESP2 and RESP3 connections, including password-only RESP3 connections that authenticate as the default user.
- Pause lifetime and credential-refresh timers while a pooled pipe is checked out, restart them when the pipe is returned, and discard a pipe if a timer callback is already active.
- Close the pipe when fetching or applying refreshed credentials fails, honoring the configured connection timeout for `AUTH`.

## Testing

- `git diff --check upstream/main...HEAD`
- `go test -run '^TestNewPipe$' -count=1 -timeout=3m .`
- `go test -race -run '^(TestNewPipe|TestPoolWithConnLifetime)$' -count=1 -timeout=5m .`

---

> [!NOTE]
> **High Risk**
> Changes authentication timing and in-flight AUTH on pooled connections; mis-timed refresh or pool/timer races could drop connections or leave stale credentials until exit.
> 
> **Overview**
> Adds **`AuthCredentials.RefreshAfter`** so `AuthCredentialsFn` can schedule per-connection credential rotation. Each pipe runs an **`authTimer`** that re-invokes the provider and issues **`AUTH`** on the live connection (RESP2/RESP3, including password-only RESP3 as `default`).
> 
> **Timer lifecycle** is aligned with connection lifetime: **`StopTimer` / `ResetTimer`** now pause and resume both lifetime and auth refresh while a pipe is checked out from the pool; **`_exit`** no longer stops timers globally. The pool **drops** newly created connections if `StopTimer` fails (timer already fired), matching reuse behavior for stale wires.
> 
> Failures while fetching or applying refreshed credentials **close the pipe**, with refresh **`AUTH`** bounded by **`ConnWriteTimeout`**. Broad **`pipe_test`** coverage plus pool tests for expired new connections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e3a859d2b51107d4bd170d95a106b14031a79ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pipe.go
```diff
@@ -380,6 +387,11 @@ func _newPipe(ctx context.Context, connFn func(context.Context) (net.Conn, error
 		p.lftm = option.ConnLifetime
 		p.lftmTimer = time.AfterFunc(option.ConnLifetime, p.expired)
 	}
+	if !nobg && option.AuthCredentialsFn != nil && !authCredentials.RefreshBefore.IsZero() {
+		p.authFn = option.AuthCredentialsFn
+		p.authContext = authCredentialsContext
+		p.scheduleAuthRefresh(authCredentials.RefreshBefore)
+	}
```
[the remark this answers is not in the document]
mturac: Addressed in `2f048bc`: auth refresh scheduling is no longer inside the `!nobg` branch, so `newPipeNoBg` connections can still schedule credential refreshes. The regression is covered by `TestNewPipe/Auth Credentials Refresh NoBg`, which creates a no-background pipe and verifies the refreshed `AUTH` command is sent.

Validation:
```sh
go test -run '^TestNewPipe$' -count=1 -timeout=2m .
go test -run 'TestNewPipe/Auth_Credentials_Refresh' ./...
git diff --check origin/main...HEAD
```

pipe.go
```diff
@@ -745,6 +757,66 @@ func (p *pipe) backgroundPing() {
...
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		if p.timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, p.timeout)
+			defer cancel()
+		}
+		if err := p.Do(ctx, cmd).Error(); err != nil {
+			p._exit(err)
+			return
+		}
```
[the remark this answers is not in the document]
mturac: Addressed in `2f048bc`: `refreshAuth` now checks the blocking signal and pending synchronous waits before fetching refreshed credentials or queueing `AUTH`, so a refresh will not enqueue behind a busy pipe. The regression is covered by `TestNewPipe/Auth Credentials Refresh Closes Busy Pipe`, which verifies the refresh closes the busy pipe and does not call the credential function a second time while the blocking command is active.

Validation:
```sh
go test -run '^TestNewPipe$' -count=1 -timeout=2m .
go test -run 'TestNewPipe/Auth_Credentials_Refresh' ./...
git diff --check origin/main...HEAD
```

pipe.go
```diff
@@ -745,6 +757,66 @@ func (p *pipe) backgroundPing() {
...
 }
 
+func authRefreshCmd(auth AuthCredentials) (Completed, bool) {
+	if auth.Username == "" && auth.Password == "" {
+		return Completed{}, false
+	}
+	if auth.Password != "" && auth.Username == "" {
+		return cmds.NewCompleted([]string{"AUTH", auth.Password}), true
+	}
+	return cmds.NewCompleted([]string{"AUTH", auth.Username, auth.Password}), true
```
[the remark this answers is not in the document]
mturac: Addressed in `2f048bc`: password-only RESP3 connections now remember that connect-time auth used the default ACL user, and refresh auth emits `AUTH default <password>` for that case. The regression is covered by `TestNewPipe/Auth Credentials Refresh RESP3 Default User`.

Validation:
```sh
go test -run '^TestNewPipe$' -count=1 -timeout=2m .
go test -run 'TestNewPipe/Auth_Credentials_Refresh' ./...
git diff --check origin/main...HEAD
```

mturac:
Validation refreshed on the current head:

```bash
git diff --check origin/main...HEAD
go test ./... -run 'TestPipe/Auth Credentials Refresh' -count=1
go test -race ./... -run 'TestPipe/Auth Credentials Refresh' -count=1
go test . -run TestPipe -count=1 -timeout=120s
```

All passed locally. The current CircleCI build/setup checks are also green.

pipe.go
```diff
@@ -24,8 +24,9 @@ const LibName = "rueidis"
 const LibVer = "1.0.76"
 
 var (
-	noHello = regexp.MustCompile("unknown command .?(HELLO|hello).?")
-	infoAZ  = regexp.MustCompile(`availability_zone:([^\r\n]+)`)
+	noHello               = regexp.MustCompile("unknown command .?(HELLO|hello).?")
+	infoAZ                = regexp.MustCompile(`availability_zone:([^\r\n]+)`)
+	errAuthRefreshBlocked = errors.New("auth refresh blocked by pending command")
```
rueian: Why do we need this? Can we remove this?

pipe.go
```diff
@@ -95,6 +97,11 @@ type pipe struct {
 	r2ps            bool // identify this pipe is used for resp2 pubsub or not
 	noNoDelay       bool
 	optIn           bool
+
+	authMu      sync.Mutex
```
rueian:
```suggestion
```

I think we just need to create the authTimer once during initialization. Is authMu really needed?

pipe.go
```diff
@@ -95,6 +97,11 @@ type pipe struct {
...
+	authFn      func(AuthCredentialsContext) (AuthCredentials, error)
+	authContext AuthCredentialsContext
+	authDefault bool
```
rueian:
```suggestion
```
Are these really needed?

pipe.go
```diff
@@ -745,6 +760,73 @@ func (p *pipe) backgroundPing() {
...
+	delay := time.Until(refreshBefore)
+	if delay < 0 {
+		delay = 0
+	}
+	p.authMu.Lock()
+	defer p.authMu.Unlock()
+	if p.authTimer != nil {
+		p.authTimer.Stop()
+	}
+	p.authTimer = time.AfterFunc(delay, p.refreshAuth)
```
rueian: If the timer is already created, then use p.authTimer.Reset.

pipe.go
```diff
@@ -745,6 +760,73 @@ func (p *pipe) backgroundPing() {
...
+}
+
+func (p *pipe) refreshAuth() {
+	if p.Error() != nil || p.authFn == nil {
+		return
+	}
+	if atomic.LoadInt32(&p.blcksig) != 0 || (atomic.LoadInt32(&p.state) == 0 && p.loadWaits() != 0) {
+		p._exit(errAuthRefreshBlocked)
+		return
+	}
```
rueian, line 772: I think we need to run the entire block below in another goroutine.

rueidis.go
```diff
@@ -483,6 +483,8 @@ type AuthCredentialsContext struct {
 type AuthCredentials struct {
 	Username string
 	Password string
+	// RefreshBefore schedules AuthCredentialsFn to be called again before this time.
+	RefreshBefore time.Time
```
rueian: I reconsidered this field a bit. This will be simpler:

```suggestion
	RefreshAfter time.Duration
```

mturac:
Updated in 10ad622:

- replaced `RefreshBefore time.Time` with `RefreshAfter time.Duration`;
- removed the auth timer mutex and reuse the timer with `Reset`;
- moved the refresh work out of the timer callback;
- kept the busy-pipe, no-background, RESP2, and RESP3 default-user refresh coverage aligned with the new API.

Validation:

- `go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=3m .`
- `go test -run '^TestNewPipe$' -count=1 -timeout=2m .`
- `go test ./... -run 'TestPipe|Auth Credentials Refresh|TestNewPipe' -count=1 -timeout=180s`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh_(Closes_Busy_Pipe|Timeout|After_StopTimer)$' -count=20 -timeout=5m .`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .`
- `git diff --check origin/main...HEAD`

mturac:
Updated in d9a5659:

- removed the custom auth-refresh blocked error and reuse the existing closing error when refresh cannot safely run while a command is pending;
- removed the stored default-user flag and derive the refresh AUTH shape from the pipe protocol state instead.

Validation:

- `go test -run '^TestNewPipe$/Auth_Credentials_Refresh_(RESP3|RESP3_Default_User|RESP2|Closes_Busy_Pipe)$' -count=1 -timeout=2m .`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .`
- `go test -run '^TestNewPipe$' -count=1 -timeout=2m .`
- `git diff --check origin/main...HEAD`

mturac:
Update: refreshed against the current head after the follow-up commits.

Current commit: `29ed971d2d6e0accd4306d411dd716325754e228`

Current CI status:

```text
ci/circleci: setup  success
ci/circleci: build  success
all visible checks passing
```

Local validation on this head:

```bash
git diff --check upstream/main...HEAD
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=5m .
```

pipe.go
```diff
@@ -745,6 +756,67 @@ func (p *pipe) backgroundPing() {
...
+		if p.timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, p.timeout)
+			defer cancel()
+		}
+		if err := p.Do(ctx, cmd).Error(); err != nil {
+			p._exit(err)
+			return
+		}
+	}
+	p.scheduleAuthRefresh(auth.RefreshAfter)
```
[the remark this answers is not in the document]
mturac: Addressed in `bdc0e65`: auth refresh now tracks open transactions and reschedules instead of fetching credentials or sending `AUTH` while a `MULTI` transaction is active. `EXEC` and `DISCARD` clear that state so refresh can resume after the transaction closes.

Added regression coverage in `TestNewPipe/Auth Credentials Refresh Waits For Open Transaction` to verify the refresh does not call the credential function while the transaction is open, then resumes and sends the refreshed `AUTH` after `EXEC`.

Validation:

```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
```

pipe.go
```diff
@@ -1093,6 +1174,21 @@ abort:
...
+	switch strings.ToUpper(commands[0]) {
+	case "MULTI":
+		if resp.Error() == nil {
+			atomic.StoreInt32(&p.txState, 1)
+		}
+	case "EXEC", "DISCARD":
+		atomic.StoreInt32(&p.txState, 0)
+	}
+}
+
```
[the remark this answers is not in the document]
mturac: Addressed in `b997245`.

- `DoMulti` now applies the same transaction tracking as `Do`, so `MULTI` opened through `DoMulti` defers auth refresh and `EXEC`/`DISCARD` sent through `DoMulti` clears the transaction state.
- Added `TestNewPipe/Auth Credentials Refresh Tracks DoMulti Transactions` to cover the `DoMulti(MULTI)` -> refresh wait -> `DoMulti(EXEC)` path.

Validation:

```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh_(Tracks_DoMulti_Transactions|Waits_For_Open_Transaction)' -count=1 -timeout=2m .
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
git diff --check origin/main...HEAD
```

pipe.go
```diff
@@ -1084,6 +1164,7 @@ queue:
 		}
 	}
 	p.decrWaitsAndIncrRecvs()
+	p.trackTransaction(cmd, resp)
```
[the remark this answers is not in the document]
mturac: Addressed in `3a4ad5d`.

- The aborted `Do` cleanup path now applies transaction tracking after the queued Redis reply arrives.
- The aborted `DoMulti` cleanup path now applies transaction tracking before returning the queued result slice to the pool.
- Added regression coverage for canceled `Do(MULTI)` and `DoMulti(MULTI)` calls: auth refresh stays deferred while the transaction is open, then resumes after `EXEC`.

Validation:

```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh_(Tracks_Aborted_Do_Transaction|Tracks_Aborted_DoMulti_Transaction)' -count=1 -timeout=2m .
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
go test -run '^TestNewPipe$' -count=1 -timeout=2m .
git diff --check origin/main...HEAD
```

pipe.go
```diff
@@ -745,6 +758,72 @@ func (p *pipe) backgroundPing() {
...
+		if p.timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, p.timeout)
+			defer cancel()
+		}
+		if err := p.Do(ctx, cmd).Error(); err != nil {
+			p._exit(err)
+			return
+		}
+	}
+	p.scheduleAuthRefresh(auth.RefreshAfter)
```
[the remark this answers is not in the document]
mturac: Addressed in `1b24f28d4700`.

- Mark `MULTI` as transaction-pending as soon as it enters the pipe queue, so auth refresh is deferred before the server reply is delivered.
- A failed `MULTI` reply clears only the pending marker; a successful `MULTI` keeps refresh deferred until `EXEC` or `DISCARD`.
- Added regression coverage for in-flight `MULTI` through both `Do` and `DoMulti`.

Validation:

```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh_Waits_For_In_Flight' -count=1 -timeout=2m .
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
go test -run '^TestNewPipe$' -count=1 -timeout=5m .
git diff --check origin/main...HEAD
```

pipe.go
```diff
@@ -745,6 +764,72 @@ func (p *pipe) backgroundPing() {
...
+	if p.Error() != nil || p.authFn == nil {
+		return
+	}
+	if atomic.LoadInt32(&p.txState) != 0 {
+		p.scheduleAuthRefresh(time.Duration(atomic.LoadInt64(&p.authRefreshAfter)))
+		return
+	}
+	if atomic.LoadInt32(&p.blcksig) != 0 || (atomic.LoadInt32(&p.state) == 0 && p.loadWaits() != 0) {
+		p._exit(ErrClosing)
+		return
```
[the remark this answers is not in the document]
mturac: Addressed in `9914a5615862556da8b8f0f06184d02d36c6ff2c`.

Non-blocking synchronous in-flight commands now defer the refresh instead of closing the pipe. Blocking commands still close the pipe when refresh cannot run safely; the blocking regression test now uses the blocking command marker explicitly.

Validation:
```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh_(Waits_For_Busy_Sync_Pipe|Rechecks_Transaction_After_Fetching_Credentials)$' -count=1 -timeout=2m .
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
go test -run '^TestNewPipe$' -count=1 -timeout=5m .
git diff --check
```

pipe.go
```diff
@@ -745,6 +764,72 @@ func (p *pipe) backgroundPing() {
...
+	}
+	cmd, ok := authRefreshCmd(auth, !p.r2ps && p.r2p == nil)
+	if ok {
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		if p.timeout > 0 {
+			ctx, cancel = context.WithTimeout(ctx, p.timeout)
+			defer cancel()
+		}
+		if err := p.Do(ctx, cmd).Error(); err != nil {
```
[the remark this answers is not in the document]
mturac: Addressed in `9914a5615862556da8b8f0f06184d02d36c6ff2c`.

Refresh now rechecks blocking, transaction, and busy sync state after fetching credentials and before sending `AUTH`. If another command opens a transaction or a sync command becomes in-flight during credential fetch, the refresh is rescheduled and `AUTH` is deferred until it is safe.

Validation:
```sh
go test -run '^TestNewPipe$/Auth_Credentials_Refresh_(Waits_For_Busy_Sync_Pipe|Rechecks_Transaction_After_Fetching_Credentials)$' -count=1 -timeout=2m .
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=3 -timeout=5m .
go test -run '^TestNewPipe$' -count=1 -timeout=5m .
git diff --check
```

pipe.go
```diff
@@ -81,6 +81,8 @@ type pipe struct {
 	psubs           *subs // pubsub pmessage subscriptions
 	r2p             *r2p
 	pingTimer       *time.Timer // timer for background ping
+	authTimerMu     sync.Mutex
```
rueian:
```suggestion
```
I think we don't need the mutex anymore.

pipe.go
```diff
@@ -90,15 +92,26 @@ type pipe struct {
...
+	txState         int32
 	state           int32
 	bgState         int32
 	r2ps            bool // identify this pipe is used for resp2 pubsub or not
 	noNoDelay       bool
 	optIn           bool
+
+	authFn           func(AuthCredentialsContext) (AuthCredentials, error)
+	authContext      AuthCredentialsContext
+	authRefreshAfter int64
```
rueian:
```suggestion
```
I think we don't need these. `AuthCredentialsContext` can be created on the fly, and `authFn` can be captured in the callback of `authTimer`. `authRefreshAfter` should be taken from the return of `authFn`.

pipe.go
```diff
@@ -745,6 +765,90 @@ func (p *pipe) backgroundPing() {
...
+func (p *pipe) scheduleAuthRefresh(refreshAfter time.Duration) {
+	if refreshAfter <= 0 || p.Error() != nil {
+		return
+	}
+	atomic.StoreInt64(&p.authRefreshAfter, int64(refreshAfter))
+	p.authTimerMu.Lock()
+	defer p.authTimerMu.Unlock()
+	if p.authTimer == nil {
+		p.authTimer = time.AfterFunc(refreshAfter, func() {
+			go p.refreshAuth()
```
rueian: My bad, time.AfterFunc has already run the function in another goroutine. So no need to start another goroutine.

pipe.go
```diff
@@ -1084,15 +1190,50 @@ queue:
...
 	go func(ch chan RedisResult) {
-		<-ch
+		resp := <-ch
+		p.trackTransaction(cmd, resp)
 		p.decrWaitsAndIncrRecvs()
 	}(ch)
 	return newErrResult(ctx.Err())
 }
 
+func (p *pipe) trackTransaction(cmd Completed, resp RedisResult) {
```
rueian: What is this doing? why do we need this?

pipe.go
```diff
@@ -745,6 +765,90 @@ func (p *pipe) backgroundPing() {
...
+		if err := p.Do(ctx, cmd).Error(); err != nil {
+			p._exit(err)
+			return
+		}
+	}
+	p.scheduleAuthRefresh(auth.RefreshAfter)
+}
+
+func (p *pipe) deferAuthRefresh() bool {
+	if atomic.LoadInt32(&p.txState) != txStateNone || (atomic.LoadInt32(&p.state) == 0 && p.loadWaits() != 0) {
```
rueian: What is this doing? why do we need this?

mturac:
Updated in `3907028`.

- Added comments explaining why auth refresh tracks pending/open `MULTI` state.
- Added comments explaining why refresh is rescheduled while a transaction or synchronous command is using the connection.

Validation:

```sh
git diff --check upstream/main...HEAD
go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .
go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=5m .
```

mturac:
Updated in `89d1ec6`.

- Removed the extra auth timer mutex and stored auth callback/context/refresh interval fields.
- The timer callback now captures the auth callback/context and keeps the current retry interval local to the callback.
- Kept the busy-pipe / transaction deferral behavior intact, including the race case where credentials are fetched but `AUTH` cannot be safely sent yet.

Validation:
- `git diff --check upstream/main...HEAD`
- `go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=5m .`
- `PATH="$(go env GOPATH)/bin:$PATH" ./dockertest.sh -skip Integration`

pipe.go
mturac: Clarified in 71ad2ab. This tracks a successful MULTI so the refresh timer does not inject AUTH into an open Redis transaction; Redis would queue it until EXEC/DISCARD and change the transaction contents. The aborted-request path updates after the server reply because client cancellation does not cancel the server-side MULTI.

pipe.go
```diff
@@ -99,6 +101,16 @@ type pipe struct {
 
 type pipeFn func(ctx context.Context, connFn func(ctx context.Context) (net.Conn, error), option *ClientOption) (p *pipe, err error)
 
+const (
+	txStateNone int32 = iota
+	// txStateOpen means a MULTI has been accepted by the server and the pipe
+	// must not send AUTH until EXEC or DISCARD closes the transaction.
```
rueian: That’s not going to happen. We don’t need to track this. Remove all these tx things you added.

rueian: Oh wait, you are right. This is indeed an issue but do we have other way to handle that? The current solution looks heavy.

mturac: Updated in `8e3e051` to remove the transaction-tracking path.

The refresh deferral now only covers the no-background synchronous write race; AUTH refresh still goes through the normal command path, without `txState` / MULTI / EXEC tracking.

Validation:
- `git diff --check origin/main`
- `go test -run ^TestNewPipe$/Auth_Credentials_Refresh -count=1 -timeout=2m .`
- `go test -race -run ^TestNewPipe$/Auth_Credentials_Refresh -count=1 -timeout=5m .`
- `PATH="$(go env GOPATH)/bin:$PATH" ./dockertest.sh -skip Integration`

pipe.go
```diff
@@ -380,6 +383,15 @@ func _newPipe(ctx context.Context, connFn func(context.Context) (net.Conn, error
...
 	}
+	if option.AuthCredentialsFn != nil && authCredentials.RefreshAfter > 0 {
+		authFn := option.AuthCredentialsFn
+		var refreshAfter atomic.Int64
+		refreshAfter.Store(int64(authCredentials.RefreshAfter))
+		p.authTimer = time.AfterFunc(authCredentials.RefreshAfter, func() {
+			nextRefreshAfter := p.refreshAuth(authFn, authCredentialsContext, time.Duration(refreshAfter.Load()))
+			refreshAfter.Store(int64(nextRefreshAfter))
+		})
+	}
```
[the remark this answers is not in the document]
mturac: Updated in `e752ebe`.

`_exit` now stops the auth refresh timer for external pipe exits, while auth-refresh initiated exits avoid stopping their own timer callback to keep the race detector clean. I added a regression test that exits the pipe before `RefreshAfter` and verifies `AuthCredentialsFn` is not called again.

Validation:
- `git diff --check origin/main`
- `go test -run ^TestNewPipe$/Auth_Credentials_Refresh -count=1 -timeout=2m .`
- `go test -race -run ^TestNewPipe$/Auth_Credentials_Refresh -count=1 -timeout=5m .`
- `PATH="$(go env GOPATH)/bin:$PATH" ./dockertest.sh -skip Integration`

pipe.go
```diff
@@ -745,6 +768,81 @@ func (p *pipe) backgroundPing() {
 	})
 }
 
+func (p *pipe) scheduleAuthRefresh(refreshAfter time.Duration) {
+	if refreshAfter <= 0 || p.Error() != nil || p.authTimer == nil {
+		return
+	}
+	p.authTimer.Reset(refreshAfter)
+}
```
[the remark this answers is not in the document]
mturac: Updated in `42c445d`.

`authTimer.Reset` and `authTimer.Stop` now share the same timer mutex, and the initial timer assignment is covered by that lock as well. This keeps auth refresh rescheduling and pipe exit from touching the timer concurrently.

Validation:
- `git diff --check origin/main`
- `go test -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=2m .`
- `go test -race -run '^TestNewPipe$/Auth_Credentials_Refresh' -count=1 -timeout=5m .`
- `PATH="$(go env GOPATH)/bin:$PATH" ./dockertest.sh -skip Integration`

pipe.go
```diff
@@ -380,6 +384,17 @@ func _newPipe(ctx context.Context, connFn func(context.Context) (net.Conn, error
 		p.lftm = option.ConnLifetime
 		p.lftmTimer = time.AfterFunc(option.ConnLifetime, p.expired)
 	}
+	if option.AuthCredentialsFn != nil && authCredentials.RefreshAfter > 0 {
+		authFn := option.AuthCredentialsFn
+		var refreshAfter atomic.Int64
```
rueian:
```suggestion
```
Why do we need this? Can we remove it?

pipe.go
```diff
@@ -393,9 +408,20 @@ func (p *pipe) background() {
...
 
 func (p *pipe) _exit(err error) {
+	p._exitWithAuthRefresh(err, true)
+}
+
+func (p *pipe) exitAuthRefresh(err error) {
+	p._exitWithAuthRefresh(err, false)
+}
+
+func (p *pipe) _exitWithAuthRefresh(err error, stopAuthRefresh bool) {
```
rueian: Please don't create a new _exitWithAuthRefresh function, and the stopAuthRefresh bool is not necessary. We should always stop the refresh if the timer is not nil.

pipe.go
```diff
@@ -745,6 +771,89 @@ func (p *pipe) backgroundPing() {
...
+}
+
+func (p *pipe) refreshAuth(authFn func(AuthCredentialsContext) (AuthCredentials, error), authContext AuthCredentialsContext, refreshAfter time.Duration) time.Duration {
+	if p.Error() != nil {
+		return refreshAfter
+	}
+	if atomic.LoadInt32(&p.blcksig) != 0 {
+		p.exitAuthRefresh(ErrClosing)
+		return refreshAfter
+	}
```
rueian:
```suggestion
```
It is still ok to send the AUTH if there is a blocking command in flight.

pipe.go
```diff
@@ -745,6 +771,89 @@ func (p *pipe) backgroundPing() {
...
+		p.exitAuthRefresh(err)
+		return refreshAfter
+	}
+	if atomic.LoadInt32(&p.blcksig) != 0 {
+		p.exitAuthRefresh(ErrClosing)
+		return refreshAfter
+	}
+	if p.deferAuthRefresh(refreshAfter) {
+		return refreshAfter
+	}
```
rueian:
```suggestion
```

pipe.go
```diff
@@ -745,6 +771,89 @@ func (p *pipe) backgroundPing() {
...
+	if p.Error() != nil {
+		return refreshAfter
+	}
+	if atomic.LoadInt32(&p.blcksig) != 0 {
+		p.exitAuthRefresh(ErrClosing)
+		return refreshAfter
+	}
+	if p.deferAuthRefresh(refreshAfter) {
+		return refreshAfter
+	}
```
rueian:
```suggestion
```

I reconsidered the transaction problem. I realized that no one should use the MULTI command with `.Do`; instead, they should use `.DoMulti` to cover the whole transaction. Please just ignore the problem for now and remove your defer code. I will probably fix that by forbidding sending the MULTI command with `.Do`.

pipe.go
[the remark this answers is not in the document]
rueian: Don't stop the timer in the Close method. Stop it at https://github.com/redis/rueidis/blob/37b8a532d1e1b55f422ce823a523ced32d66235a/pipe.go#L434-L436

pipe.go
```diff
@@ -1864,10 +1974,10 @@ func (p *pipe) Close() {
 }
 
 func (p *pipe) StopTimer() bool {
-	if p.lftmTimer == nil {
-		return true
+	if p.lftmTimer != nil {
+		return p.lftmTimer.Stop()
 	}
-	return p.lftmTimer.Stop()
+	return true
```
rueian: Do not change this function.

pipe.go
```diff
@@ -745,6 +766,90 @@ func (p *pipe) backgroundPing() {
...
+	}
+	p.authPending.Store(true)
+	if p.syncMu.TryLock() {
+		if p.authPending.CompareAndSwap(true, false) {
+			return true
+		}
+		p.syncMu.Unlock()
+	}
+	return false
+}
```
rueian: We don't need these. For pooled pipes, including those created with newPipeNoBg, we just need to stop timers when the pipe is taken from the pool and reset timers when the pipe is returned to the pool (using the existing StopTimer and ResetTimer interfaces).

pipe.go
```diff
@@ -396,6 +411,9 @@ func (p *pipe) _exit(err error) {
 	p.error.CompareAndSwap(nil, &errs{error: err})
 	atomic.CompareAndSwapInt32(&p.state, 1, 2) // stop accepting new requests
 	_ = p.conn.Close()                         // force both read & write goroutine to exit
+	if p.authTimer != nil {
+		p.authTimer.Stop()
+	}
```
rueian: We should not stop timers in the _exit function because _exit can be called concurrently. The below p.StopTimer() is also a bug, which I will fix later, including the "NoBg Close skips timer stop" issue.

pipe.go
```diff
@@ -368,6 +372,16 @@ func _newPipe(ctx context.Context, connFn func(context.Context) (net.Conn, error
 			}
 		}
 	}
+	if option.AuthCredentialsFn != nil && authCredentials.RefreshAfter > 0 {
+		authFn := option.AuthCredentialsFn
+		p.authRefreshAfter = authCredentials.RefreshAfter
+		ready := make(chan struct{})
+		p.authTimer = time.AfterFunc(p.authRefreshAfter, func() {
+			<-ready
```
rueian: is this really needed?

pool_test.go
```diff
@@ -374,22 +382,6 @@ func TestPoolWithConnLifetime(t *testing.T) {
...
-		}
-
-		pool.cond.L.Lock()
-		if pool.size != 1 {
-			t.Errorf("size must be equal to 1, actual: %d", pool.size)
-		}
-		if len(pool.list) != 1 {
-			t.Errorf("list len must equal to 1, actual: %d", len(pool.list))
-		}
-		pool.cond.L.Unlock()
```
rueian, line 392: Why are these removed?

pool_test.go
```diff
@@ -344,18 +344,23 @@ func TestPoolWithConnLifetime(t *testing.T) {
...
 					return false
 				}, // connection lifetime timer is already fired
 			},
+			&mockWire{},
 		}
-		conn := make([]wire, 0, len(wires))
+		conn := make([]wire, 0, 2)
 		pool := setup(wires)
-		for range wires {
+		for range 2 {
```
rueian, line 413: Why are these changed?

pipe.go
```diff
@@ -1021,7 +1080,6 @@ func (p *pipe) Do(ctx context.Context, cmd Completed) (resp RedisResult) {
 	if err := ctx.Err(); err != nil {
 		return newErrResult(err)
 	}
-
```
rueian, line 1024: keep these empty lines

pipe.go
```diff
@@ -1864,17 +1922,28 @@ func (p *pipe) Close() {
...
+	if p.Error() != nil {
 		return true
 	}
-	return p.lftmTimer.Reset(p.lftm)
+	reset := true
+	if p.lftmTimer != nil {
+		reset = p.lftmTimer.Reset(p.lftm)
+	}
+	if p.authTimer != nil && p.authRefreshAfter > 0 {
+		reset = p.authTimer.Reset(p.authRefreshAfter) && reset
```
rueian: Hmm, my bad, let's go back to make RefreshAfter a time.Time so that we can be more accurate when resetting the timer.

pipe.go
```diff
@@ -745,6 +761,49 @@ func (p *pipe) backgroundPing() {
...
+	if p.Error() != nil {
+		return
+	}
+	auth, err := authFn(authContext)
+	if err != nil {
+		p._exit(err)
+		return
+	}
+	if auth.Username != "" || auth.Password != "" {
+		args := []string{"AUTH"}
```
rueian: Please preallocate the slice slots for username and password.

pipe.go
```diff
@@ -66,35 +66,37 @@ type wire interface {
...
+	lftm             time.Duration // lifetime
+	wrCounter        atomic.Uint64
+	version          int32
+	blcksig          int32
+	state            int32
+	bgState          int32
+	r2ps             bool // identify this pipe is used for resp2 pubsub or not
+	noNoDelay        bool
+	optIn            bool
+	authRefreshAfter time.Duration
```
rueian: My bad, please make this a time.Time. Also, fields should be aligned by size. It won't be the last field.

mturac:
Updated in `48eed1c`.

- Changed `AuthCredentials.RefreshAfter` back to an absolute `time.Time` and reset the auth refresh timer with `time.Until(...)`.
- Removed the extra `ready` channel around the auth refresh timer.
- Preallocated the `AUTH` argument slice and restored the requested blank line in `Do`.
- Kept the current pool guard/test shape: reverting that guard locally reproduced the `TestPoolWithAcquireCtx/Acquire connections, all exceed context deadline` pool-size leak (`actual: 1`), so I left the guard in place and covered it with the targeted pool tests.

Validation:
- `go test . -run 'TestPipe/Auth Credentials Refresh|TestPoolWithConnLifetime|TestPoolWithAcquireCtx'`

pool_test.go
mturac: This block was removed because the regression moved to the make-path case: the stale wire can be created with an already-fired lifetime timer before it ever enters the idle list. The updated test covers that path by closing the stale newly made wire and retrying with a healthy replacement, while the old `Store(Acquire())` block only covered dropping an expired idle wire after reuse.

pool_test.go
mturac: These changes make the test exercise the newly made stale-wire path. The second mock wire has `StopTimer` return false, so the pool should close it and decrement `size` before retrying; the third mock wire is the healthy replacement. Reverting this guard locally reproduced the `TestPoolWithAcquireCtx/Acquire connections, all exceed context deadline` `pool.size` leak (`actual: 1`).

pool_test.go
mturac: Updated in `0e3a859`: I restored this coverage instead of dropping it. The test now keeps the idle-wire path where `Store(Acquire())` drops an expired pooled connection, and adds a separate subtest for the newly made expired-wire retry path.

Validation:
- `git diff --check upstream/main...HEAD`
- `go test . -run 'TestPipe/Auth Credentials Refresh|TestPoolWithConnLifetime|TestPoolWithAcquireCtx' -count=1`

pool_test.go
mturac: Updated in `0e3a859`: I split the paths so the original reuse test no longer has to change shape for the newly made expired-wire regression. The third mock wire is now only used by the new `Drop newly made expired connections` subtest.

Validation:
- `git diff --check upstream/main...HEAD`
- `go test . -run 'TestPipe/Auth Credentials Refresh|TestPoolWithConnLifetime|TestPoolWithAcquireCtx' -count=1`

rueian (approved):

PR #1009, opened 2026-07-15
rueidiscompat: fix HSetEXWithArgs panic when Condition is unset

jtlang:
## Summary
`HSetEXWithArgs` only built the underlying `HSETEX` command inside the `Condition == HSetEXFNX` / `Condition == HSetEXFXX` branches. Since `Condition` is optional (per `HSetEXOptions`), leaving it unset to apply only `ExpirationType` (e.g. `EX`/`PX`/`KEEPTTL`) left the command builder at its zero value, causing a nil pointer dereference in `HsetexFieldValue.FieldValue`.

## Repro
```go
adapter.HSetEXWithArgs(ctx, "key", &rueidiscompat.HSetEXOptions{
    ExpirationType: rueidiscompat.HSetEXExpirationEX,
    ExpirationVal:  100,
}, "field1", "value1")
```
panics with `invalid memory address or nil pointer dereference` in `gen_hash.go:HsetexFieldValue.FieldValue`.

## Fix
Add the missing no-condition path in the `Condition` switch, building directly off `Hsetex().Key(key)` (no `.Fnx()`/`.Fxx()`). Also added a `default` case to each `ExpirationType` switch so an unset expiration falls through to `.Fields()` instead of leaving the command unset.

## Test plan
- Added a regression test (`should HSetEXWithArgs apply ExpirationType without a Condition`) in the existing `Redis 8.6+ commands` suite, covering set + get + TTL round-trip.
- Verified the new test panics on `main` (before the fix) and passes after.
- `go build`/`go vet` clean; existing hash-command specs still pass.

---

> [!NOTE]
> **Low Risk**
> Localized fix to Redis hash command adapter command-building; behavior change only affects previously broken call paths.
> 
> **Overview**
> Fixes a **nil pointer panic** in `HSetEXWithArgs` when `HSetEXOptions.Condition` is left unset but `ExpirationType` (e.g. `EX`) is set—previously only `FNX`/`FXX` branches built the `HSETEX` command.
> 
> The adapter now handles the **no-condition** path by building `Hsetex().Key(key)` with the appropriate expiration modifiers, and adds **`default`** fallbacks on each `ExpirationType` switch (including `FNX`/`FXX`) so the field builder is always initialized before `FieldValue` runs.
> 
> A regression test covers set/get/TTL when only expiration options are provided.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff969bc1c03e1743dfc8cee6e87e82ed29ea3631. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/rueidiscompat/adapter.go
+++ b/rueidiscompat/adapter.go
@@ -1672,6 +1672,8 @@ func (c *Compat) HSetEXWithArgs(ctx context.Context, key string, options *HSetEX
 			partial = c.client.B().Hsetex().Key(key).Fnx().Pxat(options.ExpirationVal).Fields().Numfields(int64(len(fieldsAndValues) / 2)).FieldValue()
 		case HSetEXExpirationKEEPTTL:
 			partial = c.client.B().Hsetex().Key(key).Fnx().Keepttl().Fields().Numfields(int64(len(fieldsAndValues) / 2)).FieldValue()
+		default:
+			partial = c.client.B().Hsetex().Key(key).Fnx().Fields().Numfields(int64(len(fieldsAndValues) / 2)).FieldValue()
 		}
 	case HSetEXFXX:
 		switch options.ExpirationType {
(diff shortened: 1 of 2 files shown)


rueian:
Hi @jtlang, is the new behavior aligned with go-redis?

jtlang:
@rueian Yes, this will bring it in line with go-redis behavior

rueian (approved):

rueian (approved):

rueian (approved):

Issue #1019, opened 2026-08-14
Current default LRU cache may remain above CacheSizeEachConn after an update


siriookie:
The built-in LRU cache may remain above `CacheSizeEachConn` when a newly
inserted value requires multiple completed entries to be evicted.

## Problem

The eviction loop intends to continue until:

```go
c.size <= c.max
```

Current implementation:

```go
ele = c.list.Front()
for c.size > c.max && ele != nil {
    if e := ele.Value.(*cacheEntry); e.val.typ != 0 {
        kc := e.kc
        if delete(kc.cache, e.cmd); len(kc.cache) == 0 {
            delete(c.store, kc.key)
        }
        c.list.Remove(ele)
        c.size -= e.size
    }
    ele = ele.Next()
}
```

However, `container/list.Remove(ele)` clears the removed element's
`next`, `prev`, and `list` fields.
```go
//D:\go\src\container\list\list.go
// remove removes e from its list, decrements l.len
func (l *List) remove(e *Element) {
	e.prev.next = e.next
	e.next.prev = e.prev
	e.next = nil // avoid memory leaks
	e.prev = nil // avoid memory leaks
	e.list = nil
	l.len--
}
```
Therefore, calling:

```go
ele.Next()
```

after removing `ele` returns `nil`, causing the loop to stop after evicting
the first completed entry.

## Reproduction

The issue can be reproduced with the following size distribution:

```text
old entry size: 385
cache max:      770
existing size:  385 + 385 = 770
new entry size: 386
```

After inserting the new entry:

```text
total size:              1156
after one old eviction:   771
configured max:           770
```

A second old entry must be evicted, but the loop stops after the first
removal.

A regression test fails on the current implementation with:

```text
cache size exceeded after eviction: size=771 max=770
```

## Expected Behavior

`Update` should continue scanning and evicting completed entries until:

```go
c.size <= c.max
```

or no more entries are available.

## Actual Behavior

At most one completed entry is removed by each `Update` call.

The cache's accounted size can therefore remain above `CacheSizeEachConn`,
especially when entry sizes differ significantly or a large value is
inserted.

## Proposed Fix

Preserve the next element before removing the current element:

```go
ele := c.list.Front()
for c.size > c.max && ele != nil {
    next := ele.Next()

    if e := ele.Value.(*cacheEntry); e.val.typ != 0 {
        kc := e.kc
        if delete(kc.cache, e.cmd); len(kc.cache) == 0 {
            delete(c.store, kc.key)
[...]

PR #1020, opened 2026-08-14
fix: continue LRU eviction after removals

siriookie:
#1019 

---

> [!NOTE]
> **Low Risk**
> Small, localized fix to client-side LRU bookkeeping in `Update`, covered by a new regression test; no auth or persistence impact.
> 
> **Overview**
> Fixes **LRU eviction in `Update`** when a newly stored value pushes `c.size` over `c.max`. The eviction walk from the list front now saves **`next := ele.Next()` before `Remove`**, then continues with `ele = next`, instead of calling `Next()` on a removed node (which could stop eviction early and leave stale entries).
> 
> Adds **`Cache Evict Multiple Entries`** in `lru_test.go`: a cache sized for two small entries must evict both when a slightly larger third entry is updated in, asserting final size, list length, and that keys `0`/`1` are gone while the new key remains.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48a92742811c8daff0e26c5ae45e189e36670168. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rueian (approved):
Thanks @siriookie

PR #1021, opened 2026-08-15
fix(rueidiscompat): make CacheCompat a mockable interface

### LINKED ISSUE #972: Facing issue in mocking while migrating from go-redis, mockery mocks
Hi @rueian 

Currently facing mock issue in my repo for mocking the Cache() method. The Cache() method returns a concrete type of CachaCompat.

```
cacheCompat := mocks.NewCacheCompat(t)

stringCmd := &rueidiscompat.StringCmd{}
stringCmd.SetErr(rueidiscompat.Nil)
RedisClient.On("Cache", 100*time.Millisecond).Return(cacheCompat)
cacheCompat.On("Get", ctx, "key").Return(stringCmd)
```

Also creating an interface at my end does not work as mockery fails it with incorrect type.

`panic: interface conversion: interface {} is *mocks.CacheCompat, not rueidiscompat.CacheCompat [recovered, repanicked]`

Only solution which I see is to return the CacheCompat as an interface, what do you suggest?

SomSamantray:
## Summary

Client-side caching via `rueidiscompat.Cache()` could not be mocked. `Cache(ttl)` returned a concrete `CacheCompat` struct, so mockery/gomock could not generate a test double and runtime panicked with `interface conversion: interface {} is *mocks.CacheCompat, not rueidiscompat.CacheCompat`.

`CacheCompat` is now an exported interface (82 methods, matching the previous exported command surface) backed by an unexported `cacheCompat` struct, and `Cache()` returns the interface. Behavior and method set are unchanged; `Pipeline.Cache()` keeps its existing "not implemented" panic. The interface exposes only exported methods, so external test doubles (mockery, gomock, hand-written) can implement it.

## Breaking change note (for maintainers)

This is a source-level API change: code that names the concrete `CacheCompat` type (composite literals such as `CacheCompat{}` or `&CacheCompat{}`, struct embedding, reflection on the struct kind) no longer compiles. Chained-method callers (`compat.Cache(ttl).Get(ctx, key)`) are unaffected.

Please treat this as a minor (not patch) version change for the `rueidiscompat` module and apply the `breakingchange` label so `release-drafter` surfaces it under Breaking Changes.

## Tests

Adds `rueidiscompat/cache_compat_test.go` (external test package) with a literal 82-method `fakeCache` that satisfies `CacheCompat` exactly as a mockery/gomock double would. The test also:

- Pins the interface surface with a reflection check (NumMethod == 82 for interface and fake), so a method removed from the interface fails the test instead of passing silently.
- Non-vacuously dispatches `Get` through the interface (the fake returns a distinguishable payload).
- Fails to compile if an unexported method is ever added to the interface, protecting the exported-only constraint.

The interface, the `cacheCompat` struct, and the compile assertion were moved out of the 6830-line `adapter.go` into `rueidiscompat/cache_compat.go`, mirroring the package's `PubSub`/`Tx` file-per-surface convention.

## Validation

- `go build ./...` and `go vet ./...` pass at repo root, in `rueidiscompat`, and in `rueidiscompatmock` (unmodified; its embedded `Cmdable` inherits the new return type).
- Offline `rueidiscompat` test subset passes, including `TestCacheCompatMockability`.
- The live-Redis Ginkgo `Adapter` suite, which exercises `Cache()` behavior, is Docker-gated and runs in CI via `dockertest.sh`; it was not runnable locally.

Fixes #972

---

> [!NOTE]
> **Medium Risk**
> Source-level breaking change for consumers that depended on the concrete `CacheCompat` struct; runtime caching behavior should be unchanged.
> 
> **Overview**
> **`CacheCompat` is now an exported interface** (82 read-only/cached commands) returned from `Cache(ttl)`, with the real implementation moved to an unexported `cacheCompat` struct in new `cache_compat.go`. Cached command methods on that type were renamed from `CacheCompat` to `cacheCompat` receivers only; behavior is unchanged.
> 
> This unblocks mockery/gomock and hand-written fakes for client-side caching (fixes #972). Callers using `compat.Cache(ttl).Get(...)` keep working; **source breaks** if code referenced the old concrete `CacheCompat` struct (literals, embedding, struct-kind reflection).
> 
> Adds `cache_compat_test.go` with an external-package `fakeCache`, interface dispatch check, and reflection pinning the method count so the surface stays mockable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f1ff94e4cdc4257486c5ae5e8f75a24ee4a9a2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rueian:
Hi @SomSamantray, thanks for the PR, but please make changes to rueidiscompatmock instead.

SomSamantray:
Superseded by #1023. Per maintainer feedback, the fix was reworked into `rueidiscompatmock` (wiring `DoCache` into the expectation queue) instead of changing the `rueidiscompat` core API.

Issue #1022, opened 2026-08-16
rueidiscompatmock Cache() mock is always-miss: document divergence


SomSamantray:
The new rueidiscompatmock DoCache stub (routing cached commands through the expectation queue, TTL ignored) models a client with caching disabled. The real client serves repeated identical commands from its in-process LRU within TTL with zero round trips and surfaces that state via Cmd.IsCacheHit(). Two constructible failure directions for test authors: (1) two expectations for the same key with different values + two in-TTL reads pass on the mock but diverge from real client behavior; (2) a one-expectation test with two in-TTL reads fails on the mock with 'no expectation for command' while the real client serves both reads from cache. IsCacheHit() can never be true on the mock, so application branches on cache-hit state are never exercised.

Suggested fix (arm a, compatible with the settled TTL-ignoring design): document the always-miss divergence explicitly in rueidiscompatmock/README.md: repeated in-TTL reads are never served from cache, IsCacheHit() is always false, and freshness-across-two-reads tests do not reflect real client behavior.

Evidence:
- rueidiscompatmock/adapter.go:567 -- DoAndReturn(func(_ context.Context, cmd rueidis.Cacheable, _ time.Duration) rueidis.RedisResult {
- pipe.go:1563 -- if v, entry := p.cache.Flight(ck, cc, ttl, now); v.typ != 0 { return NewResult(v, nil) }
- lru.go:89 -- if e != nil && (v.typ == 0 || v.relativePTTL(now) > 0) {
- rueidiscompat/command.go:106 -- cmd.setIsCacheHit(res.IsCacheHit())

Source: branch fix/rueidiscompatmock-cache-mock @ 0045d5f (PR forthcoming)

Metadata:
Severity: P2
Confidence: 75
Reviewer(s): adversarial
Finding ID: rueidiscompatmock/adapter.go:564-570:repeated-inttl-cached-reads-diverge-from-real-clientside-caching-the-docache-stub-is-alwaysmiss

SomSamantray:
Carried by PR https://github.com/redis/rueidis/pull/1023 (the fix that introduces the DoCache stub).

PR #1023, opened 2026-08-16
fix(rueidiscompatmock): make Cache() mockable via the expectation queue

### LINKED ISSUE #972: Facing issue in mocking while migrating from go-redis, mockery mocks

SomSamantray:
`rueidiscompatmock` now supports mocking client-side caching. `Cache(ttl)` commands dispatch through the existing `Expect*` expectation queue instead of panicking with `unexpected call to DoCache` (issue #972). The new `DoCache` stub reuses the same ordered queue as `Do`/`DoMulti`, so no `Expect*` API changed and the TTL argument is deliberately not matched (documented in the README).

Design notes:
- The fix lives in `rueidiscompatmock` only; `rueidiscompat` and the root `mock/` package are untouched. This follows the maintainer's direction on #1021 instead of the rejected interface-ization of `CacheCompat`.
- `DoMultiCache` is not stubbed: the `rueidiscompat` cache surface never calls it (verified against the receiver block in `rueidiscompat/adapter.go`).

Validated with the module's full test suite including `-race`. Eight new `TestCacheCompat*` tests cover dispatch through both adapter entry points, multi-command queues, `Cacheable`-wrapped commands, nil/error results, TTL irrelevance, and unmatched-command reporting.

Fixes #972.

Related: #1021

---

> [!NOTE]
> **Low Risk**
> Test-helper only; no production Redis client or auth/data path changes. Matching origin (cache vs plain) is a behavior change for existing mock tests that mix Cache() with unmarked Expect* calls.
> 
> **Overview**
> Makes `rueidiscompat.Cache(ttl)` commands mockable in `rueidiscompatmock` instead of panicking on `DoCache`. Cached calls reuse the existing `Expect*` queue; TTL is not matched.
> 
> Matching is origin-aware: `ExpectGet("k").ViaCache()` only matches `rdb.Cache(ttl).Get(...)`, and unmarked expectations only match plain calls. Unordered matching still prefers the origin-specific entry rather than the first same-command expectation.
> 
> `ViaCache()` is added on all expected-result types. Unmatched cache calls report `via cache` in the error. `DoMultiCache` is not stubbed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit def365b1c7b630456aa2f89d32ce7547ee85e325. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/residual-review-findings/0045d5f.md
```diff
@@ -0,0 +1,12 @@
...
+- **P2** — `rueidiscompatmock/adapter.go:567` — Repeated in-TTL cached reads diverge from real client-side caching: the DoCache stub is always-miss — filed: https://github.com/redis/rueidis/issues/1022
+
+---
+
+Source run context:
+- Review run: ce-code-review run-id `20260816-094523-053cdd06` (artifact dir `/tmp/compound-engineering-501/ce-code-review/20260816-094523-053cdd06`)
+- Branch: `fix/rueidiscompatmock-cache-mock` @ `0045d5f`
+- Plan: `docs/plans/2026-08-16-002-rueidiscompatmock-cache-mock-plan.md`
+- Finding type: `settled_conflict` — stamped against KTD2 (Ignore the TTL argument in matching; session-settled). Report-only: the alternative arm (in-memory TTL cache in the stub) is the rejected alternative and is not applied; the compatible arm (document the always-miss divergence) is the ticket's suggested fix.
+- Owner: human (advisory-grade; not in the downstream-resolver actionable queue)
```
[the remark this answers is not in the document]
SomSamantray: > This file looks like internal review-tooling output rather than project docs.

Agreed — removed in f26ebd0. The record's content remains durable in the linked issue (redis/rueidis#1022); it does not need to ship in the repository.

rueidiscompatmock/adapter_test.go
```diff
@@ -3066,6 +3066,138 @@ func wantAnyErr(err error) error {
 	return nil
 }
 
+func TestCacheCompatGet(t *testing.T) {
+	ctx := context.Background()
+	rdb, cm := NewClientMock()
+	defer rdb.Close()
+
+	cm.ExpectGet("key").SetVal("v")
```
rueian: Is sharing the same ExpectGet really a good idea? What if we want to differentiate between a normal Get and a cached Get?

SomSamantray: Good catch — added in 59eaff3. `Expected*` expectations now support a chainable `.ViaCache()` modifier: `cm.ExpectGet("key").ViaCache().SetVal("v")` only matches a call routed through `DoCache` (i.e. `rdb.Cache(ttl).Get(...)`), and a plain `Get()` for the same key will *not* satisfy it (mock reports "no expectation for command", same as any other mismatch). Expectations that don't opt in keep matching either origin, so nothing in the existing suite changes behavior.

Added test coverage for all three cases (cache-required match, cache-required rejects a plain call, unmarked still matches both origins, including under `MatchExpectationsInOrder(false)`) and documented it in the README's client-side caching section.

rueidiscompatmock/adapter.go
```diff
@@ -604,25 +611,28 @@ func (m *clientMock) consumeWithRaw(cmd rueidis.Completed) resolvedExpectation {
...
 		}
 	}
 	return 0, false
 }
 
-func (e *expectation) matches(cmd rueidis.Completed) bool {
+func (e *expectation) matches(cmd rueidis.Completed, viaCache bool) bool {
+	if e.requireCache && !viaCache {
+		return false
+	}
```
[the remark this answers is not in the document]
SomSamantray: Confirmed and fixed in 814612f — good catch. `matchLocked`'s unordered scan now does a first pass looking only at `requireCache=true` entries when the call arrived via `DoCache`, falling back to the original single-pass scan otherwise. A `ViaCache()`-marked expectation is no longer stolen by an unmarked expectation queued ahead of it for the same command; plain calls are unaffected since cache-required entries never match them.

Added a regression test (`TestCacheCompatViaCacheUnorderedGenericQueuedFirst`) reproducing exactly this scenario, plus a fallback test confirming a cached call still matches a generic expectation when no cache-specific one exists. I verified the new regression test actually fails on the pre-fix code before restoring the fix, so it's a real guard against a regression here.

rueidiscompatmock/adapter_test.go
rueian: Hi @SomSamantray, ss it possible to enforce a `rdb.Cache(100 * time.Millisecond).Get(ctx, "key")` to be matched only by `.ViaCache()`? ie, a plain `cm.ExpectGet("key")` should not pass the `ExpectationsWereMet`.

SomSamantray: Fixed in 3852a76 — matching is now symmetric. `expectation.matches` rejects the call whenever `e.requireCache != viaCache`, so an unmarked `cm.ExpectGet("key")` no longer matches a call made through `rdb.Cache(ttl).Get(...)`; only a `.ViaCache()`-marked expectation does. `ExpectationsWereMet()` now fails when a cache-routed call has no `.ViaCache()` expectation queued, even if an unmarked one for the same command is present.

Updated the whole `TestCacheCompat*` suite for the new contract: every test that exercised a `Cache(ttl)` call now uses `.ViaCache()`, and the two tests that codified the old fallback (`TestCacheCompatUnmarkedMatchesBothOrigins`, `TestCacheCompatViaCacheUnorderedFallsBackToGeneric`) are replaced with `TestCacheCompatUnmarkedMatchesPlainOnly` and `TestCacheCompatViaCacheUnorderedRequiresViaCache`, which assert the opposite: an unmarked expectation now rejects a cache-routed call. I verified these new/replaced tests fail against the pre-fix code and pass after. Also updated the README's client-side caching section to document the symmetric contract.

rueidiscompatmock/viacache.go
```diff
@@ -0,0 +1,226 @@
+package rueidiscompatmock
+
+// ViaCache marks this expectation so it is only satisfied by a call routed
+// through DoCache (i.e. rueidiscompat.Cache(ttl).<Cmd>()). A plain, non-cached
+// call for the same command will not match it. Expectations that do not call
+// ViaCache are unaffected and continue to match calls from either origin.
```
rueian: This seems to be stale.

SomSamantray: Good catch, fixed in 5efc25d. This comment predated the symmetric-matching fix in 3852a76 and still described the old fallback ("expectations that do not call ViaCache are unaffected and continue to match calls from either origin"), which is no longer true — an unmarked expectation now only matches plain calls. Updated the doc comment to say so.

rueidiscompatmock/adapter.go
```diff
@@ -553,23 +554,29 @@ func (m *clientMock) wire() {
...
-func (m *clientMock) consume(n int, cmds []rueidis.Completed) []rueidis.RedisResult {
+func (m *clientMock) consume(n int, cmds []rueidis.Completed, viaCache bool) []rueidis.RedisResult {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	out := make([]rueidis.RedisResult, n)
 	for i, cmd := range cmds {
-		idx, ok := m.matchLocked(cmd)
+		idx, ok := m.matchLocked(cmd, viaCache)
 		if !ok {
 			err := fmt.Errorf("rueidiscompatmock: no expectation for command %v", cmd.Commands())
```
rueian: Can we make this error clear by differentiating the viaCache?

For example, in the following situation

```go
cm.ExpectGet("key").SetVal("v")            // forgot .ViaCache()
rdb.Cache(time.Second).Get(ctx, "key")
```

The error should be something like `no expectation for command [GET key] via cache`

rueidiscompatmock/adapter_test.go
```diff
@@ -3066,6 +3066,259 @@ func wantAnyErr(err error) error {
 	return nil
 }
 
+func TestCacheCompatGet(t *testing.T) {
+	ctx := context.Background()
+	rdb, cm := NewClientMock()
+	defer rdb.Close()
+
+	cm.ExpectGet("key").ViaCache().SetVal("v")
+	cmd := rdb.Cache(100 * time.Millisecond).Get(ctx, "key")
```
rueian:
```suggestion
	cmd := rdb.Cache(100*time.Millisecond).Get(ctx, "key")
```

rueidiscompatmock/adapter_test.go
```diff
@@ -3066,6 +3066,259 @@ func wantAnyErr(err error) error {
...
+	err := rdb.Cache(time.Second).Get(ctx, "key").Err()
+	if err == nil || !strings.Contains(err.Error(), "no expectation for command") {
+		t.Fatalf("expected no-expectation error, got %v", err)
+	}
+	if err := cm.ExpectationsWereMet(); err == nil {
+		t.Fatal("expected unmatched expectation error")
+	}
+}
+
+func TestCacheCompatViaCacheRequired(t *testing.T) {
```
rueian, line 3200: This duplicates TestCacheCompatGet.

rueidiscompatmock/adapter.go
SomSamantray: Done in a6410a9. Factored a shared `unmatchedErr(cmd, viaCache)` helper used by both `consume` and `consumeWithRaw`: when the unmatched call went through `DoCache`, the error is now `rueidiscompatmock: no expectation for command [GET key] via cache`; a plain-call mismatch is unchanged. Added `TestCacheCompatUnmatchedPlainCallOmitsViaCache` and tightened `TestCacheCompatUnmatched` to assert the suffix is present/absent as appropriate.

rueidiscompatmock/adapter_test.go
SomSamantray: Applied in a6410a9 — ran gofmt over the whole file to catch this and confirm nothing else was unformatted.

rueidiscompatmock/adapter_test.go
SomSamantray: You are right — it became an exact duplicate of `TestCacheCompatViaCacheRequired` once I added `.ViaCache()` to it in 3852a76 (TTL value is the only difference, and `TestCacheCompatTTLIrrelevant` already proves TTL does not affect matching). Removed `TestCacheCompatGet` in a6410a9.

rueidiscompatmock/viacache.go
```diff
@@ -0,0 +1,227 @@
+package rueidiscompatmock
+
+// ViaCache marks this expectation so it is only satisfied by a call routed
+// through DoCache (i.e. rueidiscompat.Cache(ttl).<Cmd>()). A plain, non-cached
+// call for the same command will not match it. Expectations that do not call
+// ViaCache match only plain calls and will not be satisfied by a call routed
+// through DoCache.
+
```
rueian:
```suggestion
```

rueian (approved):

PR #1024, opened 2026-08-18
fix(pipe): avoid nil cache panic with disabled static-TTL multi-cache

kurochan:
## Summary

Fix a nil pointer panic when `DoMultiCache` is used with multiple
`ToStaticTTL()` commands while client-side caching is disabled.

## Root cause

With `DisableCache: true`, the pipe does not initialize `p.cache`.

The `DoMultiCache` fallback converted `Cacheable` commands to `Completed`
commands but preserved the internal static-TTL flag. In the background read
loop, the second and subsequent replies then entered the static-TTL path and
called `p.cache.Update` or `p.cache.Cancel`, causing a nil pointer panic.

## Changes

- Clear static-TTL metadata before forwarding disabled-cache requests to
  `DoMulti`.
- Guard the background static-TTL path when no cache store exists.
- Add regression tests for successful replies and Redis errors with multiple
  static-TTL commands and a cancelable context.

## Compatibility

- `DisableCache` remains equivalent to `Do`/`DoMulti`.
- Client-side caching behavior is unchanged when enabled.
- Redis errors are returned to callers unchanged.

---

> [!NOTE]
> **Low Risk**
> Narrow fix for DisableCache + static-TTL edge case; enabled client-side caching paths are unchanged aside from the same clear-on-fallback in DoCache.
> 
> **Overview**
> Fixes a **nil pointer panic** when client-side caching is off (`DisableCache: true`) but callers still use `DoCache` / `DoMultiCache` with `ToStaticTTL()` commands.
> 
> With no cache store, those APIs already fall back to plain `Do` / `DoMulti`, but the static-TTL flag stayed on the wire commands. The background read loop then treated later replies as cacheable and called `p.cache.Update` / `Cancel` on a nil store—especially on the second+ command in a `DoMultiCache` batch.
> 
> **Change:** strip static-TTL metadata via `cmds.ClearStaticTTL` before forwarding in both disabled-cache paths so behavior matches untagged `Do`/`DoMulti`. Regression tests cover multi-command success, mixed Redis errors, and single `DoCache`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4245f7a90dad821ac02e6505ac582881b4e0bb5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pipe.go
```diff
@@ -1712,7 +1712,9 @@ func (p *pipe) DoMultiCache(ctx context.Context, multi ...CacheableTTL) *redisre
 	if p.cache == nil {
 		commands := make([]Completed, len(multi))
 		for i, ct := range multi {
-			commands[i] = Completed(ct.Cmd)
+			cmd := ct.Cmd
+			cmds.ClearStaticTTL(&cmd)
```
rueian, line 1717: Do we need to ClearStaticTTL at DoCache as well?

rueian (approved):

PR #1025, opened 2026-08-28
fix: resolve the ?addr=:port shorthand against the primary host

NotAFlightRisk:
Running two Redis instances on one box is common enough, and `?addr=:6380` is the shorthand for pointing at the second one. Do that against a primary that has its own port and you get back an address nothing can dial:

```go
rueidis.ParseURL("redis://cache:6379?addr=:6380")
// InitAddress: ["cache:6379", "[cache:6379]:6380"]
```

Thats v1.0.77. A port-only `?addr=` has no host of its own, so it falls back to `u.Host`, which still carries the primarys `:6379`, and the whole thing gets bracketed as though it were an IPv6 literal. It only goes wrong when the primary has an explicit port, so nearly always. And because it fails at connect time rather than at parse time, it reads like DNS or a firewall.

Giving that fallback the host the primary already resolved to leaves the primary address and the TLS `ServerName` alone. I diffed the parse output over a spread of URLs, bracketed and unbracketed IPv6 and `unix://` included, and only the `?addr=` entries move. `?addr=host` with no port I've left alone, its still resolving to the primary's host insted of `host:6379`. Most of the diff is the two test cases, copying the shape of the `?addr=` assertion above them.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to Redis URL parsing with targeted tests; wrong addresses fail at dial time but do not affect security-sensitive paths.
> 
> **Overview**
> Fixes **broken multi-instance Redis URLs** when the primary includes an explicit port and extra nodes use the `?addr=:port` shorthand (e.g. `redis://cache:6379?addr=:6380`).
> 
> In `ParseURL`’s `parseAddr` helper, the host fallback for port-only `addr` values now uses **`u.Hostname()`** instead of **`u.Host`**, so the resolved host is `cache` (or `[::1]` for IPv6) rather than the full `host:port` string that was incorrectly joined and bracketed as `[cache:6379]:6380`. Primary address and TLS `ServerName` behavior stay the same; only `?addr=` entries that omit a host change.
> 
> Tests cover hostname+port and bracketed IPv6 primaries with `?addr=:6380`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28879b562471f6fd6e600961e96c18ab0e3d13d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


url.go
```diff
@@ -23,10 +23,11 @@ func ParseURL(str string) (opt ClientOption, err error) {
 	if err != nil {
 		return opt, err
 	}
+	fallbackHost := u.Host
```
rueian:
```suggestion
```

url.go
```diff
@@ -23,10 +23,11 @@ func ParseURL(str string) (opt ClientOption, err error) {
...
 	parseAddr := func(hostport string) (host string, addr string) {
 		host, port, _ := net.SplitHostPort(hostport)
 		if host == "" {
-			host = u.Host
+			host = fallbackHost
```
rueian:
```suggestion
			host = u.Hostname()
```

url.go
```diff
@@ -52,6 +53,7 @@ func ParseURL(str string) (opt ClientOption, err error) {
 	}
 	if opt.InitAddress == nil {
 		host, addr := parseAddr(u.Host)
+		fallbackHost = host
```
rueian:
```suggestion
```

rueian (approved):

